PR #10, opened 2018-11-21
add a bit of documentation

regehr:


regehr:
I spend a while trying to figure out why I couldn't build things tonight, so I figured we might as well document which version of z3 is intended to compile against

PR #11, opened 2018-12-03
Build fixes.

LebedevRI:
There are two issues:
1. The code is using `uint32_t`/`int32_t` from <cstding>,
   while Z3 is using `__uint64`/`__int64` that are NOT aliases
   to `uint32_t`/`int32_t`, that are defined in <z3_api> as:
   ```
   #ifndef __int64
   #define __int64 long long
   #endif

   #ifndef __uint64
   #define __uint64 unsigned long long
   #endif
   ```
   To me that looks rather horrible :/
   I think just using *those* types is the simplest solution
2. `Z3_get_estimated_alloc_size()` function is very new.
   I'm on debian sid, amd64.
   ```
   $ dpkg -l | grep z3
   ii  libz3-4:amd64                                 4.4.1-0.4                                              amd64        theorem prover from Microsoft Research - runtime libraries
   ii  libz3-dev:amd64                               4.4.1-0.4                                              amd64        theorem prover from Microsoft Research - development files
   ii  python-z3                                     4.4.1-0.4                                              amd64        theorem prover from Microsoft Research - Python bindings
   ii  z3                                            4.4.1-0.4                                              amd64        theorem prover from Microsoft Research
   ```
   Ancient, i know. There is a debian bugreport asking for the update
   https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=909494
   but it is not there yet :/

   I'm not sure if i will suffer OOM's because of this yet.


LebedevRI:
Aha, and those types are gone in newer versions, good.

nunoplopes:
Sorry, I can't merge this in.
If the Z3 version you have doesn't even have those instructions, it means it's really old. I've fixed several important bugs in Z3 that show up our workloads, so it's not really safe to use such old versions.
Z3 is not hard to compiler from source, and they also provide precompiled packages for several distros.
So I don't think it's worth it to maintain support for extremely old versions of Z3.

cmake/FindZ3.cmake
```diff
@@ -2,3 +2,24 @@ find_path(Z3_INCLUDE_DIR NAMES z3.h)
 find_library(Z3_LIBRARIES NAMES z3)
```
LebedevRI: Then please fix this to specify the required version.

nunoplopes: ok, added version checking to cmake.

LebedevRI: Thank you.

PR #20, opened 2018-12-21
use add_llvm_library instead of add_llvm_loadable_module

pranavk:
add_llvm_loadable_module() was removed here:
https://reviews.llvm.org/D51748

### DIFF
--- a/tv/CMakeLists.txt
+++ b/tv/CMakeLists.txt
@@ -17,6 +17,11 @@ include(AddLLVM)
 include_directories(${LLVM_INCLUDE_DIRS})
 add_definitions(${LLVM_DEFINITIONS})
 
-add_llvm_loadable_module(tv tv.cpp)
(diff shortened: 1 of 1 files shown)


nunoplopes:
This patch fails with a checkout of LLVM from 2 days ago. Can we maintain compatibility with LLVM 7, at least?
I think it might be interesting to run experiments with older LLVM versions.

regehr:
ugh, yes, there should be something backwards compatible that you can do here Pranav. I'd have expected the first thing you tried to be back-compatible since at a glance this patch was just removing something, not adding something, but I guess it's more complicated than that...

regehr:
at worst (and please try to do it a different way) I'm sure there's a way to test the LLVM revision and do one thing or the other

pranavk:
Okay, the latest patch will work for all LLVM versions

PR #21, opened 2018-12-21
add support for llvm.cttz

pranavk:
https://llvm.org/docs/LangRef.html#llvm-cttz-intrinsic


ir/instr.h
```diff
@@ -70,6 +70,15 @@ class Select final : public Instr {
   smt::expr getTypeConstraints(const Function &f) const override;
 };
 
+class Cttz final : public Instr {
+  Value &src, &zeroflag;
+public:
+  Cttz(Type &type, std::string &&name, Value &src, Value& zeroflag)
```
regehr: plz move the last & so it's next to "zeroflag" instead of next to Value

regehr:
seems good overall!
(Nuno, Pranav works for me now and if this code is good, he'll continue adding intrinsics)

tests/unit/cttz.opt
```diff
@@ -0,0 +1,28 @@
...
+  =>
+%r = undef
+
+Name: cttz def
+%r = cttz i8 0, 0
+  =>
+%r = 8
+
+Name: cttz variable
+%k = shl 1, 3
```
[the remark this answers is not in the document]
pranavk: i am not really sure i can ask alive to check if something is greater than something.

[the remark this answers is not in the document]
regehr: something like this:

%k = lshr %in, 3
%r = cttz %k, 1
%out = icmp sge %r, 3
  =>
%out = true

also it's good to have a negative test or two

%k = lshr %in, 3
%r = cttz %k, 1
%out = icmp sge %r, 4
  =>
%out = true

ir/instr.h
```diff
@@ -70,6 +70,15 @@ class Select final : public Instr {
   smt::expr getTypeConstraints(const Function &f) const override;
 };
 
+class Cttz final : public Instr {
```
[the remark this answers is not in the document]
pranavk: I didn't make it binop because second parameter to cttz is not really an operand unlike other binop instructions; it's just a flag. Though if you think making it binop is better idea, I can do that.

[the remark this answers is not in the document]
nunoplopes: Well, it's still a binary operation. Unless it is hard to fit in BinOp, I would prefer to move it there. We really don't want an explosion in the number of IR ops. Thanks!

smt/expr.cpp
```diff
@@ -523,6 +523,19 @@ expr expr::lshr_exact(const expr &rhs) const {
...
 }
 
+expr expr::cttz() const {
+  auto nbits = bits();
+  assert(nbits > 0);
+
+  auto src = *this;
+  auto cond = mkUInt(nbits, nbits);
+  for (int i = nbits - 1; i >= 0; --i) {
+    cond = mkIf(src.extract(i, i) == 1u, mkUInt(i, nbits), cond);
```
nunoplopes: remove "src."  to avoid the copy above.

ir/instr.h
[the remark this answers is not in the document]
pranavk: And since second operand in Cttz is a flag, Binop::getTypeConstraints() will not produce correct type constraints for Cttz unless I add custom code to allow exceptions for Cttz in getTypeConstraints(). Are you good with allowing such exceptions in Binop::getTypeConstraints()?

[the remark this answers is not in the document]
nunoplopes: That's fine. Once we see a few more we can think how to abstract it properly.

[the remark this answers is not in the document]
pranavk: Okay. It's parsed as binop now. Let me know if something doesn't look good to you.

smt/expr.cpp
```diff
@@ -523,6 +523,18 @@ expr expr::lshr_exact(const expr &rhs) const {
   return (lshr(rhs) << rhs) == *this;
 }
 
+expr expr::cttz() const {
+  auto nbits = bits();
+  assert(nbits > 0);
```
nunoplopes: this assert is not needed

tools/alive_parser.cpp
```diff
@@ -365,7 +366,14 @@ static unique_ptr<Instr> parse_binop(string_view name, token op_token) {
...
   parse_comma();
-  auto &b = parse_operand(type);
+
+  // if second operand's type is not mentioned,
+  // assume type to be same as first
+  auto *rhsType = &parse_type();
+  if (!rhsType->isDefined()) {
+    rhsType = &type;
+  }
+  auto &b = parse_operand(*rhsType);
```
nunoplopes: Uhm, not sure I like this..  Can you factor out this change in a new function (one that takes a default type and returns that default type if it there's no type to parse)?

ir/type.h
```diff
@@ -17,6 +17,7 @@ class VoidType;
 class Type {
 protected:
   std::string name;
+  bool defined;
```
nunoplopes: as noted in the review of the parser, I don't think this change is needed.

ir/instr.cpp
```diff
@@ -212,15 +214,25 @@ StateValue BinOp::toSMT(State &s) const {
   case Xor:
     val = a ^ b;
     break;
+
+  case Cttz:
+    val = a.cttz();
+    not_poison &= !(a == expr::mkUInt(0, a.bits()) && b == 1u);
```
nunoplopes: can we write this as e.g. b == 0 || a != 0? I think it's more readable (plus a smaller formula).
Also, could you send a patch for LLVM's LangRef.html to switch it officially to poison?

regehr: regarding the poison switch -- this should be done for ctlz at the same time, since these intrinsics should be symmetrical in this respect

nunoplopes: yes, that would be great!

ir/type.h
pranavk: i though it would be more cleaner and efficient this way; 

1) I didn't want to use `dynamic_cast<SymbolicType>` to check if `parse_type` actually parsed something or not.

2) We were using `defined` in `IntType` class anyways and there was no way to check if any `SymbolicType` is defined or not. 

So, hoisting `defined` to the base class seemed like a good choice to me. However, your other comment about returning a default type passed if there's no type to parse sounds good and I can have something like `try_parse_type` for this, surely.

pranavk: @nunoplopes any thoughts on this? i have updated the patch with new `try_parse_type` method

nunoplopes: It doesn't seem to be needed. `try_parse_type` can simply tokenizer.peek() to see if there's a type following or not.
I'm just trying to intermingle these changes to the backbone with something else :)

Do you think the peek() solution works?

pranavk: Alright; peek solution works as well. I've updated the patch.

nunoplopes:
Awesome, thank you!

PR #22, opened 2019-01-10
Miscellaneous changes to enable Souper to use upstream Alive

manasij7479:


smt/expr.h
```diff
@@ -61,6 +59,7 @@ class expr {
     std::swap(ptr, other.ptr);
   }
 
+  expr(Z3_ast ast);
```
nunoplopes: why do you need to expose this? I did a very shallow review of your Souper patch but it wasn't immediately obvious why would you need this,

smt/expr.h
```diff
@@ -90,6 +89,7 @@ class expr {
   bool isAllOnes() const;
   bool isSMin() const;
   bool isSMax() const;
+  Z3_app isApp() const;
```
nunoplopes: Same here: this probably shouldn't be exposed either since I don't really want users to rely on us using Z3 since we may want to have multiple backends.
Can you tell me what do you need exactly?

manasij7479: (Combined answer for both)
I need a list of expressions corresponding to input variables matching some criteria.  
(i.e. two separate lists corresponding to variables and symbolic constants)

I am trying to extract this from StateValue instances which look like:  
`(ite (= ty_%var_a #b00) %var_a undef_1)`  

Is there another way to get the SMT expression for %var_a?

nunoplopes: Well, if you just want the inputs, you can ask the Function for the inputs. State also stores quite a bit of info.
You probably don't need to traverse the formula, That's very brittle anyway since encodings can (and will) change over time. We can expose further APIs if information is missing.  I just need to know exactly what's missing/needed :)

manasij7479: Using the function inputs seems to require traversing formulas too, at this moment.  
```
  for (auto &i : t.tgt.getInputs()) {
    std::cerr << i.toSMT(tgt_state) << std::endl;
  }
```
The SMT expressions I get from this look similar to the what I had before:
`(ite (= ty_%var_x #b00) %var_x undef_11) / (= #b0 ((_ extract 1 1) ty_%var_x))
`

Can we have an API that gives me the smt expr for %var_x (without requiring simplifications)?

smt/ctx.cpp
```diff
@@ -9,7 +9,7 @@ namespace smt {
 
 context ctx;
 
-void context::init() {
+void context::initialize() {
```
nunoplopes: This is problematic. Z3 also has an smt::context class, so this can be a mess in the long run.
FOr dynamic linking is fine, because symbols are hidden with the default build options. Why are statically linking Z3? Is this a useful use case?  (If so, this fix is not a future-proof one)

regehr: there's a weird cmake problem on OSX where, no matter what we try, it is pulling the system-installed libz3.dylib no matter what we do. sometimes this is OK but other times it creates problems, and I'd like to avoid it.

since nothing we've tried fixes the cmake thing, linking z3 statically seems like an OK workaround to make sure we get the one that we want.

regehr: in other words, all things like this are not working:
```
-DZ3_LIBRARIES=/Users/johnregehr/souper-regehr/third_party/z3-install/lib/libz3.dylib
```

nunoplopes:
Closing this PR since there hasn't been any progress and it has some issues as discussed previously.
Please open separate PRs for each of the issues if still relevant. Thank you!

PR #23, opened 2019-01-11
add support for llvm.ctlz and llvm.ctpop

pranavk:
https://llvm.org/docs/LangRef.html#llvm-ctlz-intrinsic


smt/expr.cpp
```diff
@@ -547,6 +547,18 @@ expr expr::ctlz() const {
   return cond;
 }
 
+expr expr::ctpop() const {
+  C();
+  auto nbits = bits();
+
+  auto res = mkUInt(0, nbits);
+  for (unsigned i = 0; i < nbits; ++i) {
+    res = res + (lshr(mkUInt(i, nbits)) & mkUInt(1, nbits));
```
nunoplopes: this would be better with an extract + zext instead of shift + and.

pranavk: indeed, looks much cleaner now. i have force-pushed the changes.

PR #24, opened 2019-01-11
add support for llvm.ctpop

pranavk:
This is a separate PR; however, this depends/rebases on top of #23 

I plan add more commits to this PR regarding llvm.bitreverse and llvm.bswap in this PR.


regehr:
I think something got garbled here, this patch also adds ctlz

pranavk:
@regehr yes, this work was done after I did the work on ctlz. If you just click on the 2nd commit in this PR (add support for llvm.ctpop), it should be easy to review.

pranavk:
Let me actually just keep updating the same earlier PR to avoid PR-dependencies and confusion like this. Closing this.

PR #25, opened 2019-01-15
add support for llvm.bitreverse and llvm.bswap

pranavk:
https://llvm.org/docs/LangRef.html#llvm-bitreverse-intrinsics


smt/expr.cpp
```diff
@@ -523,6 +523,18 @@ expr expr::lshr_exact(const expr &rhs) const {
   return (lshr(rhs) << rhs) == *this;
 }
 
+expr expr::bitreverse() const {
+  C();
+  auto nbits = bits();
+
+  expr res = mkUInt(0, nbits);
+  for (unsigned i = 0; i < nbits; i++) {
+    res = extract(i, i).zext(nbits - 1) | (res << mkUInt(1, nbits));
```
regehr: Pranav, did you investigate any other encodings for this function to see which generate less trouble for the solver?

nunoplopes: I suggest extract of each bit + concat to append the bits together.

ir/instr.cpp
```diff
@@ -254,6 +254,7 @@ void UnaryOp::print(ostream &os) const {
   const char *str = nullptr;
   switch (op) {
   case BitReverse:  str = "bitreverse "; break;
+  case BSwap:       str = "bswap "; break;
```
LebedevRI: The PR title is misleading then?

pranavk: yes, i was about to change that; you were too quick :)

smt/expr.cpp
pranavk: i will take care of this in my next push

pranavk: changed to extract + concat

PR #28, opened 2019-02-07
Support for overflow instructions

pranavk:
Still working on it. Creating a PR in case someone notice any serious mistakes.


ir/instr.cpp
```diff
@@ -227,6 +231,14 @@ StateValue BinOp::toSMT(State &s) const {
     not_poison &= (b == 0u || a != 0u);
     break;
 
+  case SAdd_Overflow:
+    val = a + b;
+    val = val.concat(expr::mkIf(a.add_no_soverflow(b), expr::mkUInt(0, 1), expr::mkUInt(1, 1)));
```
regehr: it may not matter but it seems simpler to use !add_no_soverflow() instead of an ITE

ir/instr.h
```diff
@@ -197,5 +198,4 @@ class Unreachable final : public Instr {
   StateValue toSMT(State &s) const override;
   smt::expr getTypeConstraints(const Function &f) const override;
 };
-
```
regehr: avoid whitespace changes

ir/type.cpp
```diff
@@ -231,6 +235,34 @@ void VectorType::print(ostream &os) const {
   os << "TODO";
 }
 
+expr AggregateType::getTypeConstraints() const {
+  // Only allow aggregates of type {ix, i1} for overflow instrs
+  return childrenSize.size() == 2 &&
+         childrenSize[0] <= 64 &&
```
regehr: why this constraint?

ir/instr.cpp
pranavk: that doesn't work. add_no_soverflow returns a true/false expr and implicit cast to i1 doesn't work.

ir/type.cpp
pranavk: removed.

ir/instr.cpp
```diff
@@ -227,25 +231,47 @@ StateValue BinOp::toSMT(State &s) const {
...
+    break;
+
+  case ExtractValue:
+    val = expr::mkIf(b == 1u, a.extract(0, 0), a.extract(a.bits() - 1, 1));
```
nunoplopes: well, this needs to check the type info.

ir/type.cpp
```diff
@@ -231,10 +236,47 @@ void VectorType::print(ostream &os) const {
...
+}
+
+void AggregateType::print(ostream &os) const {
+  assert(childrenSize.size() > 0);
+  os << "{";
+  os << "i" << childrenSize[0];
+  for (unsigned i = 1; i < childrenSize.size(); i++) {
+    os << ", i" << childrenSize[i];
+  }
+  os << "}";
```
nunoplopes: nitpick: '}'

ir/type.cpp
```diff
@@ -231,10 +236,47 @@ void VectorType::print(ostream &os) const {
...
+}
+
+expr AggregateType::getChildrenConstraints(Type &type) const {
+  return (type.enforceIntType() && type.sizeVar() == childrenSize[0]) ||
+         (type.enforceIntType() && type.sizeVar() == childrenSize[1]);
```
nunoplopes: can you generalize this so that aggregate type is a container of arbitrary types pls?

pranavk: do you want to make use of Z3 records/tuple for this purpose? Currently, I am using bitvectors as representation of aggregate type (since we currently only allow integers as children components and it's enough for overflow intrinsics). Not sure, but adding support for arbitrary types for aggregate type with aribtrary number of components may not be possible with just bitvectors.

ir/instr.cpp
pranavk: we already enforcing the type info in BinOp::getTypeConstraints() for extractvalue. So, we can safely assume `a` to be a aggregate bitvector and `b` to be a BV of size 1, isn't it?

ir/type.cpp
regehr: @pranavk I don't think it makes sense that something like this isn't possible using bitvectors -- alive2 simple has to keep track of the type information instead of relying on z3

@nunoplopes I had mentioned to Pranav that you said that Z3 data types don't play well with quantifiers -- I assume that this forces a low-level solution here?

nunoplopes: Right, no Z3 datatypes.
You can continue to use bitvectors. The result will be a concatenation of the underlying representation of each of the components (which will all be bitvectors).

ir/instr.cpp
pranavk: okay, this makes sense now after we made our aggregate type generic; i will make it generic here too in my next commit

pranavk:
okay. Aggregate type is more generic now - extractvalue works on any sort of aggregate type. One thing that I am not happy with is that you have to mention the type on the RHS to disambiguate between the possible values that extractvalue can return (see first test, eg). Since second argument to extractvalue is supposed to be a constant always, maybe we can put some tighter type constraints so that extractvalue knows its type at the typechecking time completely. This is still a TODO.

ir/instr.cpp
```diff
@@ -6,6 +6,7 @@
 #include "smt/expr.h"
 #include "smt/solver.h"
 #include "util/compiler.h"
+#include "ir/type.h"
```
nunoplopes: please sort this

ir/instr.cpp
```diff
@@ -227,25 +232,57 @@ StateValue BinOp::toSMT(State &s) const {
...
+    val = val.concat(expr::mkIf(a.add_no_soverflow(b), expr::mkUInt(0, 1), expr::mkUInt(1, 1)));
+    break;
+
+  case ExtractValue:
+  {
+    auto &aggType = static_cast<AggregateType&>(lhs.getType());
+    val = aggType.extract(a, b);
+
+    uint64_t n;
+    not_poison &= b.isUInt(n);
```
nunoplopes: is this not_poison bit needed? ENSURE() seems more appropriate since this should always hold by the type checker.

ir/instr.cpp
```diff
@@ -227,25 +232,57 @@ StateValue BinOp::toSMT(State &s) const {
...
 expr BinOp::getTypeConstraints(const Function &f) const {
   expr instrconstr;
   switch (op) {
+  case ExtractValue:
+  {
+    instrconstr = lhs.getType().enforceAggregateType() &&
+                  rhs.getType().enforceIntType();
+    auto aggregateType = dynamic_cast<AggregateType&>(lhs.getType());
+    instrconstr = instrconstr &&
+                  aggregateType.getChildrenConstraints(getType());
```
nunoplopes: missing check for non-nullness of agregateType

ir/type.cpp
```diff
@@ -231,10 +236,90 @@ void VectorType::print(ostream &os) const {
...
+  return res;
+}
+
+expr AggregateType::operator==(const AggregateType &rhs) const {
+  if (childrenType.size() != rhs.childrenType.size())
+    return false;
+
+  expr res(false);
+  for (unsigned i = 0; i < childrenType.size(); i++) {
+    res = res || *childrenType[i] == *rhs.childrenType[i];
```
nunoplopes: please use |=

ir/type.h
```diff
@@ -119,9 +123,38 @@ class VectorType final : public Type {
...
+// of int type.
+// Eg: {ix, i1}. Enforced in getTypeConstraints()
+class AggregateType final : public Type {
+  std::vector<Type*> childrenType;
+
+public:
+  AggregateType(std::string &&name) : Type(std::move(name)) {}
+  AggregateType(std::string &&name, std::initializer_list<Type*> types)
+    : Type(std::move(name)), childrenType(types) {
+    assert(childrenType.size() > 0);
```
nunoplopes: !empty

ir/instr.cpp
pranavk: typechecker can only enforce the number of bits of the type but what happens if the real value is within the enforced bits but still greater than the total aggregate size? shouldn't it be poison then?

pranavk: fixed

pranavk:
in the new commit, i moved much of the important logic to typechecker. It can now detect out of bounds arguments (when argument is greater than aggregate size) and deduce the return type of the `extractvalue` based on the argument. This also removes the concern/TODO I had mentioned in previous comment.

Removing the WIP now as I think this PR is in much better state.

pranavk:
Can we get some feedback on this code, please? If it looks good, I would move forward adding other overflow instructions.

ir/type.cpp
```diff
@@ -247,10 +252,87 @@ void VectorType::print(ostream &os) const {
...
+  os << '{';
+  childrenType[0]->print(os);
+  for (unsigned i = 1; i < childrenType.size(); i++) {
+    os << ", ";
+    childrenType[i]->print(os);
+  }
+  os << '}';
+}
+
+expr AggregateType::getChildConstraints(Type &type, unsigned index) const {
```
nunoplopes: This method is strange. I thing something that returns the type of element x is sufficient.

ir/type.cpp
```diff
@@ -247,10 +252,87 @@ void VectorType::print(ostream &os) const {
...
+  if (index < childrenType.size())
+    return *childrenType[index] == type;
+  return false;
+}
+
+expr AggregateType::extract(const expr &a, const expr &b) const {
```
nunoplopes: This method should accept a constant index and use more descriptive names than a & b.
I don't think there's any use case where a non-constant index is used, but please correct me if I'm mistaken.

ir/type.h
```diff
@@ -123,9 +128,38 @@ class VectorType final : public Type {
...
+// of int type.
+// Eg: {ix, i1}. Enforced in getTypeConstraints()
+class AggregateType final : public Type {
+  std::vector<Type*> childrenType;
+
+public:
+  AggregateType(std::string &&name) : Type(std::move(name)) {}
+  AggregateType(std::string &&name, std::initializer_list<Type*> types)
+    : Type(std::move(name)), childrenType(types) {
+    ENSURE(!childrenType.empty());
```
nunoplopes: it can be empty, no problem.
Also, ENSURE != assert.

ir/type.h
```diff
@@ -123,9 +128,38 @@ class VectorType final : public Type {
...
+
+  unsigned bits() const override;
+  smt::expr enforceAggregateType() const override;
+  smt::expr getTypeConstraints() const override;
+  smt::expr operator==(const AggregateType &rhs) const;
+  void fixup(const smt::Model &m) override;
+  smt::expr enforceIntOrPtrOrVectorType() const override;
+  void print(std::ostream &os) const override;
+
+  unsigned getChildrenSize() const { return childrenType.size(); }
```
nunoplopes: name is strange. Also there's casting going on. You can use auto as return type instead.
What about numElements()?

ir/type.h
```diff
@@ -123,9 +128,38 @@ class VectorType final : public Type {
 };
 
 
+// Currently only supports aggregate type with two children
+// of int type.
+// Eg: {ix, i1}. Enforced in getTypeConstraints()
+class AggregateType final : public Type {
+  std::vector<Type*> childrenType;
```
nunoplopes: I think Type is uneeded. children is enough.

ir/type.h
```diff
@@ -123,9 +128,38 @@ class VectorType final : public Type {
 };
 
 
+// Currently only supports aggregate type with two children
+// of int type.
+// Eg: {ix, i1}. Enforced in getTypeConstraints()
+class AggregateType final : public Type {
```
nunoplopes: Please call it structure rather than aggregate, to match LLVM's type system.

ir/value.h
```diff
@@ -42,6 +42,8 @@ class Value {
 
   static void reset_gbl_id();
 
+  virtual bool isIntConst(int64_t &n) const { return false; }
```
nunoplopes: this is not needed.

ir/instr.cpp
```diff
@@ -226,25 +231,55 @@ StateValue BinOp::toSMT(State &s) const {
     val = a.ctlz();
     not_poison &= (b == 0u || a != 0u);
     break;
+  
+  case SAdd_Overflow:
+    val = a + b;
+    val = val.concat(expr::mkIf(a.add_no_soverflow(b), expr::mkUInt(0, 1), expr::mkUInt(1, 1)));
```
nunoplopes: this line is over 80 columns, so are many others.

ir/type.cpp
```diff
@@ -247,10 +252,87 @@ void VectorType::print(ostream &os) const {
   os << "TODO";
 }
 
+expr AggregateType::getTypeConstraints() const {
+  if (childrenType.size() < 2)
```
nunoplopes: please remove this artificial restriction. I don't see any reason for this.

tools/alive_parser.cpp
```diff
@@ -220,9 +220,38 @@ static void parse_comma() {
 }
 
 
+static unordered_map<unsigned, unique_ptr<AggregateType>> overflow_aggregate_types;
 static vector<unique_ptr<SymbolicType>> sym_types;
 static unsigned sym_num;
 
+static Type& get_overflow_type(unsigned bitwidth) {
+  if (overflow_aggregate_types.find(bitwidth) != overflow_aggregate_types.end()) {
+    return *overflow_aggregate_types[bitwidth];
```
nunoplopes: this code does 3 map lookups, while it can be change to do only emplace (that is patched if type not there).

tools/alive_parser.cpp
```diff
@@ -220,9 +220,38 @@ static void parse_comma() {
...
+  }
+
+  initializer_list<Type*> children = { int_types[bitwidth].get(), int_types[1].get() };
+  auto t = make_unique<AggregateType>("oaggty_" + to_string(bitwidth), move(children));
```
nunoplopes: just change the constructor to receive a vector&&.

tools/alive_parser.cpp
```diff
@@ -418,10 +446,18 @@ static unique_ptr<Instr> parse_binop(string_view name, token op_token) {
   case UADD_SAT: op = BinOp::UAdd_Sat; break;
   case SSUB_SAT: op = BinOp::SSub_Sat; break;
   case USUB_SAT: op = BinOp::USub_Sat; break;
+  case SADD_OVERFLOW:
+    op = BinOp::SAdd_Overflow;
+    rettype = std::ref(get_overflow_type(type.bits()));
+    break;
+  case EXTRACTVALUE:
+    op = BinOp::ExtractValue;
+    rettype = std::ref(get_sym_type());
```
nunoplopes: why override rettype?

tools/alive_parser.cpp
```diff
@@ -418,10 +446,18 @@ static unique_ptr<Instr> parse_binop(string_view name, token op_token) {
   case UADD_SAT: op = BinOp::UAdd_Sat; break;
   case SSUB_SAT: op = BinOp::SSub_Sat; break;
   case USUB_SAT: op = BinOp::USub_Sat; break;
+  case SADD_OVERFLOW:
+    op = BinOp::SAdd_Overflow;
+    rettype = std::ref(get_overflow_type(type.bits()));
```
nunoplopes: this will crash if type is symbolic

tools/alive_parser.cpp
```diff
@@ -394,8 +417,13 @@ static unique_ptr<Instr> parse_binop(string_view name, token op_token) {
   auto &type = parse_type();
   auto &a = parse_operand(type);
   parse_comma();
-  auto &type2 = try_parse_type(type);
+
+  // 2nd binop argument handling
+  auto type2 = std::ref(try_parse_type(type));
+  if (op_token == EXTRACTVALUE)
+    type2 = std::ref(get_int_type(64));
```
nunoplopes: why override type2?

ir/type.h
pranavk: hmm, my understanding is that assert() only asserts if NDEBUG is not defined. Looking at definition of ENSURE in compiler.h, it also says the same: if NDEBUG is not defined, it asserts, otherwise it's a nop. So then, isn't writing a ENSURE or assert same thing then?

I also think that definition of ENSURE in compiler.h right now is superfluous as assert is going to be nop anyways if NDEBUG is defined - we don't need to specifically take care of that.

nunoplopes: ENSURE(x) always executes 'x'. It guarantees side-effects, while assert doesn't. That's the difference.

tools/alive_parser.cpp
pranavk: because `extractvalue` is different than other instructions in the sense that it doesn't return the same type as its first argument - it rather returns type based on second argument; therefore, I assigned it a symbolic type to be determined later.

tools/alive_parser.cpp
pranavk: If I don't override it and we don't want second argument of `extractvalue` to mention a type with the constant, it will automatically assume second type to be first argument's type (which in this case is aggregate type and incorrect). Therefore, I made second argument of `extractvalue` be implicitly parse as 64 bit integer.

tools/alive_parser.cpp
nunoplopes: ok, makes sense since we aren't parsing the result type, but please get rid of this std::ref ugliness :)

tools/alive_parser.cpp
nunoplopes: Then just use try_parse_type() without passing type. If the type is given it should be respected, instead of overriding it.

ir/value.h
pranavk: i added this to be able to retrieve the value of the constant in `BinOp::getTypeConstraints()`'s `::ExtractValue` switch case. I couldn't find any easy way to retrieve this constant from a `Value`. Do you have some suggestions?

tools/alive_parser.cpp
pranavk: Just wanted to avoid raw pointers; thought reference_wrapper is cleaner than using raw pointers. But okay, i have removed the use of reference wrapper now.

pranavk:
Addressed most of your comments from the last review. But now it breaks the tests. I need to spend another session on this before it works again.

pranavk:
ok. tests passes again now

nunoplopes:
BTW please watch out for tabs next time and line limits. I'll fix them this time.

pranavk:
Thanks for merging this. I wanted to add other overflow instructions to this PR too after you are happy with the changes for this one. But no worries, I will do that in a separate PR now.

> BTW please watch out for tabs next time and line limits. I'll fix them this time.

I configured my editor to use spaces only but it seems it still uses tabs when indenting a line of code which spans multiple lines. Eg: `val = val.concat(...` in `Sadd_Overflow` switchcase. I will see into it. Line limits should already have been taken care of but apologies if I missed a couple.

LebedevRI:
> I configured my editor

You can add a git pre-commit hook that will happily reformat the diff in your to-be-commit.
1. Download/symlink https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/git-clang-format into somewhere in `$PATH`.
2. Add this snippet to `/usr/share/git-core/templates/hooks/pre-commit`, and/or `.git/hooks/pre-commit` of the repos you want:
```diff
$ diff -p -U4 /usr/share/git-core/templates/hooks/pre-commit.sample /usr/share/git-core/templates/hooks/pre-commit
--- /usr/share/git-core/templates/hooks/pre-commit.sample       2019-01-22 09:32:28.000000000 +0300
+++ /usr/share/git-core/templates/hooks/pre-commit      2019-03-17 19:36:01.659392441 +0300
@@ -44,6 +44,17 @@ If you know what you are doing you can d
 EOF
        exit 1
 fi
 
+# format everything
+res=$(git clang-format --diff | wc -l)
+
+if [ $res -ne 1 ]
+then
+       git clang-format
+       echo "Commit did not match clang-format"
+       exit 1;
+fi
+
 # If there are whitespace errors, print the offending file names and fail.
 exec git diff-index --check --cached $against --
+
```

pranavk:
Great! Thanks for this. I didn't know we have a `.clang-format` sitting in the repo. I've integrated it with my git now; new commits should be formatted now. However, it seems the current code is still far from how the current clang-format configuration wants it to be. Running `clang-format -i` on any random file suggests a lot of changes. It sounds weird to me now that we should be running clang-format on new commits while rest of the code base doesn't conform to clang-format.

LebedevRI:
That was my personal suggestion, neither do i claim that the codebase fully follows that `.clang-format` style (i just added it in my pr because @nunoplopes noted that some code was not formatted 'as it should be', which means there are more inconsistencies between what clang-format will now produce and what is 'expected'), nor am i in the position to require for all the code to be formatted that way.

LebedevRI:
@pranavk @nunoplopes i wanted to verify `smul_overflow`, and was surprised to notice that this only added support for a single `SAdd_Overflow`.
Do you plan to look into adding support for the rest real soon, or can i take a look?

nunoplopes:
@LebedevRI not me; dunno about @pranavk

regehr:
yeah, I believe this change involved enough serious alive2 changes that he just wanted to get those out of the way before doing the rest, which should be pretty easy now!

pranavk:
@LebedevRI go ahead, if you have time. If not, I'll do this by next week. let me know.

LebedevRI:
Okay, thank you for replying, i will try to look into this tomorrow.

PR #29, opened 2019-02-07
CMake: mark 'check' target as using terminal.

LebedevRI:
Else, all the `lit` output is being buffered, and is flushed after
the target finishes 'building'. In particular, that hides the
progress meter.

Same as https://github.com/google/souper/pull/404


LebedevRI:
Thank you.

nunoplopes:
Thank you for the patches :)

LebedevRI:
Was just going through the initial accustomization with the code before trying to finally look into funnel shift support here.

PR #30, opened 2019-02-07
CMake: don't override the rest of CMAKE_CXX_FLAGS_*

LebedevRI:
Same as https://github.com/AliveToolkit/alive2/pull/6
I wanted to pass custom/have righter control over
-O/-g/-UNDEBUG flags, but that is not possible currently,
since the per-build-type flags are still being overrided.

Unless that is intentional of course?

The alternative, is [for me] to specify custom CMAKE_BUILD_TYPE.


LebedevRI:
Hm, i guess i wanted to reference https://github.com/AliveToolkit/alive2/pull/4 here, `git log` digging failed me.

LebedevRI:
Thank you.

PR #31, opened 2019-02-08
TransformVerify: only check src instrs that exist in dst too

LebedevRI:
At least the lhs-intermediate-variable.opt test causes a crash otherwise.
Because we go through every instruction on the LHS, and verify it against
it's counterpart on RHS. But naturally, if said instruction only exists
on one side, we can't get it's counter-part from the other side
so the std::unordered_map<>::at() throws an exception, and alive crashes.

### DIFF
--- a/tests/unit/lhs-intermediate-variable.opt
+++ b/tests/unit/lhs-intermediate-variable.opt
@@ -1,0 +1,7 @@
+Name: check that an intermediate variable that only exists on LHS does not cause an exception
+%c = udiv %a, %b
+%e = add %c, %d
+  =>
+%e = shl %f, %g
+
+; ERROR: Target is more poisonous than source for i1 %e
--- a/tests/unit/rhs-intermediate-variable.opt
+++ b/tests/unit/rhs-intermediate-variable.opt
@@ -1,0 +1,7 @@
+Name: check that an intermediate variable that only exists on RHS does not cause an exception
+%c = shl %a, %b
+  =>
+%f = udiv %d, %e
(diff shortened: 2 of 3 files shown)


nunoplopes:
This is by design actually. You can use -root-only if you only to check the final result of the expression.
Essentially Alive2 is still missing the functionality to copy instructions from LHS to RHS. Alive1 has this feature, and it's very convenient. Once that feature lands, this bug will magically go away.
In summary, well-formed programs shouldn't arrive this far.
Sorry you found this the hard way; Alive2 is still in very early days :)

LebedevRI:
> This is by design actually. You can use -root-only if you only to check the final result of the expression.
> Essentially Alive2 is still missing the functionality to copy instructions from LHS to RHS. Alive1 has this feature, and it's very convenient. Once that feature lands, this bug will magically go away.

Uh oh, ok.
I'm guessing this piece of broken glass is left there so there won't be need to comb through
the code trying to get rid of such sanity checks when they are no longer needed?

> In summary, well-formed programs shouldn't arrive this far.
> Sorry you found this the hard way; Alive2 is still in very early days :)

Well, i was trying to write test cases for funnel shift :)

nunoplopes:
Right, and also this kind of checks aren't sufficient to support things like:
```
%b = add %a, %a
%c = mul %b, 2
=>
%c = shl %b, 1
```

(I know the example isn't very interesting; it's just to show that RHS can reuse vars from LHS without having to redeclare them)

PR #32, opened 2019-02-09
Support for LLVM IR funnel shift left/right intrinsic

LebedevRI:
LangRef: https://releases.llvm.org/7.0.0/docs/LangRef.html#llvm-fshl-intrinsic

```
The ‘llvm.fshl’/‘llvm.fshr’ family of intrinsic functions performs
a funnel shift left/right: the first two values are concatenated
as { %a : %b } (%a is the most significant bits of the wide value),
the combined value is shifted left/right,
and the most/least significant bits are extracted to produce a result
that is the same size as the original arguments.
<...>
For vector types, the operation occurs for each element of the vector.
The shift argument is treated as an unsigned amount modulo
the element size of the arguments.
```

Questions:
* Should there be constant-evaluation fastpaths in these `expr::rotate()` / `expr::funnelShift()`?
  I have avoided adding them for now because the bit width where these computations are
  done matters, and right now all such `binop_ufold()` implicitly widen to 64-bit.
  Hmm, but now that i think about it, maybe it would still work, will take a second look..
* Any specific tests i'm missing?
* Should this be a `TernaryOp`, or should i just add `FunnelShiftOp`?

Bugs:
* There seems to be something off with `poison` propagation.
  See `tests/unit/fshl/poison-7.opt` - that **should** be poison,
  we combined a known-poison value with something else.
  It does work without all that extra instructions - see `tests/unit/fshl/poison-7-v2.opt`

Observations: alive2 api has a few rough edges
* `zext` takes the increment, not the destination width
* `extract` takes bit positions instead of `{skip, len}`
* I really don't like non-**explicit** single-argument constructors.
  I have stumbled into a few bugs because of this here already, e.g.:
```c++
  auto nbits = bits(); // decltype(nbits) will be `unsigned`!
  expr rhs_mod_width = rhs.urem(nbits); // compiles, fails at run time
```

Similar to https://github.com/google/souper/pull/406


LebedevRI:
> Questions:
> 
> * Should there be constant-evaluation fastpaths in these `expr::rotate()` / `expr::funnelShift()`?
>   I have avoided adding them for now because the bit width where these computations are
>   done matters, and right now all such `binop_ufold()` implicitly widen to 64-bit.
>   Hmm, but now that i think about it, maybe it would still work, will take a second look..

Ah yes, of course, https://rise4fun.com/Alive/94r it will just work, will only need a mask afterwards.

smt/expr.cpp
```diff
@@ -511,6 +512,59 @@ expr expr::lshr(const expr &rhs) const {
...
+  C(rhs);
+  if (isZero() || isAllOnes() || rhs.isZero())
+    return *this;
+
+  auto nbits = bits();
+  expr rhs_mod_width = rhs.urem(mkUInt(nbits, nbits));
+  if (rhs_mod_width.isZero())
+    return *this;
+
+  // FIXME: could do constant-folding, but *only* for i8/i16/i32/i64.
```
LebedevRI: Hmm, actually, i would love to hear some recommendations about this.
Currently the best i can think of is one of:
* changing `binop_[us]fold` to take `uint64_t (*native)(uint64_t, uint64_t, unsigned /*width*/)` callback.
* or, adding `binop_[us]fold_width` taking that param, and not touching the current `binop_[us]fold`.
* or, changing `binop_[us]fold` to be templated, and take lambda, not a pointer to callback.
(this doesn't cover the need for `ternaryop_[us]fold`..)

I'm not sure which way is most forward.

smt/expr.cpp
```diff
@@ -511,6 +512,59 @@ expr expr::lshr(const expr &rhs) const {
...
+  uint64_t rhs_mod_width_const;
+  if (rhs_mod_width.isUInt(rhs_mod_width_const))
+    return shift_left ? Z3_mk_rotate_left(ctx(), rhs_mod_width_const, ast())
+                      : Z3_mk_rotate_right(ctx(), rhs_mod_width_const, ast());
+
+  return shift_left ? Z3_mk_ext_rotate_left(ctx(), ast(), rhs_mod_width.ast())
+                    : Z3_mk_ext_rotate_right(ctx(), ast(), rhs_mod_width.ast());
+}
+
+expr expr::funnelShift(const expr &a, const expr &b, const expr &c,
```
nunoplopes: I think this is overly complicated. I would split this in fshl/fshr and just do: (a.concat(b) << (c.urem(bits).zext(bits)).extract(...).
Constant folding should happen in concat, urem, shift, etc. (even if it doesn't today for all cases; we can improve those separately)

smt/expr.cpp
```diff
@@ -1,9 +1,10 @@
 // Copyright (c) 2018-present The Alive2 Authors.
 // Distributed under the MIT license that can be found in the LICENSE file.
 
-#include "smt/ctx.h"
 #include "smt/expr.h"
+#include "smt/ctx.h"
 #include <cassert>
+#include <limits>
```
nunoplopes: This doesn't seem to be used

ir/type.cpp
```diff
@@ -223,6 +229,8 @@ void VectorType::fixup(const Model &m) {
   // TODO
 }
 
+expr VectorType::enforceIntOrVectorType() const { return true; }
```
nunoplopes: Please stick to the project's style :)

ir/instr.cpp
```diff
@@ -295,6 +295,71 @@ expr UnaryOp::getTypeConstraints(const Function &f) const {
...
+    break;
+  case FShr:
+    str = "fshr ";
+    break;
+  }
+
+  os << getName() << " = " << str << A << ", " << B << ", " << C;
+}
+
+expr TernaryOp::funnelShiftNotPoison(const expr &a, const expr &ap,
```
nunoplopes: are you sure about the semantics here?
Anything that is not simply "ap && bp && cp" looks very suspicious to me. The reason is that 1) it makes it harder to transform funnel shifts to other instructions (e.g. fshl -> shl) since other instructions don't have conditional poison (only select).

smt/expr.cpp
LebedevRI: Hmm, okay.
And no `rotate` (`Z3_mk_{,ext}_rotate_{left,right}()`) short-circuiting either?

ir/instr.cpp
LebedevRI: I'm not sure, and i thought i have listed that as a question in the PR's description,
but apparently not.
On the first look this *seems* more correct - if we **know** we won't use any of
the bits of the value `a` (or `b`), (and since we know that, we **already** _should_
be able to transform into normal shift), should the fact that that unused value
is `poison` matter?
But i can dumb this down into the usual `ap && bp && cp`,
or start a llvm-dev thread with this question too.

LebedevRI:
@nunoplopes thank you for taking a look!

smt/expr.cpp
nunoplopes: No. The reason is that internally Z3 doesn't really have rotate anyway. So it's better not to use it and use plain shifts. We can add constant-folding to shifts if performance is impacted (overall; not just in these test cases).

LebedevRI: Ok, understood, thank you.

ir/instr.cpp
nunoplopes: From my experience, having short-circuiting definitions of poison doesn't usually work. The reason is what I listed about: converting between instructions then becomes tricky because some instructions will have short-circuit for some cases and others won't, causing a mismatch (and hence forbidding certain transformations).
I would say that if all your test cases go through with the simple definition of poison, stick with it. It's also consistent with the rest of LLVM semantics on poison. Select and phi are the only instructions at the moment that short-circuit poison.

ir/type.cpp
LebedevRI: Ah, you mean the formatting, not the naming.
This is what clang-format gave me by default.
Should the .clang-format be customized then?

LebedevRI: Okay, `.clang-tidy` with
```
---
Language:        Cpp
BasedOnStyle:  LLVM
AllowShortFunctionsOnASingleLine: Empty
...
```
did it.

PR #39, opened 2019-02-24
Support for LLVM IR funnel shift left/right intrinsic

nunoplopes:
LangRef: https://releases.llvm.org/7.0.0/docs/LangRef.html#llvm-fshl-intrinsic
Add support for ‘llvm.fshl’ & ‘llvm.fshr’.
Same as #32, but I screwed up git somehow, so redoing it here..


LebedevRI:
Since there is no mention of those commits in the repo, was there a force-push involved?
Perhaps simply reopening #32 will work.

nunoplopes:
@LebedevRI sorry I screwed up the merge somehow. It was conflicting and then smth went wrong.
Anyway, it should be merged correctly now.
Thanks for your patch!

LebedevRI:
No problem, thank you for the review!

PR #46, opened 2019-04-14
Add support for the rest of overflow instructions

LebedevRI:
There is nothing elaborate here. All the heavy lifting was already done in
https://github.com/AliveToolkit/alive2/pull/28 by @pranavk.
So this is just a repetition of what it took to add SAdd_Overflow support.

Though i suspect the test coverage, while is good, not that great actually.
E.g. there are not tests for TV stuff.
It may be interesting to add code coverage report into CI.


LebedevRI:
@nunoplopes thank you for the review!

PR #47, opened 2019-04-22
add support for disabling poison/undef on input values

pranavk:


ir/value.cpp
```diff
@@ -85,10 +87,12 @@ StateValue Input::toSMT(State &s) const {
   expr undef = expr::mkVar(uname.c_str(), bw);
   s.addUndefVar(undef);
 
-  return { expr::mkIf(type == expr::mkUInt(0, 2),
-                      expr::mkVar(getName().c_str(), bw),
-                      move(undef)),
-           type.extract(1,1) == expr::mkUInt(0, 1) };
+  return {expr::mkIf(config::disable_undef_input ? true
+                                                 : type == expr::mkUInt(0, 2),
+                     expr::mkVar(getName().c_str(), bw), move(undef)),
```
nunoplopes: Things are a bit more complicated than this.
This encoding will confuse the counterexample printer since it assumes that 1x is poison. If poison is disabled, you may have a counterexample with e.g. type=11, which here will mean undef not poison.
Not sure what's the best way to fix, though. We could inject a precondition to ignore types that are disabled, for example. Or make the counterexample printing less hard-coded. (then this variable could be switched to 1-bit when one of the options is disabled)

util/config.h
```diff
@@ -9,4 +9,7 @@ extern bool symexec_print_each_value;
 
 extern bool skip_smt;
 
+extern bool disable_poison_input;
```
nunoplopes: Is this what we want? i.e. a global switch?
Or is there a use case where each input has a potential different behavior?  I can't think of anything right away, just checking your thoughts.

pranavk: For souper, we don't have any use-case as of now for disabling poison/undef for each input. I'd say let's just keep the global switch as of now and then we can add the input-specific support later when we need it.

regehr: I agree we should keep things as simple as possible until a use-case for the other thing emerges

ir/value.cpp
pranavk: i injected couple of preconditions that prevents the type variable now to take the disabled value. See the updated patch, please.

pranavk:
oops! ignore this. will push an updated patch

ir/value.cpp
pranavk: Latest patch should fix this issue. can you please have a look?

tools/alive.cpp
```diff
@@ -19,17 +19,18 @@ using namespace std;
...
-    "Options:\n"
-    " -root-only\tCheck the expression's root only\n"
-    " -v\t\tVerbose mode\n"
-    " -smt-stats\tShow SMT statistics\n"
-    " -smt-to:x\tTimeout for SMT queries in ms\n"
-    " -max-mem:x\tMax memory consumption in MB (aprox)\n"
-    " -smt-verbose\tPrint all SMT queries\n"
-    " -skip-smt\tSkip all SMT queries\n"
-    " -h / --help\tShow this help\n";
+  cerr << "Usage: alive2 <options> <files.opt>\n"
```
nunoplopes: please avoid committing these random changes with the main patch.

tools/transform.cpp
```diff
@@ -129,11 +130,21 @@ static void error(Errors &errs, State &src_state, State &tgt_state,
...
+                       const set<expr> &undef_qvars, expr e) {
+
+  // restrict type variable from taking disabled values
+  for (auto &i : t.src.getInputs()) {
+    auto var = static_cast<const Input &>(i).getTyVar();
+
+    if (config::disable_undef_input)
+      e = e && var != expr::mkUInt(1, 2);
+    if (config::disable_poison_input)
+      e = e && var != expr::mkUInt(2, 2) && var != expr::mkUInt(3, 2);
```
nunoplopes: use &=
!= 3 and != 2 is the same as var.extract(1,1) == 0.

tools/alive.cpp
regehr: yes, @pranavk always review your diffs for stuff like this before submitting

pranavk: ah, this was `clang-format` integrated in my workflow that forced this change; i will fix this.

nunoplopes:
Thanks!
As a perf improvement, if you wish you could simplify the expression at creation time in Input::toSMT() as a follow up patch.

PR #48, opened 2019-05-02
factor out the LLVM->Alive parser so it can be reused in contexts other than TV

regehr:
not sure I'm super happy with how this refactoring came out, but let me know what you think...


regehr:
a few thoughts:
- not sure about the name "llvm utils"
- maybe value_names and friends should go into some sort of context object?
- does the identifiers map want to be cleared also?
- should I put a namespace in llvm-utils? I'm never sure about this...

nunoplopes:
Thanks!
I merged this manually because there were some build issues. I've tested all combinations (TV, llvm but no TV, no TV/llvm) and all work here.
The only technical change was to hide the visitor class from the users. Seems nicer like this.
Let me know if something doesn't look right or doesn't work.

PR #50, opened 2019-05-17
initial cut at a standalone TV tool for bitcode files

regehr:


tools/alive-tv.cpp
```diff
@@ -0,0 +1,135 @@
...
+  std::unique_ptr<llvm::Module> M =
+    ExitOnErr(getOwningLazyBitcodeModule(std::move(MB), Context,
+                                         /*ShouldLazyLoadMetadata=*/true));
+  ExitOnErr(M->materializeAll());
+  return M;
+}
+
+static llvm::Function *getSingleFunction(llvm::Module *M) {
+  llvm::Function *Ret = 0;
+  for (auto &F : *M) {
```
LebedevRI: This only looks at the actual function definition, it ignores the declarations?
I.e. this should be just one function:
https://godbolt.org/z/RhmJqc

tools/alive-tv.cpp
```diff
@@ -0,0 +1,135 @@
...
+#include <string_view>
+#include <vector>
+
+using namespace IR;
+using namespace tools;
+using namespace util;
+using namespace std;
+using namespace llvm_util;
+
+static llvm::cl::opt<std::string>
```
LebedevRI: Would be useful to add these to a new standalone category
```
static llvm::cl::OptionCategory AliveTV("alive-tv options");

static llvm::cl::opt<std::string>
opt_file1(llvm::cl::Positional, llvm::cl::desc("first_bitcode_file"),
    llvm::cl::Required, llvm::cl::value_desc("filename"),
    cl::cat(AliveTV));
```

tools/alive-tv.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+static llvm::ExitOnError ExitOnErr;
+
+// adapted from llvm-dis.cpp
+static std::unique_ptr<llvm::Module> openInputFile(llvm::LLVMContext &Context,
+                                                   std::string InputFilename) {
+  std::unique_ptr<llvm::MemoryBuffer> MB =
+    ExitOnErr(errorOrToExpected(llvm::MemoryBuffer::getFile(InputFilename)));
+  std::unique_ptr<llvm::Module> M =
+    ExitOnErr(getOwningLazyBitcodeModule(std::move(MB), Context,
```
LebedevRI: Hmm, is it too much to ask for it to just natively handle the `.ll`, too (or it does already) ? :)
I kinda expect to be heavily used, much like the `alive` with sudo-ir

tools/alive-tv.cpp
regehr: sorry, I don't understand what you're getting at.
I only want to look at the function definition, and I only want to look at modules containing a single such definition.

tools/alive-tv.cpp
regehr: yeah, I'll do this

tools/alive-tv.cpp
regehr: do you have a pointer to a utility function that handles .bc and .ll? this must exist somewhere...

tools/alive-tv.cpp
LebedevRI: That was the question :)
I was just checking that understood correctly that it won't error-out on https://godbolt.org/z/RhmJqc

regehr: it errors out :)

LebedevRI: Sounds like a bug then :)

regehr: fixed thanks

tools/alive-tv.cpp
regehr: done

tools/alive-tv.cpp
regehr: let's leave this for now, it's not pressing and easy to add later

PR #54, opened 2019-05-23
floating point support

zhengyangl:
The patch now only supports fadd, fsub operators and double, single floating points. However, the basic framework is done and supporting more operators and types would be easy in the future.


ir/constant.h
```diff
@@ -33,6 +34,18 @@ class IntConst final : public Constant {
   auto getInt() const { return std::get_if<int64_t>(&val); }
 };
 
+class FloatConst final : public Constant {
+  std::variant<std::bitset<8>, std::bitset<16>, std::bitset<32>,
+               std::bitset<64>, std::bitset<128>, std::bitset<80>,
```
regehr: ok, I was wondering why you used a bitset instead of a uint32_t and uint64_t, but this makes sense if you're going to need to represent 80 and 128 bit values

regehr: it's not a big deal but you might as well put <80> before <128> in this list

tools/alive_lexer.re
```diff
@@ -104,6 +104,16 @@ re2c:yyfill:check = 0;
   return INT_TYPE;
 }
 
+[-+]? [0-9]* "."? [0-9]+ ([eE] [-+]? [0-9]+)? "d" {
+  yylval.double_num = strtod((char*)YYTEXT, nullptr);
+  return DOUBLE_NUM;
+}
+
+[-+]? [0-9]* "."? [0-9]+ ([eE] [-+]? [0-9]+)? "f" {
```
nunoplopes: it's unclear to me why we need this distinction. If a user wants a specific type, they can write the type f64/f32/f16/whatever, like we do with integers. Can we always parse the number as double?

tests/unit/fp/fadd.opt
```diff
@@ -0,0 +1,40 @@
...
+
+
+Name: fadd identical, type infer, double
+%t = fadd %a, 0.0d
+  =>
+%t = fadd %a, 0.0d
+
+
+Name: fadd identical
+%t = fadd single %a, 0.0f
```
nunoplopes: uhm, I would stick with LLVM's syntax (float rather than single)

smt/expr.h
```diff
@@ -73,8 +73,12 @@ class expr {
   static expr mkUInt(uint64_t n, unsigned bits);
   static expr mkInt(int64_t n, unsigned bits);
   static expr mkInt(const char *n, unsigned bits);
+  static expr mkDouble(double n);
+  static expr mkSingle(float n);
```
nunoplopes: `mkFloat`

smt/expr.cpp
```diff
@@ -1058,6 +1083,7 @@ expr expr::subst(const expr &from, const expr &to) const {
 }
 
 void expr::printUnsigned(ostream &os) const {
+  // TODO: handle fp values
```
nunoplopes: I don't think these functions will be able to support floats, nor is probably desired?

smt/expr.cpp
```diff
@@ -680,6 +696,15 @@ expr expr::ctpop() const {
   return res;
 }
 
+// TODO: make rounding mode customizable
+expr expr::fadd(const expr &rhs) const {
+  return Z3_mk_fpa_add(ctx(), Z3_mk_fpa_round_nearest_ties_to_even(ctx()), ast(), rhs());
```
nunoplopes: these lines are too long

ir/value.cpp
```diff
@@ -79,6 +79,17 @@ void Input::print(std::ostream &os) const {
 }
 
 StateValue Input::toSMT(State &s) const {
+  if (auto fpType = getType().getAsFloatType()) {
+    FpType fpTy = fpType->getFpType();
+    if (fpTy == Single)
```
nunoplopes: may a switch here (that way you don't need the default unreachable case)?

ir/type.h
```diff
@@ -13,6 +13,11 @@ namespace smt { class Model; }
 
 namespace IR {
 
+enum FpType {
```
nunoplopes: can you move this to the FloatType class pls? Also, don't assign to the enum; doesn't seem necessary?

ir/type.cpp
```diff
@@ -222,27 +256,60 @@ expr FloatType::getDummyValue() const {
...
+    os << "quarter";
+    return;
+  case FpType::Half:
+    os << "half";
+    return;
+  case FpType::BFloat16:
+    os << "bfloat16";
+    return;
+  case FpType::Single:
+    os << "single";
```
nunoplopes: float

ir/type.cpp
```diff
@@ -222,27 +256,60 @@ expr FloatType::getDummyValue() const {
...
 }
 
 expr FloatType::sameType(const FloatType &rhs) const {
   return true;
 }
 
 void FloatType::fixup(const Model &m) {
-  // TODO
+  if (!defined)
+    fpType = (FpType) m.getUInt(sizeVar());
```
nunoplopes: please add an assertion to ensure the type belongs to the enum, like in the int class.

ir/type.cpp
```diff
@@ -222,27 +256,60 @@ expr FloatType::getDummyValue() const {
 }
 
 expr FloatType::getTypeConstraints() const {
-  // TODO
-  return false;
+  if (defined) {
+    return true;
+  } else {
```
nunoplopes: no else needed

ir/type.cpp
```diff
@@ -210,10 +218,36 @@ void IntType::print(ostream &os) const {
     os << 'i' << bits();
 }
 
+FpType FloatType::getFpType() const {
```
nunoplopes: you can inline this method in the header

ir/instr.cpp
```diff
@@ -326,6 +337,12 @@ expr BinOp::getTypeConstraints(const Function &f) const {
                   rhs.getType().enforceIntType() &&
                   rhs.getType().sizeVar() == 1u;
     break;
+  case FAdd:
+  case FSub:
+    instrconstr = getType().enforceFloatType() &&
+      getType() == lhs.getType() &&
```
nunoplopes: please follow the indentation style around

ir/instr.cpp
```diff
@@ -282,6 +286,13 @@ StateValue BinOp::toSMT(State &s) const {
     val = structType->extract(a, idx);
     break;
   }
+
+  case FAdd:
+    val = a.fadd(b);
+    break;
+
+  case FSub:
+    val = a.fsub(b);
```
nunoplopes: + break

ir/constant.cpp
```diff
@@ -29,6 +30,21 @@ IntConst::IntConst(Type &type, int64_t val)
...
+FloatConst::FloatConst(Type &type, string &&val)
+  : Constant(type, string(val)), val(move(val)) {}
+
+FloatConst::FloatConst(Type &type, float fval)
+  : Constant(type, to_string(fval)) {
+  uint32_t *c = reinterpret_cast<uint32_t *>(&fval);
+  val = std::bitset<32>(*c);
+}
+
+FloatConst::FloatConst(Type &type, double dval)
```
nunoplopes: please don't interleave methods of different classes

ir/constant.cpp
```diff
@@ -29,6 +30,21 @@ IntConst::IntConst(Type &type, int64_t val)
 IntConst::IntConst(Type &type, string &&val)
   : Constant(type, string(val)), val(move(val)) {}
 
+FloatConst::FloatConst(Type &type, string &&val)
+  : Constant(type, string(val)), val(move(val)) {}
+
+FloatConst::FloatConst(Type &type, float fval)
+  : Constant(type, to_string(fval)) {
+  uint32_t *c = reinterpret_cast<uint32_t *>(&fval);
```
nunoplopes: please don't use reinterpret_cast, but rather a memcpy

ir/constant.cpp
```diff
@@ -45,6 +61,31 @@ expr IntConst::getTypeConstraints() const {
...
+  return true;
+}
+
+pair<expr, expr> FloatConst::toSMT_cnst() const {
+  const FloatType *ft = getType().getAsFloatType();
+  if (ft->getFpType() == FpType::Double) {
+    auto v = get_if<std::bitset<64>>(&val);
+    const auto ival = v->to_ullong();
+    double dval;
+    std::memcpy(&dval, &ival, sizeof(double));
```
nunoplopes: it's quite odd to me why you are converting back and forth from float/doubles to bits. Can we stick with floats/doubles for now and worry about other types when they come up?

ir/constant.cpp
```diff
@@ -45,6 +61,31 @@ expr IntConst::getTypeConstraints() const {
          getType().sizeVar().uge(min_bits);
 }
 
+expr FloatConst::getTypeConstraints() const {
+  return true;
+}
+
+pair<expr, expr> FloatConst::toSMT_cnst() const {
+  const FloatType *ft = getType().getAsFloatType();
+  if (ft->getFpType() == FpType::Double) {
```
nunoplopes: use a switch here

ir/type.h
zhengyangl: @nunoplopes I don't understand the second sentence. Could you please elaborate a bit?

smt/expr.cpp
zhengyangl: No, I'll add pretty printer for fp constants in another function.

tools/alive_lexer.re
zhengyangl: I am thinking about parsing constants some larger FP types like fp80 and fp128. Just like the suffix "l"(long), "ll"(long long) in integer constants, user may reach larger floating point range by this distinction.

Also another possible solution is that we'll always parse the constants with the biggest type "fp128" and cast them to the corresponding type later.

ir/constant.cpp
zhengyangl: What you suggest was my first draft of this code. Generally, it is just a tradeoff between the support of fp128/fp80 and the clearness of the code. I am Okay with both the bitset way and the "double only" way and I'll switch it back if you prefer the "double only" way.

ir/type.h
```diff
@@ -103,14 +99,22 @@ class IntType final : public Type {
...
   FpType fpType;
   bool defined = false;
+
 public:
   FloatType(std::string &&name) : Type(std::move(name)) {}
   FloatType(std::string &&name, FpType fpType)
     : Type(std::move(name)), fpType(fpType), defined(true) {}
   unsigned bits() const override;
-  FpType getFpType() const;
+  inline FpType getFpType() const { return fpType; };
```
nunoplopes: no need for the inline keyword.

ir/type.h
```diff
@@ -103,14 +99,22 @@ class IntType final : public Type {
 
 
 class FloatType final : public Type {
+public:
+  enum FpType {
+    Quarter = 1, Half = 2, BFloat16 = 3,
```
nunoplopes: no need for initializers

ir/constant.cpp
nunoplopes: I would rather have a simple version for now (double), and then we can improve as we go. Right now it seems a bit overengineered, and it also makes it harder to debug since you cannot directly inspect the value.

ir/type.h
zhengyangl: fixed

ir/type.h
zhengyangl: fixed.

ir/constant.cpp
zhengyangl: fixed :-)

ir/constant.cpp
```diff
@@ -5,6 +5,7 @@
 #include "smt/expr.h"
 #include "util/compiler.h"
 #include <cassert>
+#include <cstring>
```
nunoplopes: this isn't needed anymore

ir/constant.cpp
```diff
@@ -45,6 +46,34 @@ expr IntConst::getTypeConstraints() const {
...
+  : Constant(type, string(val)), val(move(val)) {}
+
+FloatConst::FloatConst(Type &type, float val)
+  : Constant(type, to_string(val)), val(val) {}
+
+FloatConst::FloatConst(Type &type, double val)
+  : Constant(type, to_string(val)), val(val) {}
+
+expr FloatConst::getTypeConstraints() const {
+  return true;
```
nunoplopes: this shouldn't be true. see IntConst for example.

ir/constant.h
```diff
@@ -4,6 +4,7 @@
 // Distributed under the MIT license that can be found in the LICENSE file.
 
 #include "ir/value.h"
+#include <bitset>
```
nunoplopes: not needed for now

ir/constant.h
```diff
@@ -33,6 +34,16 @@ class IntConst final : public Constant {
   auto getInt() const { return std::get_if<int64_t>(&val); }
 };
 
+class FloatConst final : public Constant {
+  std::variant<float, double, std::string> val;
+public:
+  FloatConst(Type &type, float val);
+  FloatConst(Type &type, double val);
+  FloatConst(Type &type, std::string &&val);
```
nunoplopes: this path isn't supported in the SMT expr generation, is constructor called somewhere?  If not, pls delete it.

ir/type.cpp
```diff
@@ -222,27 +252,61 @@ expr FloatType::getDummyValue() const {
 }
 
 expr FloatType::getTypeConstraints() const {
-  // TODO
-  return false;
+  if (defined) {
+    return true;
+  }
```
nunoplopes: nitpick: no braces needed

ir/type.cpp
```diff
@@ -635,10 +699,21 @@ expr SymbolicType::enforceAggregateType() const {
...
+expr SymbolicType::enforceFloatType() const {
+  return isFloat();
+}
+
 const StructType* SymbolicType::getAsStructType() const {
   return &s;
 }
 
+const FloatType* SymbolicType::getAsFloatType() const {
+  if (typ == Float)
```
nunoplopes: is this check needed?  why different than getAsStructType()?

smt/expr.cpp
```diff
@@ -185,6 +193,19 @@ expr expr::mkBoolVar(const char *name) {
   return ::mkVar(name, Z3_mk_bool_sort(ctx()));
 }
 
+expr expr::mkFloatVar(const char *name) {
+  return ::mkVar(name, Z3_mk_fpa_sort_single(ctx()));
+}
+
+expr expr::mkDoubleVar(const char*name) {
```
nunoplopes: space

smt/expr.cpp
```diff
@@ -185,6 +193,19 @@ expr expr::mkBoolVar(const char *name) {
...
+  return ::mkVar(name, Z3_mk_fpa_sort_double(ctx()));
+}
+
+expr expr::toBVBool(const expr &e) {
```
nunoplopes: uhm, bad merge here?

smt/expr.cpp
```diff
@@ -675,6 +696,17 @@ expr expr::ctpop() const {
   return res;
 }
 
+// TODO: make rounding mode customizable
```
nunoplopes: not sure we care about customizing rouding mode?

tools/alive_lexer.re
nunoplopes: Ye, I think it should be transparent to the user

ir/type.cpp
zhengyangl: oops, fixed :-)

ir/constant.h
zhengyangl: oops, sorry.

ir/constant.cpp
zhengyangl: fixed.

smt/expr.cpp
zhengyangl: oops, fixed.

ir/type.cpp
zhengyangl: Here I want a class member function for class Type to check the given Type object is in FP type or not and returns a FloatType pointer. This function works similarly to the "dynamic_cast" and later a simple null check is able to tell whether a Type object is a FP.

In other word, I want the function to 
* return itself when it is a FloatType, 
* return member "f" when it is a SymbolicType and holds "typ == Float"
* returns nullptr otherwise. 

If we don't do this check, then SymbolicType objects will always returns a non-null pointer, even when the SymbolicType is not a FP type after fixup. This is not what it supposed to be.

smt/expr.cpp
zhengyangl: I think we should care about supporting non-default RMs, I noticed that LLVM now has a new set of experiment intrinsics called "constrained floating points" which allows user to specify RM by their own, check https://llvm.org/docs/LangRef.html#constrained-floating-point-intrinsics . Also the lifejacket paper also list supporting various RMs as a limitation to their work.

tools/alive_lexer.re
zhengyangl: fixed this :-)

ir/constant.cpp
```diff
@@ -45,6 +45,31 @@ expr IntConst::getTypeConstraints() const {
          getType().sizeVar().uge(min_bits);
 }
 
+FloatConst::FloatConst(Type &type, double val)
+  : Constant(type, to_string(val)), val(val) {}
+
+expr FloatConst::getTypeConstraints() const {
+  return Value::getTypeConstraints() &&
+         getType().enforceFloatType() &&
+         getType().sizeVar().ule(FloatType::FP128) &&
```
nunoplopes: these ule/uge aren't needed. enforceFloatType() already does the job.

ir/type.cpp
```diff
@@ -210,10 +218,32 @@ void IntType::print(ostream &os) const {
     os << 'i' << bits();
 }
 
-
```
nunoplopes: pls don't remove my newlines! :)

ir/type.cpp
```diff
@@ -222,27 +252,60 @@ expr FloatType::getDummyValue() const {
...
 
 expr FloatType::sameType(const FloatType &rhs) const {
   return true;
 }
 
 void FloatType::fixup(const Model &m) {
-  // TODO
+  if (!defined)
+    fpType = (FpType) m.getUInt(sizeVar());
+  assert(fpType >= Quarter && fpType <= FP128);
```
nunoplopes: you need to do the assert with an auxiliary unsigned var. Once you cast to FpType, it's UB if it falls outside the enum, so the assert will just go away.  See e.g. `SymbolicType::fixup()`.

ir/type.cpp
```diff
@@ -222,27 +252,60 @@ expr FloatType::getDummyValue() const {
...
+    os << "double";
+    return;
+  case FpType::IEEE80:
+    os << "ieee80";
+    return;
+  case FpType::FP128:
+    os << "fp128";
+    return;
+  }
+  UNREACHABLE();
```
nunoplopes: not needed; all compilers will warn if switch doesn't cover all options

ir/type.cpp
```diff
@@ -222,27 +252,60 @@ expr FloatType::getDummyValue() const {
...
+    fpType = (FpType) m.getUInt(sizeVar());
+  assert(fpType >= Quarter && fpType <= FP128);
 }
 
 void FloatType::print(ostream &os) const {
-  os << "TODO";
+  switch (fpType) {
+  case FpType::Quarter:
+    os << "quarter";
+    return;
```
nunoplopes: I would use break; here. Most compilers prefer a single exit.

ir/type.cpp
nunoplopes: ok for now. The input functionality needs to be redesigned. I'll clean this up later.

ir/type.cpp
zhengyangl: Oh, its interesting to know!

smt/expr.h
```diff
@@ -73,8 +73,13 @@ class expr {
   static expr mkUInt(uint64_t n, unsigned bits);
   static expr mkInt(int64_t n, unsigned bits);
   static expr mkInt(const char *n, unsigned bits);
+  static expr mkDouble(double n);
+  static expr mkFloat(float n);
   static expr mkVar(const char *name, unsigned bits);
   static expr mkBoolVar(const char *name);
+  static expr mkFloatVar(const char *name);
+  static expr mkDoubleVar(const char *name);
+  static expr toBVBool(const expr &e);
```
nunoplopes: bad merge here

zhengyangl: Oops, sorry.

PR #57, opened 2019-05-29
clear error message that accesses to globals aren't yet supported

regehr:


llvm_util/utils.cpp
```diff
@@ -141,6 +141,11 @@ Value* get_operand(llvm::Value *v) {
     return ret;
   }
 
+  if (auto ptr = dyn_cast<llvm::GlobalVariable>(v)) {
```
nunoplopes: this won't compile on gcc because ptr isn't used and we build with `-Werror` :)

LebedevRI:
```suggestion
  if (isa<llvm::GlobalVariable>(v)) {
```
?

PR #73, opened 2019-07-02
Added parser support for fmul, fdiv, added test cases for these operators

zhengyangl:


tests/unit/fp/fmul.opt
```diff
@@ -0,0 +1,30 @@
...
+Name: x * 2.0 * 3.125 => × * 6.25
+%t1 = fmul half %a, 2.0
+%t = fmul half %t1, 3.125
+  =>
+%t1 = fmul half %a, 2.0
+%t = fmul half %a, 6.25
+
+; TODO: reassoc
+Name: x * 3.125 * 2.0 => × * 6.25
+; ERROR: Value mismatch for half %t
```
nunoplopes: please don't mix positive and negative tests in the same file, otherwise we don't know which test failed (if the one it was supposed to fail or some other). This is a limitation of the current test system.

nunoplopes:
Can you rebase this branch so it picks up the update to the Travis config pls?  It may help getting rid of the build failure

tests/unit/fp/fmul.opt
zhengyangl: Gotcha! Fixed it.

zhengyangl:
Still got 1 failure on OSX, should I disable the test case?

nunoplopes:
> Still got 1 failure on OSX :(, should I disable the test case?

It seems all bots are now failing on the same test. At least it's deterministic now :)

zhengyangl:
Finally passed travis :)

PR #76, opened 2019-07-11
Support of the fast math flags NNaN and NInf

zhengyangl:
WIP, not working yet, comments are welcome


ir/instr.cpp
```diff
@@ -277,18 +280,34 @@ StateValue BinOp::toSMT(State &s) const {
 
   case FAdd:
     val = a.fadd(b);
+    if (flags & NINF)
+      not_poison &= !(a.isInf()) && !(b.isInf()) && !(val.isInf());
```
nunoplopes: no parenthesis needed

ir/value.cpp
```diff
@@ -65,7 +65,8 @@ void PoisonValue::print(ostream &os) const {
 }
 
 StateValue PoisonValue::toSMT(State &s) const {
-  return { expr::mkUInt(0, bits()), false };
+  expr var = getType().getDummyValue();
+  return { move(var), false };
```
nunoplopes: good catch!
Can be formatted in 1 line only, though.

ir/instr.cpp
```diff
@@ -277,18 +280,34 @@ StateValue BinOp::toSMT(State &s) const {
...
+    if (flags & NNAN)
+      not_poison &= !(a.isNaN()) && !(b.isNaN()) && !(val.isNaN());
```
nunoplopes: also, this code probably wants to be moved to a function since there's a lot of duplication here and there are many more fast-math flags to be supported

nunoplopes:
There's a conflict; can't merge without it being fixed first.

zhengyangl:
@nunoplopes rebased to the latest upstream. However, there are still bugs when doing the translations on poison and undef for FP sorts (check the failing case in freeze.opt) and I'm trying to fix. If you can have a look that would be better :)

zhengyangl:
@nunoplopes Hi Nuno, can you review the last commit I made :)

PR #84, opened 2019-08-02
Support for Vector

zhengyangl:
WIP, comments are welcome!


zhengyangl:
Since llvm extractelement operator support variable indexes, it is a little bit tricky to implement the operator if the vectors are represented in bitvectors, since bitvector extract only takes known values for the  arguments, check https://z3prover.github.io/api/html/group__capi.html#ga41827fdb3fa65f518f0ae29c1d900ce6 .

So I rewrite the vector with the Sequences, the Z3 nth operator makes the llvm extractelement trivial. Till now the easiest test cases are passing. But note we need to manually patch this recent Z3 bugfix https://github.com/Z3Prover/z3/commit/bbfac99b22949b5ac1d7892e74036fefb49d1ba7 to make nth work.

nunoplopes:
extract of non-constant indexes can be implemented in BV with shifts + extract. It's not very complicated. I would rather use that solution than use yet another theory (sequences).

BTW, can you please split this PR into multiple ones, one per change?  This PR grew too big and it's trying to do too much stuff at once, so it will never go in (or it will take infinite time).
Introducing vectors is non-trivial and there are many aspects that need to be discussed. So it's better to do it one by one.
We can start with the type system, for example. Then 1 operation or parsing, etc.
Thanks!

PR #86, opened 2019-08-20
Support for Vector: type checker

zhengyangl:


ir/type.cpp
```diff
@@ -704,12 +751,12 @@ expr SymbolicType::getDummyValue() const {
...
-  c |= isFloat()  && f.getTypeConstraints();
-  c |= isPtr()    && p.getTypeConstraints();
-  c |= isArray()  && a.getTypeConstraints();
+  c |= isInt()    && i.getTypeConstraints() && (v.getElementTy().sizeVar() == 0);
+  c |= isFloat()  && f.getTypeConstraints() && (v.getElementTy().sizeVar() == 0);
+  c |= isPtr()    && p.getTypeConstraints() && (v.getElementTy().sizeVar() == 0);
+  c |= isArray()  && a.getTypeConstraints() && (v.getElementTy().sizeVar() == 0);
   c |= isVector() && v.getTypeConstraints();
-  c |= isStruct() && s.getTypeConstraints();
+  c |= isStruct() && s.getTypeConstraints() && (v.getElementTy().sizeVar() == 0);
```
zhengyangl: symbolic vector brings a new free variable, this set the constraints on it when reasoning non-vector symbolic types. This looks ugly, though. @nunoplopes do you have comments for this?

zhengyangl:
Hi @nunoplopes , as you suggested, I am now sending pull requests in smaller ones. This pull request covers the VectorType class and implements the type reasoning for Vector types. 

It uses a symbolic variable sizeVar() to store the length of the vector. When doing initialization of a symbolic vector, a symbolic IntType is created and assigned as its child type. The sizeVar() of vector, as well as the sizeVar() of its child type, can be reasoned in solver and thus we can obtain the concrete type of the vector.

ir/type.cpp
```diff
@@ -482,57 +491,95 @@ void ArrayType::print(ostream &os) const {
...
 expr VectorType::enforceIntOrVectorType() const {
   // TODO: check if elements are int
-  return false;
+  return true;
 }
 
 expr VectorType::enforceIntOrPtrOrVectorType() const {
   // TODO: check if elements are int/ptr
-  return false;
+  return true;
```
nunoplopes: this needs to check that the element type is int or ptr (same above)

ir/type.cpp
```diff
@@ -482,57 +491,95 @@ void ArrayType::print(ostream &os) const {
...
 unsigned VectorType::bits() const {
   // TODO
   return 0;
 }
 
 expr VectorType::getDummyValue() const {
-  // TODO
-  return {};
+  expr res;
+  for (unsigned idx = 0; idx < length ; idx++) {
```
nunoplopes: ++idx

ir/type.cpp
```diff
@@ -482,57 +491,95 @@ void ArrayType::print(ostream &os) const {
   os << "TODO";
 }
 
+static std::vector<std::unique_ptr<IntType>> child_types;
+static unsigned child_ty_num;
```
nunoplopes: no global variables here pls.

ir/type.cpp
```diff
@@ -482,57 +491,95 @@ void ArrayType::print(ostream &os) const {
...
+
+ // TODO: Support symbolic vector on floating point types
+VectorType::VectorType(std::string &&name) : Type(std::move(name)) {
+  elementTy = child_types.emplace_back(
+    std::make_unique<IntType>("child_" + std::to_string(child_ty_num++))).get();
```
nunoplopes: it should receive the element type as input. Even if it's a symbolic type, for example.

ir/type.cpp
```diff
@@ -482,57 +491,95 @@ void ArrayType::print(ostream &os) const {
...
+  if (!defined)
+    r &= bw.ule(expr::mkUInt(var_vec_length_bits, var_bw_bits));
+
+  return (r && elementTy->getTypeConstraints());
 }
 
 expr VectorType::operator==(const VectorType &rhs) const {
-  // TODO
-  return false;
+  return (sizeVar() == rhs.sizeVar()) && (getElementTy() == rhs.getElementTy());
```
nunoplopes: remove (), not needed

ir/type.cpp
nunoplopes: Right, please don't do this :)
There's no need for these constraints. In the worst case we'll finally need the model minimizer to avoid going over equivalent models of types. I had to disable some stuff in pointers as well already. I'll look into implementing that.

ir/type.cpp
zhengyangl: Dear @nunoplopes , I don't catch your comment, what do you mean by "receive element types as input", does it mean to drop this constructor?  We do have another constructor with takes element type as its argument. 

Also allowing vector of symbolic types may leads to infinite recursions, like vector of vector of endless of vectors.... I am thinking of creating a new ScalarType class. IntType, FloatType and Ptr types are inherited from the new class and VectorType can only produce vector on scalars.

nunoplopes:
I've implemented it myself, since the changes in SymbolicType were quite involved (and needed for arrays & structs as wells). See here: https://github.com/AliveToolkit/alive2/commit/8f4e2b086461c8996b7df886fcd120e415e4d7b3
I haven't tested the code since there are no callers.
Closing this PR since most of functionality is now implemented. Please open other PRs for all the remaining stuff :)

PR #88, opened 2019-08-23
add memcpy to alive2

Dongjoo-Kim:
1) Add memcpy's SMT encoding(in memory.cpp).

2) Make Alive2 interpret memset and memcpy correctly(in instr.cpp, instr.h, utils.cpp).
   
    Previous Alive2 doesn't interpret memset and memcpy, because there is no visit function for them.
    So basic tests in 'test/memcpy/' made 'LLVM error: Unsupported instruction'.
    I fixed it, as a result basic tests which I made are successfully verified by Alive2.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1502,51 +1502,72 @@ unique_ptr<Instr> Store::dup(const string &suffix) const {
 }
 
 
-void Memset::print(std::ostream &os) const {
-  os << "memset " << ptr << " align " << align << ", " << val << ", " << bytes;
+vector<Value*> Memset::operands() const {
+  return { ptr, val, bytes };
+}
+
+void Memset::rauw(const Value &what, Value &with) {
+  RAUW(ptr);
+  RAUW(val);
+  RAUW(bytes);
+}
+
+void Memset::print(ostream &os) const {
+  os << "memset " << *ptr << " align " << align << ", " << *val
+     << ", " << *bytes;
 }
 
 StateValue Memset::toSMT(State &s) const {
-  // TODO: check following lines are correct
-  auto &[vptr, np] = s[ptr];
-  auto &[vbytes, np2] = s[bytes];
-  s.addUB(np);
-  s.addUB(np2);
-  s.getMemory().memset(vptr, s[val], vbytes, align);
+  auto &[vptr, np] = s[*ptr];
+  auto &[vbytes, np2] = s[*bytes];
+  s.addUB(vbytes.ugt(0).implies(np));
+  s.addUB(vbytes.ugt(0).implies(np2));
+  s.getMemory().memset(vptr, s[*val], vbytes, align);
   return {};
 }
 
 expr Memset::getTypeConstraints(const Function &f) const {
-  return ptr.getType().enforcePtrType() && bytes.getType().enforceIntType();
+  return ptr->getType().enforcePtrType() && bytes->getType().enforceIntType();
 }
 
 unique_ptr<Instr> Memset::dup(const string &suffix) const {
-  return make_unique<Memset>(ptr, val, bytes, align);
+  return make_unique<Memset>(*ptr, *val, *bytes, align);
 }
 
 
-void Memcpy::print(std::ostream &os) const {
-  os << "memcpy " << dst  << " align " << align_dst << ", " << src  << " align " << align_src << ", " << bytes;
+vector<Value*> Memcpy::operands() const {
+  return { dst, src, bytes };
+}
+
+void Memcpy::rauw(const Value &what, Value &with) {
+  RAUW(dst);
+  RAUW(src);
+  RAUW(bytes);
+}
+
+void Memcpy::print(ostream &os) const {
+  os << "memcpy " << *dst  << " align " << align_dst << ", " << *src
+     << " align " << align_src << ", " << *bytes;
 }
 
 StateValue Memcpy::toSMT(State &s) const {
-  // TODO: check following lines are correct
-  auto &[vdst, np] = s[dst];
-  s.addUB(np);
-  auto &[vsrc, np2] = s[src];
-  s.addUB(np2);
-  auto &[vbytes, np3] = s[bytes];
-  s.addUB(np3);
+  auto &[vdst, np] = s[*dst];
+  auto &[vsrc, np2] = s[*src];
+  auto &[vbytes, np3] = s[*bytes];
+  s.addUB(vbytes.ugt(0).implies(np));
+  s.addUB(vbytes.ugt(0).implies(np2));
+  s.addUB(vbytes.ugt(0).implies(np3));
   s.getMemory().memcpy(vdst, vsrc, vbytes, align_dst, align_src);
   return {};
 }
 
 expr Memcpy::getTypeConstraints(const Function &f) const {
-  return dst.getType().enforcePtrType() && dst.getType().enforcePtrType() && bytes.getType().enforceIntType();
+  return dst->getType().enforcePtrType() && dst->getType().enforcePtrType()
+         && bytes->getType().enforceIntType();
 }
 
 unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
-  return make_unique<Memcpy>(dst, src, bytes, align_dst, align_src);
+  return make_unique<Memcpy>(*dst, *src, *bytes, align_dst, align_src);
 }
 
 }
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1518,16 +1518,17 @@ void Memset::print(ostream &os) const {
 }
 
 StateValue Memset::toSMT(State &s) const {
-  auto &[vptr, np] = s[*ptr];
-  auto &[vbytes, np2] = s[*bytes];
-  s.addUB(vbytes.ugt(0).implies(np));
-  s.addUB(vbytes.ugt(0).implies(np2));
+  auto &[vptr, np_ptr] = s[*ptr];
+  auto &[vbytes, np_bytes] = s[*bytes];
+  s.addUB(vbytes.ugt(0).implies(np_ptr));
+  s.addUB(np_bytes);
   s.getMemory().memset(vptr, s[*val], vbytes, align);
   return {};
 }
 
 expr Memset::getTypeConstraints(const Function &f) const {
-  return ptr->getType().enforcePtrType() && bytes->getType().enforceIntType();
+  return ptr->getType().enforcePtrType() &&
+         bytes->getType().enforceIntType();
 }
 
 unique_ptr<Instr> Memset::dup(const string &suffix) const {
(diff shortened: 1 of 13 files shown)


llvm_util/utils.cpp
```diff
@@ -335,6 +335,44 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
+    // TODO: add isvolatile, alignment
+    if (!ptr || !val || !bytes)
+      return error(i);
+
+    RETURN_IDENTIFIER(make_unique<Memset>(*ptr, *val, *bytes, 1));
+  }
+
+  RetTy visitMemCpyInst(llvm::MemCpyInst &i) {
+    auto fn = i.getCalledFunction();
+    if (!fn) // TODO: check real memcpy
```
nunoplopes: this error checking isn't needed. The intrinsic should be valid

tests/memcpy/1-src.ll
```diff
@@ -0,0 +1,12 @@
+declare void @llvm.memcpy.p0i8.p0i8.i32(i8*, i8*, i32, i1)
+
+define i8 @f() {
+  %p1 = alloca [16 x i8]
```
nunoplopes: Have you run these tests with e.g. llvm-tv?  I was fairly sure that arrays aren't implemented yet

ir/instr.h
```diff
@@ -381,4 +381,32 @@ class Store final : public Instr {
...
+  std::unique_ptr<Instr> dup(const std::string &suffix) const override;
+};
+
+
+class Memcpy final : public Instr {
+  Value &dst, &src, &bytes;
+  unsigned align_dst, align_src;
+public:
+  Memcpy(Value &dst, Value &src, Value &bytes, unsigned align_dst, unsigned align_src)
+    : Instr(Type::voidTy, "memcpy"), dst(dst), src(src), bytes(bytes), align_dst(align_dst), align_src(align_src) {}
```
nunoplopes: these lines are too long. Please wrap at 80 columns

ir/memory.cpp
```diff
@@ -304,10 +312,27 @@ void Memory::memset(const expr &p, const StateValue &val, const expr &bytes,
 
 void Memory::memcpy(const expr &d, const expr &s, const expr &bytes,
                     unsigned align_dst, unsigned align_src) {
+  // TODO: Add ub condition - bytes == 0
```
nunoplopes: this requites some API changes. Either address it in this PR or in a subsequent patch.

ir/memory.cpp
```diff
@@ -149,6 +149,14 @@ void Pointer::is_dereferenceable(unsigned bytes, unsigned align) {
   is_dereferenceable(expr::mkUInt(bytes, m.bits_for_offset), align);
 }
 
+expr disjoint(expr offset1, const expr &len1, expr offset2, const expr &len2) {
```
nunoplopes: should be static

ir/memory.cpp
```diff
@@ -149,6 +149,14 @@ void Pointer::is_dereferenceable(unsigned bytes, unsigned align) {
...
+  return ((offset1+len1).ule(offset2) || (offset2+len2).ule(offset1));
```
nunoplopes: this has too many parenthesis

nunoplopes: Is this condition correct if offset + len overflows?  Or can't that happen (because of the half of the address space limitation)?  If that's the case, we need a comment there explaining the assumption/reasoning.

tests/memcpy/1-src.ll
Dongjoo-Kim: I ran the tests using alive-tv and their results are as expected.

I think that Alive2 encodes the code to alloca n bytes.

For example, 'alloca [16 x i8]' is encodded 'alloca 16 bytes'.
And 'alloca i2' is encode 'alloca 1 bytes'.

ir/memory.cpp
Dongjoo-Kim: I'm not sure that when 'bytes' is zero and 'src (or tgt) pointer' is poison, the code has ub or not.

I am going to address similar cases in a subsequent patch.

ir/memory.cpp
Dongjoo-Kim: When I use is_disjoint in memcpy, overflow checking is run in is_dereferenceable.

If we need this condition at other code, overflow condition is considered.

I add a comment in the code also.

tests/memcpy/1-src.ll
nunoplopes: ok, we (well, I) need to extend the unit test runner to run alive-tv. I'll do that later this week.

ir/instr.cpp
```diff
@@ -1280,4 +1280,52 @@ unique_ptr<Instr> Store::dup(const string &suffix) const {
   return make_unique<Store>(ptr, val, align);
 }
 
+
+void Memset::print(std::ostream &os) const {
+  os << "memset " << ptr << " align " << align << ", " << val << ", " << bytes;
+}
+
+StateValue Memset::toSMT(State &s) const {
+  // TODO: check following lines are correct
```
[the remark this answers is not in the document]
Dongjoo-Kim: I modified memset, memcpy and is_dereferenceable using your conditions.
is_disjoint doesn't need modifying because when bytes is 0, is_disjoint is already no ub.

ir/instr.cpp
```diff
@@ -1495,4 +1495,73 @@ unique_ptr<Instr> Store::dup(const string &suffix) const {
...
+void Memcpy::print(ostream &os) const {
+  os << "memcpy " << *dst  << " align " << align_dst << ", " << *src
+     << " align " << align_src << ", " << *bytes;
+}
+
+StateValue Memcpy::toSMT(State &s) const {
+  auto &[vdst, np] = s[*dst];
+  auto &[vsrc, np2] = s[*src];
+  auto &[vbytes, np3] = s[*bytes];
+  s.addUB(vbytes.ugt(0).implies(np));
```
nunoplopes: please combine the two checks into one: ".implies(np && np2)"

ir/instr.cpp
```diff
@@ -1495,4 +1495,73 @@ unique_ptr<Instr> Store::dup(const string &suffix) const {
...
+void Memset::print(ostream &os) const {
+  os << "memset " << *ptr << " align " << align << ", " << *val
+     << ", " << *bytes;
+}
+
+StateValue Memset::toSMT(State &s) const {
+  auto &[vptr, np] = s[*ptr];
+  auto &[vbytes, np2] = s[*bytes];
+  s.addUB(vbytes.ugt(0).implies(np));
+  s.addUB(vbytes.ugt(0).implies(np2));
```
nunoplopes: this second line should be simply "s.addUB(np2)".  If vbytes is poison, the copy is UB since the value to copy is unbounded.

ir/instr.cpp
```diff
@@ -1495,4 +1495,73 @@ unique_ptr<Instr> Store::dup(const string &suffix) const {
...
+     << " align " << align_src << ", " << *bytes;
+}
+
+StateValue Memcpy::toSMT(State &s) const {
+  auto &[vdst, np] = s[*dst];
+  auto &[vsrc, np2] = s[*src];
+  auto &[vbytes, np3] = s[*bytes];
+  s.addUB(vbytes.ugt(0).implies(np));
+  s.addUB(vbytes.ugt(0).implies(np2));
+  s.addUB(vbytes.ugt(0).implies(np3));
```
nunoplopes: same comment as for memset: should be addUB(np3) only.
Also, please rename vars to np_dst, np_src, or something like that. np2/np3 are error prone :)

ir/instr.cpp
```diff
@@ -1495,4 +1495,73 @@ unique_ptr<Instr> Store::dup(const string &suffix) const {
...
+  s.getMemory().memcpy(vdst, vsrc, vbytes, align_dst, align_src);
+  return {};
+}
+
+expr Memcpy::getTypeConstraints(const Function &f) const {
+  return dst->getType().enforcePtrType() && dst->getType().enforcePtrType()
```
nunoplopes: please do this like:
```
return enforcefoo() &&
       enforceBar &&
       etc;
```

Dongjoo-Kim:
I reflected your comments and pushed it.

ir/memory.h
```diff
@@ -68,6 +68,7 @@ class Pointer {
   smt::expr is_aligned(unsigned align) const;
   void is_dereferenceable(unsigned bytes, unsigned align);
   void is_dereferenceable(const smt::expr &bytes, unsigned align);
+  void is_disjoint(const smt::expr &len1, const Pointer &ptr2, const smt::expr &len2) const;
```
nunoplopes: this is over 80 characters

ir/memory.cpp
```diff
@@ -149,6 +151,20 @@ void Pointer::is_dereferenceable(unsigned bytes, unsigned align) {
   is_dereferenceable(expr::mkUInt(bytes, m.bits_for_offset), align);
 }
 
+// general disjoint check for unsigned integer
+static expr disjoint(expr begin1, const expr &len1, expr begin2,
```
nunoplopes: typo: begin1 & begin2 should be const references as well.

ir/memory.cpp
```diff
@@ -149,6 +151,20 @@ void Pointer::is_dereferenceable(unsigned bytes, unsigned align) {
   is_dereferenceable(expr::mkUInt(bytes, m.bits_for_offset), align);
 }
 
+// general disjoint check for unsigned integer
```
nunoplopes: Please add a comment saying that this function assumes that begin1+len1 and begin2+len2 don't overflow (unsigned)

ir/memory.cpp
```diff
@@ -307,7 +323,24 @@ void Memory::memcpy(const expr &d, const expr &s, const expr &bytes,
...
+      auto src_i = (src + i).release();
+      auto dst_i = (dst + i).release();
+      blocks_val = blocks_val.store(dst_i, blocks_val.load(src_i));
+    }
+  } else {
+    string name = "#idx_" + to_string(last_idx_ptr++);
+    Pointer idx(*this, expr::mkVar(name.c_str(), dst.bits()));
+
+    expr cond = idx.uge(dst).both() && idx.ult(dst + bytes).both();
+    expr val = expr::mkIf(cond, blocks_val.load((src + idx.get_offset()).release()),
```
nunoplopes: this lines seems over 80 characters

nunoplopes: not sure the condition for the src load is correct. It assumes that offset of src = 0.
Remember that memcpy both src and dst may start in the middle of an object.
So we need src + "delta between start of copy and idx". Which may not be `idx.get_offset()` if the memcpy wasn't done the start of the object.
What do you think?

Dongjoo-Kim: Thank you for your comment. I fixed this bug.

Dongjoo-Kim:
Sorry for late response. I reflect your comments.

PR #89, opened 2019-08-25
Add Z3_LIBRARIES to alive-tv's dependency

aqjune:
This resolves linking error when compiling Alive2 at Mac.

Tested at ubuntu 18.04.3 and Mac 10.14.6

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,7 +142,7 @@ add_library(alive2 SHARED ${IR_SRCS} ${SMT_SRCS} ${TOOLS_SRCS} ${UTIL_SRCS} ${LL
 if (BUILD_LLVM_UTILS OR BUILD_TV)
   llvm_map_components_to_libnames(llvm_libs support core irreader)
   target_link_libraries(alive2 PRIVATE ${llvm_libs})
-  target_link_libraries(alive-tv PRIVATE ${ALIVE_LIBS_LLVM} ${llvm_libs})
+  target_link_libraries(alive-tv PRIVATE ${ALIVE_LIBS_LLVM} ${llvm_libs} ${Z3_LIBRARIES})
 endif()
 
 if (CYGWIN)


CMakeLists.txt
```diff
@@ -142,7 +142,7 @@ add_library(alive2 SHARED ${IR_SRCS} ${SMT_SRCS} ${TOOLS_SRCS} ${UTIL_SRCS} ${LL
 if (BUILD_LLVM_UTILS OR BUILD_TV)
   llvm_map_components_to_libnames(llvm_libs support core irreader)
   target_link_libraries(alive2 PRIVATE ${llvm_libs})
-  target_link_libraries(alive-tv PRIVATE ${ALIVE_LIBS_LLVM} ${llvm_libs})
+  target_link_libraries(alive-tv PRIVATE ${ALIVE_LIBS_LLVM} ${llvm_libs} ${Z3_LIBRARIES})
```
LebedevRI: This seems weird.
`alive-tv` executable is being linked to `alive2` library, right?
Why shouldn't `Z3_LIBRARIES` be specified for `alive2` itself?

nunoplopes: Actually no, it's not linked against the `alive2`, but it rather `${ALIVE_LIBS_LLVM}`. Linking against `alive2` directly doesn't work; dunno why.
I'm actually surprised Z3 gets linked it on all my computers. This patch won't work on Cygwin though. Juneyoung, please see the code a few lines below re Cygwin.

aqjune:
Sure, I will check at possibly 7 pm today (in CA timezone) :-)

2019년 8월 25일 (일) 오전 11:48, Nuno Lopes <notifications@github.com>님이 작성:

CMakeLists.txt
LebedevRI: Aha, that is still the same symptom.

Does `find_package(Z3 4.8.4 REQUIRED)` already provide imported target?
It could be `Z3` or `Z3::Z3`.

If not, i think you want to create it here, move that `if (CYGWIN)` block up
and change it to work on said `Z3::Z3`, and finally change stuff to **unconditionally**
`target_link_libraries(??? PRIVATE Z3::Z3)`

nunoplopes: FindZ3.cmake is in the repo. It only sets Z3_INCLUDE_DIR & Z3_LIBRARIES.
I'm no expert in cmake. Roman, can you elaborate a bit more on what needs to be done pls? Thanks :)

LebedevRI: > FindZ3.cmake is in the repo.

Right. I'll try to just come up with a patch then.

aqjune: My reply to the mail didn't go to this thread :/
If @LebedevRI 's upcoming patch resolves underlying problem, it is okay for me to use the patch instead of this as well.

LebedevRI:
Close this?

PR #91, opened 2019-08-27
Support for Vector: VectorValue

zhengyangl:


ir/value.cpp
```diff
@@ -110,4 +111,37 @@ expr Input::getTyVar() const {
   return expr::mkVar(tyname.c_str(), 2);
 }
 
+
+static string genVectorName(const std::vector<Value *> &vals) {
+  ostringstream ss;
+  ss << "<";
+  for (unsigned idx = 0; idx < vals.size(); ++idx) {
+    ss << *vals[idx];
+    if (idx != vals.size() -1)
```
nunoplopes: it's easier to check for idx != 0 and pre-add a comma

ir/value.cpp
```diff
@@ -110,4 +111,37 @@ expr Input::getTyVar() const {
   return expr::mkVar(tyname.c_str(), 2);
 }
 
+
+static string genVectorName(const std::vector<Value *> &vals) {
```
nunoplopes: we don't use std:: in .cpp files

ir/value.cpp
```diff
@@ -110,4 +111,37 @@ expr Input::getTyVar() const {
...
+  UNREACHABLE();
+}
+
+StateValue VectorValue::toSMT(State &s) const {
+  expr val;
+  expr not_poison = true;
+
+  for (unsigned idx = 0; idx < vals.size(); ++idx) {
+    auto [v, vs] = vals[idx]->toSMT(s);
+    val = idx == 0 ? v : val.concat(v);
```
nunoplopes: float values need to be cast to bit-vectors.
I would suggest we introduce a new function in the VectorType that creates an SMT value from a vector of individual values (to avoid repeating this code to cast to BV, as it will probably needed later).

ir/value.cpp
```diff
@@ -110,4 +111,37 @@ expr Input::getTyVar() const {
...
+    not_poison &= vs;
```
nunoplopes: poison is per element, so we need concat here as well. I think the function just mention can take a vector of StateValues and return a StateValue

ir/value.cpp
```diff
@@ -110,4 +111,37 @@ expr Input::getTyVar() const {
...
+  }
+
+  return { move(val), true };
```
nunoplopes: true is certainly not correct here :)

nunoplopes:
Today I had to do quite a few changes related with vectors (and aggregate types in general) in order to support checking of some miscompilation in LLVM we were chasing. In the process I did a large refactoring of the aggregate types (arrays/structs/vectors) to reduce code duplication.
Could you please have a look?  This may have clashed with this patch, I'm sorry.

zhengyangl:
Sure, thank you @nunoplopes for the review and the notify of this potential clash. I will be a little occupied before the CGO deadline (this Friday). I will take care of all the reviews afterwards.

ir/value.cpp
zhengyangl: fixed :)

ir/value.cpp
zhengyangl: fixed :)

ir/value.cpp
zhengyangl: Hi @nunoplopes , I've fixed this and If you're okay with the code I'll send more code for review (parser and vector operations).

ir/value.cpp
zhengyangl: oops:)

ir/type.cpp
```diff
@@ -664,6 +668,20 @@ VectorType::VectorType(string &&name, unsigned elements, Type &elementTy)
...
 
+StateValue VectorType::aggregateVals(const vector<StateValue> &vals) const {
+  assert(defined);
+  assert(vals.size() == elements);
+
+  expr val;
+  expr not_poison;
+  for (unsigned idx = 0; idx < elements; ++idx) {
+    auto [v, vs] = children[idx]->toBV(vals[idx]);
+    val = idx == 0 ? v : val.concat(v);
```
nunoplopes: StateValue now supports concat. You can do this in one go :)
See AggregateConst::toSMT().  Actually that code should be replaced with a call to this.

nunoplopes: BTW, I think this method can be hoisted to the AggregateType class. The extract is also there.

ir/value.h
```diff
@@ -83,4 +83,12 @@ class Input final : public Value {
   smt::expr getTyVar() const;
 };
 
+class VectorValue final : public Value {
```
nunoplopes: Is this class still needed?  I added a ConstantAggregate class that should be sufficient for vectors, structs and arrays.

ir/type.cpp
zhengyangl: Fixed, these makes code extremely clean!

ir/value.h
zhengyangl: Good point, I've removed this function.

zhengyangl:
@nunoplopes This PR is ready for review again:) I'll send some real caller to these updates in the next pull request!

ir/type.h
```diff
@@ -229,6 +229,7 @@ class AggregateType : public Type {
     smt::expr(*map)(const IR::StateValue&, const IR::StateValue&),
     smt::expr(*reduce)(const std::set<smt::expr>&),
     const IR::StateValue &a, const IR::StateValue &b) const override;
+  StateValue aggregateVals(const std::vector<StateValue> &vals) const;
```
nunoplopes: seems like a tab here?

ir/type.cpp
```diff
@@ -637,6 +637,18 @@ expr AggregateType::map_reduce(expr(*map)(const StateValue&, const StateValue&),
   return reduce(r);
 }
 
+StateValue AggregateType::aggregateVals(const vector<StateValue> &vals) const {
+  assert(defined);
```
nunoplopes: this assert isn't valid. `defined` is used for types that are fully defined (nothing symbolic).
So if the type is inferred, `defined` will be false, though the type will be then fine for consumption.
TL;DR: just remove this assert :)

ir/type.h
zhengyangl: No my emacs would never produce evil tab :), see the above lines for the indentation.

ir/type.cpp
zhengyangl: fixed :)

PR #93, opened 2019-09-03
Implement Memory::free, let Alive2 parser read malloc/free

aqjune:
This implements
- lifetime of a memory block by defining an array which stores liveness bit per block id.
If the bit is 1, the block is alive. Otherwise, the block is freed.
- Also, a kind of memory block (heap or non-heap) in a similar manner.


ir/instr.h
```diff
@@ -364,11 +364,25 @@ class Alloc final : public Instr {
...
+  std::vector<Value*> operands() const override;
+  void rauw(const Value &what, Value &with) override;
+  void print(std::ostream &os) const override;
+  StateValue toSMT(State &s) const override;
+  smt::expr getTypeConstraints(const Function &f) const override;
+  std::unique_ptr<Instr> dup(const std::string &suffix) const override;
+};
 
 class Free final : public Instr {
+  Type *type;
```
nunoplopes: why does free() need a type?

ir/memory.cpp
```diff
@@ -329,4 +343,12 @@ Memory Memory::mkIf(const expr &cond, const Memory &then, const Memory &els) {
   return ret;
 }
 
+smt::expr Memory::isBlockAlive(const smt::expr &bid) {
```
nunoplopes: pls don't use the smt:: namespace in the c++ file.

ir/memory.h
```diff
@@ -83,6 +83,8 @@ class Memory {
   unsigned bits_size_t = 64;
 
   smt::expr blocks_val;  // array: (bid, offset) -> StateValue
+  smt::expr blocks_liveness;  // array: bid -> uint(1bit), 1 if alive, 0 if freed
```
nunoplopes: this line is over 80 characters

tools/alive_parser.cpp
```diff
@@ -272,31 +274,55 @@ static Type& get_int_type(unsigned size) {
...
+  if (address_space_number >= pointer_types.size())
+    pointer_types.resize(address_space_number + 1);
+
+  if (!pointer_types[address_space_number])
+    pointer_types[address_space_number] = make_unique<PtrType>(address_space_number);
+
+  return *pointer_types[address_space_number].get();
+}
+
+static Type& parse_type(bool can_fail = true) {
```
nunoplopes: typed pointers are being discontinued from LLVM. I would prefer to keep just `*` as the type (i.e., no `i32*` allowed).
The load/store operations receive the type for the operation instead as 1st argument.

aqjune:
Hi, this week was a bit busy :/ (and upcoming weekdays are Korean holiday :) ) I'll update this PR after then.

aqjune:
- Resolve conflicts
- Use `*` instead of `<ty>*` for pointer types 
- Let `llvm2alive_` deal with `free()` and `malloc()` call. This used isFreeCall() / isMallocLikeFn() from LLVM’s MemoryBuiltin, which required creation of TargetLibraryInfo instance and adding `analysis` to dependency
- Add alive-tv tests for malloc() , free()

ir/instr.h
aqjune: No special reason, just because free in LLVM IR was taking the type of the pointer operand as well (`call void free(i8* ..)`)
I didn't have special preference, and `free %ptr` seems succinct,  so I updated it

ir/memory.cpp
```diff
@@ -218,17 +223,27 @@ pair<expr, vector<expr>> Memory::mkInput(const char *name) {
   return { Pointer(*this, offset, local_bid, bid).release(), { var } };
 }
 
-expr Memory::alloc(const expr &bytes, unsigned align, bool local) {
+expr Memory::alloc(const expr &bytes, unsigned align, bool local, bool heap) {
```
nunoplopes: My idea with the `local` flag was to mark is as stack or heap. Maybe just get rid of `local` altogether?

ir/memory.cpp
```diff
@@ -218,17 +223,27 @@ pair<expr, vector<expr>> Memory::mkInput(const char *name) {
...
   Pointer p(*this, ++last_bid, local);
   state->addPre(p.is_aligned(align));
 
   expr size = bytes.zextOrTrunc(bits_size_t);
   state->addPre(p.block_size() == size);
+
+  state->addPre(blocks_liveness.load(p.get_bid()) == 0);
```
nunoplopes: We should we use the initial `blocks_liveness` for this.  Otherwise code following another branch may conclude that the block was live before.

ir/memory.cpp
```diff
@@ -348,4 +363,12 @@ Memory Memory::mkIf(const expr &cond, const Memory &then, const Memory &els) {
   return ret;
 }
 
+expr Memory::isBlockAlive(const expr &bid) {
```
nunoplopes: what do you think about this and the next function to the pointer class instead?

tools/alive_parser.cpp
```diff
@@ -11,6 +11,7 @@
 #include <memory>
 #include <unordered_map>
 #include <vector>
+#include <optional>
```
nunoplopes: sort this pls

llvm_util/utils.cpp
```diff
@@ -332,6 +336,23 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
     if (!fn) // TODO: support indirect calls
       return error(i);
 
+    if (llvm::isMallocLikeFn(&i, &TLI, false) && fn->getName() == "malloc") {
```
nunoplopes: uhm, why the name check?

aqjune: Because isMallocLikeFn and isFreeCall returns true on variants such as `new` / `delete` operators in C++ as well.
It is conservatively checked because I couldn't find from C++ standard whether `new` should always use the `malloc` call provided.

Or, what do you think about enabling this with a command line flag so people can assume that they are equivalent if needed?

ir/memory.cpp
aqjune: I have no special preference, I'll move them to Pointer.

ir/memory.cpp
aqjune: I think it is still okay because it still satisfies uniqueness of block id. It would depend on how the join operation on memory is defined as well.

ir/memory.cpp
aqjune: Ok. I can rename `Pointer::get_local_bid()` / `Pointer::get_nonlocal_bid()` etc if you want.

aqjune:
- Move isBlockAlive / isAtHeap into Pointer class

ir/memory.cpp
aqjune: Oh, does your comment mean this precondition may not hold with any bid, leading to making Z3 always answer UNSAT?
If my understanding is correct, it would be related with Memory::bits_for_local/nonlocal_bid which is the number of bits that are used to encode bids? If Memory::bits_for_local/nonlocal_bid are large enough, Z3 will be able to assign proper bids so so this precondition can hold.

ir/memory.cpp
nunoplopes: My idea was that local blocks ids would be just for stack, while non-local would be for globals + heap. This is because when we get a pointer as argument it can be either in globals or heap. (it can also be in the stack of some other function, but definitely not on the stack of the current function).
This idea was a cheap way to do ensure that stuff on stack doesn't alias anything else. We don't need any constraint; it's already given by the separated block ids.

Hence I would like to keep the concepts local/non-local as something different from heap/global/stack allocation.

ir/memory.cpp
nunoplopes: Imagine a diamond CFG;
```
      A
    /   \
   B     C
    \   /
      D
```

If an allocation happens in basic block B, the block id gets reserved there and we say that it is now live. But what happens before in block A? And in block C?  And when you merge in D?
You need to state that the block wasn't alive in A. So you can't use `blocks_liveness` of block B. You need to state the property w.r.t. to the original array before execution of A even.

tools/alive_parser.cpp
```diff
@@ -849,17 +873,29 @@ static unique_ptr<Instr> parse_call(string_view name) {
...
+
 static unique_ptr<Instr> parse_copyop(string_view name, token t) {
   tokenizer.unget(t);
   auto &ty = parse_type();
   auto &op = parse_operand(ty);
   return make_unique<UnaryOp>(ty, string(name), op, UnaryOp::Copy);
 }
 
-static unique_ptr<Instr> parse_instr(string_view name) {
+static unique_ptr<Instr> parse_instr(optional<string_view> nameopt) {
```
nunoplopes: I guess here you can just pass an empty string_view and avoid all the changes below.
Also, the code crashes if not given a register when expected. This needs checking.

ir/memory.cpp
aqjune: > You need to state that the block wasn't alive in A.

`blocks_liveness` is preserved before/after execution of instructions that are not malloc/free/alloca, as Memory's `blocks_val` is preserved after execution of non-memory instructions.
If A has no malloc/free/alloca, `blocks_liveness` before A will be equivalent to that of B.
So, this precondition snippet is equivalent to stating that the block wasn't alive before A.

The underlying principle of `blocks_liveness` is that `malloc` is equivalent to a `store`  that writes bit 1 to a hidden memory that records liveness, after checking that originally it was 0.
Similarly, `free` is equivalent to another `store` that writes 0 after checking 1 was written (otherwise UB due to double-free).
So, I believe it is enough to see the memory just before execution of this malloc instruction to avoid having two live blocks with the same id.

tools/alive_parser.cpp
```diff
@@ -829,6 +847,12 @@ static unique_ptr<Instr> parse_freeze(string_view name) {
   return make_unique<Freeze>(ty, string(name), op);
 }
 
+static unique_ptr<Instr> parse_free() {
+  // free %op
+  auto &op = parse_operand(get_pointer_type(0));
```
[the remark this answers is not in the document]
aqjune: https://github.com/AliveToolkit/alive2/pull/93#discussion_r322477704 was about removing ty from free, I thought.

aqjune:
Replied to a few of comments, will return and update code soon

aqjune:
- Update `Memory::mkIf` so it joins `blocks_lifetime` and `blocks_kind`.
- Update `Memory::alloc` by removing `local` parameter
- Add a short comment about local bid and nonlocal bid to memory.h

tools/alive_parser.cpp
aqjune: Ok, makes sense.

tools/alive_parser.cpp
```diff
@@ -601,7 +618,8 @@ static BinOp::Flags parse_binop_flags(token op_token) {
   }
 }
 
-static unique_ptr<Instr> parse_binop(string_view name, token op_token) {
+static unique_ptr<Instr> parse_binop(string_view nameview, token op_token) {
+  string name(nameview);
```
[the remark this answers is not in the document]
aqjune: Done

aqjune:
- Use empty string instead of nullopt to represent an instruction without name
- Don't print "=" if an instruction has no name

tools/alive_parser.cpp
aqjune: Done

ir/memory.cpp
nunoplopes: The issue is when there's nothing written in the liveness record. That means that the block may be alive or not. When you reserve a block id you need to assert that it was dead in the beginning of the function and becomes alive after the malloc. But it couldn't haven't been possibly alive before.
The current code doesn't implement this and is therefore incorrect. It's also creating a bigger expression than needed because `blocks_liveness` at the current BB will already have plenty of stuff in there.

llvm_util/utils.cpp
nunoplopes: I think it's fine. `isMallocLikeFn` should only return true for functions where the first argument is the size and that allocate memory.

ir/instr.cpp
```diff
@@ -145,7 +145,9 @@ void BinOp::print(ostream &os) const {
   case Exact:  flag = " exact "; break;
   }
 
-  os << getName() << " = " << str << flag << print_type(getType())
+  if (!getName().empty())
```
nunoplopes: I don't understand this change. Why would the name be empty? It can't be!

ir/instr.cpp
```diff
@@ -1324,6 +1346,38 @@ unique_ptr<Instr> Alloc::dup(const string &suffix) const {
...
+    os << getName() << " = ";
+  os << "malloc " << *size;
+}
+
+StateValue Malloc::toSMT(State &s) const {
+  auto &[sz, np] = s[*size];
+  s.addUB(np);
+  // TODO: malloc's alignment is implementation defined.
+  return { s.getMemory().alloc(sz, 8, true), true };
```
nunoplopes: 8 is not enough because it needs to return memory that is sufficiently aligned for all types. For a 64-bit machine it's usually 64-bit aligned.  I would just set it to 64 and ignore potential bugs related with over-assumptions regarding alignment (pls leave a comment though).

tools/alive_parser.cpp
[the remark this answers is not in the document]
nunoplopes: no, sorry, that was about the return type, not the input type :)  Sorry for the confusion

aqjune:
Will update this PR this weekend.

aqjune:
Never mind, a bug that I was struggling was magically resolved.
Updates:
- Free checks blocks_liveness of the initial memory. I made State store the initial memory.
- Rollback the empty variable name print thingy
- `free %op` -> `free ty %op`

aqjune:
- Adds negative tests

PR #97, opened 2019-09-16
Parser support for vector types and vector constants

zhengyangl:


tools/alive_parser.cpp
```diff
@@ -309,13 +322,40 @@ static Type& parse_type(bool optional = true) {
...
   }
+}
+
+static Type& parse_vector_type() {
+  tokenizer.ensure(VECTOR_TYPE_PREFIX);
+  unsigned elements = yylval.num;
+
+  auto t = *tokenizer;
+  if (t != IDENTIFIER || yylval.str != "x")
+    error ("Expect x");
```
nunoplopes: extra space before (

tools/alive_parser.cpp
```diff
@@ -425,6 +481,8 @@ static Value& parse_operand(Type &type) {
     return get_fp_constant(yylval.fp_num, type);
   case NUM_STR:
     return get_num_constant(yylval.str, type);
+  case CSLT:
```
nunoplopes: this seems like we are going to have parsing problems with e.g. `<3, 4>`. This won't parse because the code assumes that `<number` is a type. I guess we would need more than 1 token of lookahead; not hard to implement, though.

zhengyangl: How about we assume `<number x` as a type in tokenizer?

nunoplopes: I guess that works as well :)

PR #99, opened 2019-09-20
Binary operators for vectors

zhengyangl:
Hi @nunoplopes , this is a draft implementation of vector binary operator, I would like to take your opinion as early as possible. I'll continue working on this only if you're okay with the approach :)

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -437,12 +437,12 @@ StateValue BinOp::toSMT(State &s) const {
       auto a_i = aty->extract(s[*lhs], i);
       auto b_i = aty->extract(s[*rhs], i);
       auto &cty = aty->getChild(i);
-      vals.emplace_back(scalar_op(move(a_i), move(b_i), cty));
+      vals.emplace_back(scalar_op(a_i, b_i, cty));
     }
     return aty->aggregateVals(vals);
   }
 
-  return scalar_op(move(s[*lhs]), move(s[*rhs]), getType());
+  return scalar_op(s[*lhs], s[*rhs], getType());
 }
 
 expr BinOp::getTypeConstraints(const Function &f) const {
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -170,13 +170,14 @@ static expr fm_poison(const expr &a, const expr &b, const expr &val,
 }
 
 StateValue BinOp::toSMT(State &s) const {
+  bool use_default_poison = true;
   std::function<StateValue(const StateValue&,
-                           const StateValue&, Type&)> scalar_op;
+                           const StateValue&, Type&)> scalar_partial, scalar_op;
 
   switch (op) {
   case Add:
-    scalar_op = [&](auto a, auto b, auto &t) -> StateValue {
-      auto non_poison = a.non_poison && b.non_poison;
+    scalar_partial = [&](auto a, auto b, auto &t) -> StateValue {
+      expr non_poison = true;
       if (flags & NSW)
         non_poison &= a.value.add_no_soverflow(b.value);
       if (flags & NUW)
@@ -186,8 +187,8 @@ StateValue BinOp::toSMT(State &s) const {
     break;
 
   case Sub:
-    scalar_op = [&](auto a, auto b, auto &t) -> StateValue {
-      auto non_poison = a.non_poison && b.non_poison;
+    scalar_partial = [&](auto a, auto b, auto &t) -> StateValue {
+      expr non_poison = true;
       if (flags & NSW)
         non_poison &= a.value.sub_no_soverflow(b.value);
       if (flags & NUW)
@@ -197,8 +198,8 @@ StateValue BinOp::toSMT(State &s) const {
     break;
 
   case Mul:
-    scalar_op = [&](auto a, auto b, auto &t) -> StateValue {
-      auto non_poison = a.non_poison && b.non_poison;
+    scalar_partial = [&](auto a, auto b, auto &t) -> StateValue {
+      expr non_poison = true;
       if (flags & NSW)
         non_poison &= a.value.sub_no_soverflow(b.value);
       if (flags & NUW)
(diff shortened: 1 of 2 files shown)


ir/instr.cpp
```diff
@@ -169,12 +169,44 @@ static expr fm_poison(const expr &a, const expr &b, const expr &val,
...
+                              const AggregateType *aty) {
+  expr res_val;
+  expr res_np;
+  for (unsigned i = 0 ; i < aty->getNumElements(); ++i) {
+    auto a = aty->extract(av, i);
+    auto b = aty->extract(bv, i);
+    expr val;
+    expr non_poison = a.non_poison && b.non_poison;
+
+    switch (op) {
```
zhengyangl: Note there would be a lot of code duplication, between vector and non-vector paths. I will wrap the code with the same shapes with macros.

nunoplopes:
Yeh, this approach is not ideal; it requires a lot of code duplication. There can't be any duplication at all.
The code that computes the value & non-poison expression should be wrapped in a lambda and then applied to each element.
Please let me know if you want to me do this or whether you want to give it a try.

zhengyangl:
Hi @nunoplopes this pull request is ready for review again :)

ir/instr.cpp
```diff
@@ -170,184 +170,279 @@ static expr fm_poison(const expr &a, const expr &b, const expr &val,
...
-  return { move(val), move(not_poison) };
+  if (getType().isVectorType()) {
+    StateValue v;
+    auto aty = getType().getAsAggregateType();
+    for (unsigned i = 0 ; i < aty->numElementsConst(); ++i) {
+      auto a_i = aty->extract(s[*lhs], i);
+      auto b_i = aty->extract(s[*rhs], i);
+      auto &cty = aty->getChild(i);
+      auto vv = cty.toBV(scalar_op(a_i, b_i, cty));
+      v = i == 0 ? move(vv) : v.concat(vv);
```
nunoplopes: please use the aggregateVals() function here instead of toBV/concat by hand (to avoid breaking the abstraction barrier).

ir/instr.cpp
```diff
@@ -170,184 +170,279 @@ static expr fm_poison(const expr &a, const expr &b, const expr &val,
...
 
-  return { move(val), move(not_poison) };
+  if (getType().isVectorType()) {
+    StateValue v;
+    auto aty = getType().getAsAggregateType();
+    for (unsigned i = 0 ; i < aty->numElementsConst(); ++i) {
+      auto a_i = aty->extract(s[*lhs], i);
+      auto b_i = aty->extract(s[*rhs], i);
+      auto &cty = aty->getChild(i);
+      auto vv = cty.toBV(scalar_op(a_i, b_i, cty));
```
nunoplopes: pls do a move(a_i), move(b_i) here.

nunoplopes: actually, I think the signature of the function needs to take `const StateValue &` instead because that's the return value of state[].  Right now it's always calling the copy constructor even for the scalar case.

zhengyangl:
@nunoplopes are you saying hoisting `a.poison && b.poison` out of lambda and moving to the top of the function? This wouldn't work since this binding of `a` of `b` changes for every element when processing a vector.
I cannot come up with any good way that can eliminate the redundant `a.poison && b.poison` (all my solutions are too tricky). Do you have any thoughts?

ir/instr.cpp
```diff
@@ -170,184 +170,279 @@ static expr fm_poison(const expr &a, const expr &b, const expr &val,
...
+    for (unsigned i = 0 ; i < aty->numElementsConst(); ++i) {
+      auto a_i = aty->extract(s[*lhs], i);
+      auto b_i = aty->extract(s[*rhs], i);
+      auto &cty = aty->getChild(i);
+      vals.emplace_back(scalar_op(move(a_i), move(b_i), cty));
+    }
+    return aty->aggregateVals(vals);
+  }
+
+  return scalar_op(move(s[*lhs]), move(s[*rhs]), getType());
```
nunoplopes: no move here, you already changed to const ref

ir/instr.cpp
```diff
@@ -170,184 +170,279 @@ static expr fm_poison(const expr &a, const expr &b, const expr &val,
...
 
-  return { move(val), move(not_poison) };
+  if (getType().isVectorType()) {
+    vector<StateValue> vals;
+    auto aty = getType().getAsAggregateType();
+    for (unsigned i = 0 ; i < aty->numElementsConst(); ++i) {
+      auto a_i = aty->extract(s[*lhs], i);
+      auto b_i = aty->extract(s[*rhs], i);
+      auto &cty = aty->getChild(i);
+      vals.emplace_back(scalar_op(move(a_i), move(b_i), cty));
```
nunoplopes: no move here

nunoplopes:
> @nunoplopes are you saying hoisting a.poison && b.poison out of lambda and moving to the top of the function? This wouldn't work since this binding of a of b changes for every element when processing a vector.
> I cannot come up with any good way that can eliminate the redundant a.poison && b.poison (all my solutions are too tricky). Do you have any thoughts?

My idea was to have a Boolean flag at the beginning like (or whatever name):
```
bool use_default_poison = true;

switch (op) {
case Add:
[] {
non_poison = nsw/nuw constraints;
return { value, non_poison}

case UDiv: ... return { value, true } ;   // just use the default;
}

if (is vector) {
  ...same as scalar, but done for each element ...
}

expr non_poison = a.non_poison && b.non_poison;
auto [v, np] = ..
return { move(v), use_default_poison ? non_poison && np : move(np) } ;
```

ir/instr.cpp
zhengyangl: sure :)

zhengyangl:
@nunoplopes sounds good to me, I will fix as you described soon.

zhengyangl:
@nunoplopes fixed, please review:)

PR #103, opened 2019-09-21
Let lit print expected fail

aqjune:
This is a very small patch that helps lit test print the number of expected fails:
```
$ ninja check
Testing Time: 8.74s
  Expected Passes    : 59
  Expected Failures  : 22
```


nunoplopes:
Finding a bug in a test that has a bug means that Alive passed the test. It found the bug. XFAIL is for tests that are indeed failing and that the developers know about the problem.
Therefore I'm not merging this.

PR #106, opened 2019-09-22
Add null pointer, let free(null) be nop

aqjune:
This PR adds null pointer support and makes free(null) be noop.


ir/value.h
```diff
@@ -74,6 +74,14 @@ class VoidValue final : public Value {
 };
 
 
+class NullPointerValue final : public Value {
```
nunoplopes: Why do we need a NullPointerValue and a NullPointerConst?
Isn't the const sufficient?

ir/memory.cpp
```diff
@@ -254,9 +265,11 @@ expr Memory::alloc(const expr &bytes, unsigned align, bool heap) {
 
 void Memory::free(const expr &ptr) {
   Pointer p(*this, ptr);
-  state->addUB(p.get_offset() == 0);
-  state->addUB(p.is_block_alive());
-  state->addUB(p.is_at_heap());
+  auto isNullPointer = p.get_bid() == 0 && p.get_offset() == 0;
```
nunoplopes: maybe create a method in Pointer, since nullness checks will be frequent I guess.

ir/value.h
aqjune: Yeah, NullPointerValue was enough. I removed NullPointerConst

ir/memory.cpp
aqjune: Done

aqjune:
I applied the comments. :)
Running llvm's unit tests shows +12 failures:
```
Testing Time: 1163.97s -> 1186.59s
Expected Passes    : 4494 -> 4482
Expected Failures  : 25
Unsupported Tests  : 320 
Unexpected Failures: 121 -> 133
```
Here is the list of files:
```
Transforms/SimplifyCFG/PhiEliminate3.ll
Transforms/NewGVN/rle-nonlocal.ll
Transforms/NewGVN/2008-12-14-rle-reanalyze.ll
Transforms/LICM/2007-09-17-PromoteValue.ll
Transforms/JumpThreading/phi-known.ll
Transforms/InstCombine/gep-vector.ll
Transforms/GVN/equality-assume.ll
Transforms/GVN/non-integral-pointers.ll
Transforms/GVN/rle-nonlocal.ll
Transforms/DeadStoreElimination/pr11390.ll
Transforms/CorrelatedValuePropagation/sub.ll
Transforms/CorrelatedValuePropagation/basic.ll
```
I'll investigate why and make a separate PR.

Issue #100, opened 2019-09-20
Function argument and newly allocated block should not alias


aqjune:

aqjune:
This contains two things to consider:
(1) Function arguments that are pointer types should not have block ids that are newly allocated after function began.
```
def void @f(i32* %p) {
%q = call malloc(i32 4) ; p.bid cannot be q.bid
}
```
This can be efficiently encoded by adding preconditions saying that function pointer arguments have upper bound of block id.

(2) The initial memory should not have pointer bytes with block ids newly allocated after beginning.
```
def void @f(i32** %p) {
%q = call malloc(i32 4)
%p2 = load i32*, i32** %p ; p2.bid cannot be q.bid
}
```
One possible solution is to use quantification.
We can say that, for all byte b inside the block that p points to, b's block id less than its upper bound.

aqjune:
Just realized that we don't need upper bound of block id to do (1) and (2), because we already have local_bid and nonlocal_bid.
Making malloc be a local block will resolve (1). I guess unescaped malloc is equivalent to alloca (without behavior w.r.t. free call and possible null pointer return), so I think this is okay.
I'll open a PR that does this.

nunoplopes:
I believe it's all good now. Closing, thanks!

aqjune:
(2) is not resolved, but I'll make a new issue dedicated for that later.

PR #107, opened 2019-09-23
A function argument and new malloc should not alias

aqjune:
This is a small patch that supports no-alias between a function argument and new malloc ( 
#100 , case (1) ).

This is done by making `malloc()` be local, as `alloca` does.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -172,11 +172,11 @@ void Pointer::is_disjoint(const expr &len1, const Pointer &ptr2,
                            len2.zextOrTrunc(m.bits_size_t)));
 }
 
-expr Pointer::is_block_alive() {
+expr Pointer::is_block_alive() const {
   return m.blocks_liveness.load(get_bid());
 }
 
-expr Pointer::is_at_heap() {
+expr Pointer::is_at_heap() const {
   return m.blocks_kind.load(get_bid()) == 1;
 }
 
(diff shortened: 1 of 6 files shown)


tests/alive-tv/freshbid-alloca.src.ll
```diff
@@ -0,0 +1,10 @@
+define i8 @freshbid_alloca(i8* %ptr0) {
+  %ptr = alloca i8
+  store i8 10, i8* %ptr
+  store i8 20, i8* %ptr0
+  %v = load i8, i8* %ptr
+  ret i8 %v
+}
+
+
+; ERROR: Value mismatch
```
nunoplopes: I think I'm misunderstanding something here.
You do a store of 10 and then a load from the same ptr. Isn't `%v` guaranteed to be 10 always? Why does the test fail then?
Same comment for the malloc test

ir/memory.cpp
```diff
@@ -172,17 +172,18 @@ void Pointer::is_disjoint(const expr &len1, const Pointer &ptr2,
...
-expr Pointer::is_at_heap() {
+expr Pointer::is_at_heap() const {
   return m.blocks_kind.load(get_bid()) == 1;
 }
 
 ostream& operator<<(ostream &os, const Pointer &p) {
-  os << "pointer(" << (p.is_local().simplify().isTrue() ? "local" : "non-local")
-     << ", block_id=";
+  os << "pointer(" << (p.is_local().simplify().isTrue() ? "local" : "non-local");
+  os << ", " << (p.is_at_heap().simplify().isTrue() ? "heap" : "non-heap");
```
nunoplopes: uhm, I don't think this will work. When a pointer is constructed for printing, the smt expressions are constant (taken from the model). The memory isn't filled with the model. Anyway, this is a property of the memory block, not of the pointer /per se/.
My suggestion is to revert this change. We can (should) add a printer to the memory later, showing the properties of each block (size, type, etc) and contents.

tests/alive-tv/freshbid-alloca.src.ll
aqjune: I am sorry, this was my mistake. The `; ERROR: Value mismatch` shouldn't have been there.
I removed those.

ir/memory.cpp
aqjune: This worked for input cases in alive-tv test, printing such as
```
Source:                                                                                                      
* %ptr = pointer(local, heap, block_id=256, offset=0)
```
or 
```
Example:                                                                                                     
* %ptr0 = pointer(non-local, non-heap, block_id=0, offset=3458764513820540927) 
```
so I guessed there was magic in simplify() thingy.
But I am not very sure, and also fine with printing them at memory side, so I reverted the change.

aqjune:
Comments applied, llvm unit tests shows no diff at failed tests (121 tests).
Elapsed time slightly shrinked (1163 -> 1107 secs)

PR #108, opened 2019-09-24
Deal with the case when either src or tgt has unreachable

aqjune:
This patch addresses the case when source ends with unreachable but target does (and vice versa):

```
define i32 @f() {
  unreachable
}
=>
define i32 @f() {
  ret i32 0
}
```

The solution this PR suggests is to lower `unreachable` into `assume 0` followed by `ret poison`.

Adding `ret poison` at the last block doesn’t work in the following case:

```
define i32 @f() {
  unreachable
bb: ; unreachable block
  ret i32 0
}
```

llvm unit tests (compared with master + #106 + #107): 1197 sec -> 1193sec , 133 fails -> 109 fails

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -596,13 +596,6 @@ public:
       }
     }
 
-    // infinite void functions may not have a return stmt in LLVM
-    if (!Fn.hasReturn()) {
-      assert(f.getReturnType()->isVoidTy());
-      Fn.getBBs().back()->addInstr(
-        make_unique<Return>(Type::voidTy, Value::voidVal));
-    }
-
     return move(Fn);
   }
 };
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -16,6 +16,8 @@ State::State(const Function &f, bool source)
   predecessor_data[&f.getFirstBB()].try_emplace(nullptr,
                                                 DomainTy(true, set<expr>()),
                                                 *this);
+  return_domain = false;
+  return_val.first = f.getType().getDummyValue(false);
 }
 
 const StateValue& State::exec(const Value &v) {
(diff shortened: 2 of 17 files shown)


llvm_util/llvm2alive.cpp
```diff
@@ -596,11 +614,9 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
       }
     }
 
-    // infinite void functions may not have a return stmt in LLVM
+    // a function with infinite loop may not have a return stmt
     if (!Fn.hasReturn()) {
-      assert(f.getReturnType()->isVoidTy());
-      Fn.getBBs().back()->addInstr(
-        make_unique<Return>(Type::voidTy, Value::voidVal));
```
aqjune: This assertion is removed because non-void function can have an infinite loop as well:

```
; This is valid 
define i32 @f() {
br label %loop
loop:
br label %loop
}
```

aqjune:
Updated the PR by (1) initializing the return_domain and return_val (2) checking return_domain using check_expr

llvm unit tests result: 133 fails -> 110 fails, time: 1197 -> 1076 secs.

aqjune:
- Reduce # of check_expr calls
- Update VoidType::getDummyValue
- Indentation

llvm unit tests: 111 fails; +1 fail (Transforms/SimplifyCFG/sink-common-code.ll), I guess failure is right because it seems it's about sinking function calls.

aqjune:
- No need to call check_expr

llvm unit tests: 107 fails. from the previous result, 4 tests are changed from failure to success:
```
Transforms/InstCombine/assume.ll
Transforms/InstSimplify/assume.ll
Transforms/SimplifyCFG/assume.ll
Transforms/SimplifyCFG/sink-common-code.ll
```
I will investigate why they have changed, and if they have underlying bugs, I'll make a separate PR.

PR #111, opened 2019-09-26
Resolve segmentation fault when aborting 

aqjune:
This is a small patch that resolves segmentation fault that happened due to double-free.

For example, `./alive2/scripts/opt-alive.sh ./llvm/test/Transforms/InstSimplify/gep.ll -instsimplify`
raises assertion failure, but this patch resolves it.

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -129,8 +129,10 @@ struct TVPass : public llvm::FunctionPass {
           !errs.isTimeout() &&
           !errs.isInvalidExpr() &&
           !errs.isOOM() &&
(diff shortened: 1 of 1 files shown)


nunoplopes:
Thanks for reporting the bug. I've fixed it here: https://github.com/AliveToolkit/alive2/commit/7ad22e0f64003cec43bcf6cff2efa45e341d4445
The root cause was that we would double free certain variables when an exception is thrown. The old code wasn't needed since global destructors will take care of destroying those vars. So the easiest fix is just to remove the code. Your PR isn't fully safe in the presence of other exceptions.

Issue #110, opened 2019-09-25
malloc can return null


aqjune:
Currently it is assumed that malloc always returns a valid pointer, but malloc can return null in runtime, meaning that refinement may not hold.

We can make malloc nondeterministically return null, but this will make many practical programs raise UB.

One possible solution is to support a flag such as `malloc_can_return_null` & run TV with the flag set as true and false.

nunoplopes:
Actually LLVM and gcc assume that malloc cannot return null. AFAIR LLVM attaches a nonnull tag to the return value.

aqjune:
LLVM seems to attach `dereferenceable_or_null` to malloc: https://godbolt.org/z/vTpHQ_ .
`new` operator attaches `dereferenceable` (because it raises bad_alloc if there is no storage), and it has nonnull as well.

I think we need a test to see what happens if  malloc is defined as returning null nondeterministically. Having a command line flag that switches the semantics will be helpful for the test, I believe.

aqjune:
Implementing `dereferenceable` and `dereferenceable_null` seems interesting; I'll do that. :) Given
```
%ptr = call dereferenceable(4) i8* @new(i64 4)
```
, we can say that `%ptr` is dereferenceable (with 4 bytes) at this program point.
I have #112 ongoing, and #106 and #108 to investigate unit test changes; I can do this dereferenceable thingy after these.

aqjune:
As this is addressed by #114, I'll open a separate issue for dereferenceable thingy.

PR #112, opened 2019-09-26
Support global variables, alloca/glbvar's address should be non-zero

aqjune:
This patch adds support for global variables, and makes alloca/glbvar's address never be zero.
malloc's address is not addressed here, as malloc() can return null in runtime. (#110 )

Global variable support is done as follows:
- A global variable in src and tgt are given the same block id , and this information is passed as
```
    sym_exec(src_state);
    tgt_state.setGlobalVarBidsFrom(src_state);
    sym_exec(tgt_state);
```
at TransformVerify::verify().
- To give the block id precisely, `Memory::alloc` now can take an optional argument that specifies which bid to use. It is caller's responsibility to give a unique bid.

LLVM unit tests result:
- Failures: 133 -> 173 (+40), around +30 tests are global (test/Transforms/SafeStack , test/Transforms/GlobalOpt). For other tests, investigation is needed.
- Running time: 1197 -> 296 secs; speedup exists, which is a bit weird :/

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -248,16 +248,38 @@ pair<expr, vector<expr>> Memory::mkInput(const char *name) {
   return { Pointer(*this, offset, local_bid, bid).release(), { var } };
 }
 
-expr Memory::alloc(const expr &size, unsigned align, bool heap) {
-  Pointer p(*this, ++last_bid, true);
+expr Memory::alloc(const expr &size, unsigned align, BlockKind blockKind,
+                   optional<unsigned> bidopt, unsigned *bid_out) {
+  unsigned bid = bidopt ? *bidopt : last_bid;
+  if (!bidopt)
+    last_bid++;
+  assert(bid < last_bid);
+
+  if (bid_out)
+    *bid_out = bid;
+
+  // It is allocated as a local block if blockKind is heap or stack.
+  bool is_local = blockKind != GLOBAL;
+
+  Pointer p(*this, bid, is_local);
   state->addPre(p.is_aligned(align));
 
   expr size_zextOrTrunced = size.zextOrTrunc(bits_size_t);
   state->addPre(p.block_size() == size_zextOrTrunced);
+  // TODO: If its address space is not 0, its address can be 0.
+  state->addPre(p.get_address() != 0);
 
-  state->addPre(!mk_liveness_uf().load(p.get_bid()));
-  blocks_liveness = blocks_liveness.store(p.get_bid(), true);
-  blocks_kind = blocks_kind.store(p.get_bid(), expr::mkUInt(heap, 1));
+  if (is_local) {
+    // Initially there was no such block, now it is allocated.
+    state->addPre(!mk_liveness_uf().load(p.get_bid()));
+    blocks_liveness = blocks_liveness.store(p.get_bid(), true);
+  } else {
+    // The memory block was initially alive.
+    // TODO: const global variables should be read-only
+    state->addPre(mk_liveness_uf().load(p.get_bid()));
+  }
+  blocks_kind = blocks_kind.store(p.get_bid(),
+      expr::mkUInt(blockKind == HEAP, 1));
 
   return p();
 }
@@ -384,6 +406,11 @@ Memory Memory::mkIf(const expr &cond, const Memory &then, const Memory &els) {
   ret.last_bid     = max(then.last_bid, els.last_bid);
   ret.last_idx_ptr = max(then.last_idx_ptr, els.last_idx_ptr);
   return ret;
+}
+
+void Memory::bumpLastBid(unsigned bid) {
+  assert(last_bid <= bid + 1);
+  last_bid = bid + 1;
 }
 
 }
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -13,9 +13,6 @@ namespace IR {
 
 State::State(const Function &f, bool source)
   : f(f), source(source), memory(*this) {
-  predecessor_data[&f.getFirstBB()].try_emplace(nullptr,
-                                                DomainTy(true, set<expr>()),
-                                                *this);
   return_domain = false;
   return_val.first = f.getType().getDummyValue(false);
 }
@@ -81,6 +78,11 @@ bool State::startBB(const BasicBlock &bb) {
   assert(undef_vars.empty());
   ENSURE(seen_bbs.emplace(&bb).second);
   current_bb = &bb;
+
+  if (f.getFirstBB().getName() == bb.getName()) {
+    domain.first = true;
+    return true;
+  }
 
   auto I = predecessor_data.find(&bb);
   if (I == predecessor_data.end())
@@ -175,6 +177,35 @@ void State::addUndefVar(const expr &var) {
 void State::resetUndefVars() {
   quantified_vars.insert(undef_vars.begin(), undef_vars.end());
   undef_vars.clear();
+}
+
+void State::addGlobalVarBid(const string &glbvar, unsigned bid) {
+  assert(glbvar_bids.count(glbvar) == 0);
+  glbvar_bids.emplace(glbvar, bid);
+}
+
+bool State::hasGlobalVarBid(const string &glbvar,
+    unsigned &bid) const {
+  auto itr = glbvar_bids.find(glbvar);
+  bool found = itr != glbvar_bids.end();
+  if (found) {
+    bid = itr->second;
+  }
+  return found;
+}
+
+void State::copyGlobalVarBidsFromSrc(const State &src) {
+  assert(glbvar_bids.empty());
+  assert(src.isSource());
+  // Copy glbvar_bids.
+  glbvar_bids = src.glbvar_bids;
+  // Bump memory's last_bid so Memory::alloc()'s newly assigned bids never
+  // collide with this
+  unsigned maxid = 0;
+  for (const auto &itm: glbvar_bids) {
+    maxid = max(maxid, itm.second);
+  }
+  getMemory().bumpLastBid(maxid);
 }
 
 }
--- a/ir/value.cpp
+++ b/ir/value.cpp
@@ -88,7 +88,31 @@ StateValue NullPointerValue::toSMT(State &s) const {
 }
 
 
-void Input::print(std::ostream &os) const {
+void GlobalVariable::print(ostream &os) const {
+  UNREACHABLE();
+}
+
+StateValue GlobalVariable::toSMT(State &s) const {
+  auto sizeexpr = expr::mkInt(allocsize, 64);
+  expr ptrval;
+  unsigned glbvar_bid;
+  if (s.hasGlobalVarBid(getName(), glbvar_bid)) {
+    // Use the same block id that is used by src
+    assert(!s.isSource());
+    ptrval = s.getMemory().alloc(sizeexpr, align, Memory::GLOBAL, glbvar_bid);
+  } else {
+    ptrval = s.getMemory().alloc(sizeexpr, align, Memory::GLOBAL, nullopt,
+                                 &glbvar_bid);
+    s.addGlobalVarBid(getName(), glbvar_bid);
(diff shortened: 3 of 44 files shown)


aqjune:
Writing to const global variable should raise UB, but it is not addressed in this PR.

llvm_util/utils.h
```diff
@@ -5,6 +5,7 @@
 
 #include <ostream>
 #include <string>
+#include "llvm/IR/DataLayout.h"
```
nunoplopes: don't include this header here; add a forward decl below

llvm_util/utils.h
```diff
@@ -29,7 +30,7 @@ IR::Type& get_int_type(unsigned bits);
 IR::Type* llvm_type2alive(const llvm::Type *ty);
 
 IR::Value* make_intconst(uint64_t val, int bits);
-IR::Value* get_operand(llvm::Value *v);
+IR::Value* get_operand(llvm::Value *v, const llvm::DataLayout &DL);
```
nunoplopes: this makes all calls more verbose. it's better to pass this to init_llvm_utils() below.

llvm_util/utils.cpp
```diff
@@ -51,7 +53,7 @@ string value_name(const llvm::Value &v) {
   if (!v.getName().empty())
     return name = '%' + v.getName().str();
   return name = v.getType()->isVoidTy() ? "<void>"
-                                        : '%' + to_string(value_id_counter++);
+                                        : "%" + to_string(value_id_counter++);
```
nunoplopes: why this change?

ir/value.cpp
```diff
@@ -5,7 +5,6 @@
 #include "smt/expr.h"
 #include "util/compiler.h"
 #include "util/config.h"
-
```
nunoplopes: ...

ir/value.cpp
```diff
@@ -88,7 +87,32 @@ StateValue NullPointerValue::toSMT(State &s) const {
...
+  unsigned glbvar_bid;
+  if (s.hasGlobalVarBid(getName(), glbvar_bid))
+    // Use the same block id that is used by src
+    ptrval = s.getMemory().alloc(sizeexpr, align, Memory::GLOBAL, glbvar_bid);
+  else {
+    ptrval = s.getMemory().alloc(sizeexpr, align, Memory::GLOBAL, nullopt,
+                                 &glbvar_bid);
+    s.addGlobalVarBid(getName(), glbvar_bid);
+  }
+  if (isconst && initval) {
```
nunoplopes: is there any reason to store the init val if !isconst?  if not, we shouldn't store it at all.

ir/value.cpp
```diff
@@ -88,7 +87,32 @@ StateValue NullPointerValue::toSMT(State &s) const {
...
+void GlobalVariable::print(ostream &os) const {
+  UNREACHABLE();
+}
+
+StateValue GlobalVariable::toSMT(State &s) const {
+  auto sizeexpr = expr::mkInt(allocsize, 64);
+  expr ptrval;
+  unsigned glbvar_bid;
+  if (s.hasGlobalVarBid(getName(), glbvar_bid))
+    // Use the same block id that is used by src
```
nunoplopes: this should assert(!is_src)

tools/transform.cpp
```diff
@@ -310,6 +310,7 @@ Errors TransformVerify::verify() const {
 
   try {
     sym_exec(src_state);
+    tgt_state.setGlobalVarBidsFrom(src_state);
```
nunoplopes: maybe call it copyGlobalVarsFromSrc() to make it explicit

tests/alive-tv/glb-reorder.src.ll
```diff
@@ -0,0 +1,9 @@
+@A = global i32 0
+@B = global i32 0
+
+define i32 @foo() {
+  %l = load i32, i32* @A
+  store i32 0, i32* @B
```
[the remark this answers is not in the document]
aqjune: Updated. :)

llvm_util/utils.h
aqjune: Yep, updated.

llvm_util/utils.cpp
aqjune: It was because visual studio code couldn't syntax highlight the following code if it was '%', just a silly reason :/ Undoed the change.

ir/value.cpp
aqjune: Sorry :P updated

ir/value.cpp
aqjune: Nope, if a global variable is not constant, it is not necessary to store its value because a caller may change that. This if block states that it is already ignoring the initial value.

ir/value.cpp
aqjune: Added

tools/transform.cpp
aqjune: Updated

ir/state.cpp
```diff
@@ -13,9 +13,6 @@ namespace IR {
 
 State::State(const Function &f, bool source)
   : f(f), source(source), memory(*this) {
-  predecessor_data[&f.getFirstBB()].try_emplace(nullptr,
-                                                DomainTy(true, set<expr>()),
-                                                *this);
```
aqjune: The reason why I removed the predecesor_data for FirstBB was that it was storing an empty memory which made State::startBB(FirstBB) ovewrite the initial memory which contained initial values of const global vars. I made predecessor_data[FirstBB] store nothing and updated State::startBB.

aqjune:
Updated PR.
llvm unit tests: 108 fails (master) -> 133 fails. It seems all fails are related with global optimizations / X86-specific optimizations (safestack thingy) / atomic operations.

ir/memory.cpp
```diff
@@ -248,16 +248,37 @@ pair<expr, vector<expr>> Memory::mkInput(const char *name) {
...
+  // It is allocated as a local block if blockKind is heap or stack.
+  bool is_local = blockKind != GLOBAL;
+
+  Pointer p(*this, bid, is_local);
   state->addPre(p.is_aligned(align));
 
   expr size_zextOrTrunced = size.zextOrTrunc(bits_size_t);
   state->addPre(p.block_size() == size_zextOrTrunced);
+  // TODO: If its address space is not 0, its address can be 0.
+  state->addPre(p.get_address() != 0);
```
nunoplopes: wait, I thought the conclusion is that malloc can return null?

ir/value.cpp
nunoplopes: Then let's remove the `&& initval` bit.

llvm_util/utils.cpp
```diff
@@ -165,6 +170,34 @@ Value* get_operand(llvm::Value *v) {
...
+      // TODO: Array type of a global variable is not supported.
+      return nullptr;
+    } else if (auto st = dyn_cast<llvm::StructType>(gv->getValueType())) {
+      if (st->isOpaque())
+        // TODO: Global variable of opaque type is not supported.
+        return nullptr;
+    }
+    Value *initval = nullptr;
+    if (gv->hasInitializer()) {
+      initval = get_operand(gv->getInitializer());
```
nunoplopes: pls don't the initializer if it isn't const

ir/memory.cpp
```diff
@@ -248,16 +248,37 @@ pair<expr, vector<expr>> Memory::mkInput(const char *name) {
   return { Pointer(*this, offset, local_bid, bid).release(), { var } };
 }
 
-expr Memory::alloc(const expr &size, unsigned align, bool heap) {
-  Pointer p(*this, ++last_bid, true);
+expr Memory::alloc(const expr &size, unsigned align, BlockKind blockKind,
+                   optional<unsigned> bidopt, unsigned *bid_out) {
+  unsigned bid = bidopt ? *bidopt : last_bid;
+  if (last_bid <= bid) {
```
nunoplopes: you do a bump in tgt, and in src bidopt should be null. I would ensure this invariant with:
```
if (!bidopt)
  ++last_bid;
assert(bid < last_bid);
```

ir/state.cpp
```diff
@@ -177,4 +178,33 @@ void State::resetUndefVars() {
   undef_vars.clear();
 }
 
+void State::addGlobalVarBid(const std::string &glbvar, unsigned bid) {
+  assert(glbvar_bids.find(glbvar) == glbvar_bids.end());
+  glbvar_bids[glbvar] = bid;
+}
+
+bool State::hasGlobalVarBid(const std::string &glbvar,
```
nunoplopes: no std::

ir/state.cpp
```diff
@@ -177,4 +178,33 @@ void State::resetUndefVars() {
   undef_vars.clear();
 }
 
+void State::addGlobalVarBid(const std::string &glbvar, unsigned bid) {
```
nunoplopes: no std::

ir/state.cpp
```diff
@@ -177,4 +178,33 @@ void State::resetUndefVars() {
...
+  assert(glbvar_bids.find(glbvar) == glbvar_bids.end());
+  glbvar_bids[glbvar] = bid;
```
nunoplopes: .emplace() is more efficient

ir/state.cpp
```diff
@@ -177,4 +178,33 @@ void State::resetUndefVars() {
   undef_vars.clear();
 }
 
+void State::addGlobalVarBid(const std::string &glbvar, unsigned bid) {
+  assert(glbvar_bids.find(glbvar) == glbvar_bids.end());
```
nunoplopes: .count(glbvar) == 0

ir/state.cpp
```diff
@@ -82,26 +79,30 @@ bool State::startBB(const BasicBlock &bb) {
   ENSURE(seen_bbs.emplace(&bb).second);
   current_bb = &bb;
 
+  bool isFirstBB = f.getFirstBB().getName() == bb.getName();
```
nunoplopes: instead of this huge change, can we do instead (since none of the code later is relevant):
```
if (f.getFirstBB().getName() == bb.getName())
  return true;
```

llvm_util/utils.cpp
```diff
@@ -165,6 +170,34 @@ Value* get_operand(llvm::Value *v) {
     return ret;
   }
 
+  if (auto gv = dyn_cast<llvm::GlobalVariable>(v)) {
+    auto itr = globalvars.find(gv);
+    if (itr != globalvars.end())
+      return itr->second;
+
+    if (gv->getValueType()->isArrayTy()) {
+      // TODO: Array type of a global variable is not supported.
```
nunoplopes: I don't see any reason for limiting the type. In the worst case, get_operand returns a null expr.

ir/memory.cpp
aqjune: I believe this can be addressed by making Malloc::toSMT return `mkIf(random_bool, Memory::alloc(), nullptr)`. I will make a separate PR for this so discussion for its implementation can be done there.

ir/state.cpp
aqjune: Updated

llvm_util/utils.cpp
aqjune: A global array's type is a pointer to array, so llvm_type2alive(v->getType()) at the beginning of get_operand successfully returns IR type, that's why isArrayTy() here is needed.

ir/memory.cpp
aqjune: I see, updated.

ir/value.cpp
aqjune: Const global may not have its initial value if it is extern variable, so initial is still needed. Instead, I'll leave initval only.

llvm_util/utils.cpp
aqjune: updated

aqjune:
Addressed comments.
Unit tests: 116 fails -> 141 fails (+25, as described in the previous iteration).

PR #114, opened 2019-09-30
Let malloc nondeterministically return null, let alive parser support ptrtoint

aqjune:
This PR
- Updates malloc's semantics so it nondeterministically returns nullptr (#110)
- Adds `--malloc-nonnull` flag which makes malloc always return valid pointer
- Adds parser support for ptrtoint 
- Has some random changes that removes std:: uses at .cpp 

There was no diff at llvm unit test results, as expected.
I guess this PR won't conflict with #99 because this patch is not touching binary operators.


ir/memory.cpp
```diff
@@ -248,8 +248,12 @@ pair<expr, vector<expr>> Memory::mkInput(const char *name) {
   return { Pointer(*this, offset, local_bid, bid).release(), { var } };
 }
 
-expr Memory::alloc(const expr &size, unsigned align, bool heap) {
+expr Memory::alloc(const expr &size, unsigned align, bool heap,
+    unsigned *bid_out) {
```
aqjune: This change is included at #112 , will be removed after the PR is merged.

ir/instr.cpp
```diff
@@ -1379,7 +1380,18 @@ StateValue Malloc::toSMT(State &s) const {
...
+  unsigned new_bid;
+  auto p = s.getMemory().alloc(sz, 8, true, &new_bid);
+
+  if (isNonNull)
+    return { move(p), true };
+
+  auto nullp = Pointer::mkNullPointer(s.getMemory());
+  stringstream ss;
+  ss << "malloc_" << new_bid << "_isnull_" << (s.isSource() ? "src" : "tgt");
+  auto flag = expr::mkBoolVar(ss.str().c_str());
```
nunoplopes: Please add a API for this: `smt::expr::mkFreshVar(const char *prefix, const expr &type)`.
This should call Z3_mk_fresh_var().

nunoplopes: Probably there are other places where we should have used this already

aqjune: Updated, let freeze use it too

aqjune:
> looks good. My only concern is whether it's worth it to be able to disable this behavior. Do you have run-time numbers?

If there are multiple mallocs, it seems to have performance difference. I added mallocs-null.src/tgt.ll that are used to test it: with the default nondeterministic semantics, its user time is 0.028 secs in average, whereas with `--malloc-nonnull` it takes about 0.020 secs in average.
So, it will be useful to turn on `--malloc-nonnull` when possible.
One possible case that I can come up with is when all mallocs `p` have trivial guards `if (p == null) { return; }`, `--malloc-nonnull` can be turned-on, because uses are already checked.

aqjune:
I belive this flag is also useful to those who want to run their own test cases with/without the malloc-returning-null semantics. There are people who may have heard about the malloc-null issue (as googling "compiler optimization malloc null" tells), and this flag helps them play with it.

aqjune:
- Remove --malloc-nonnull flag

PR #115, opened 2019-10-01
fix alignment check when bytes is 0

Dongjoo-Kim:
I did two minor fix.

1) Pointer must be aligned when bytes in memset (memcpy) is 0.

2) When alignments in memset (memcpy) is zero, make them 1.

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -200,11 +200,8 @@ public:
     if (!ptr || !val || !bytes)
       return error(i);
 
-    auto destAlign = i.getDestAlignment();
-    if(destAlign == 0)
-      destAlign = 1;
-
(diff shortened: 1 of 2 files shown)


llvm_util/llvm2alive.cpp
```diff
@@ -200,8 +200,11 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
     if (!ptr || !val || !bytes)
       return error(i);
 
-    RETURN_IDENTIFIER(make_unique<Memset>(*ptr, *val, *bytes,
-                                          i.getDestAlignment()));
+    auto destAlign = i.getDestAlignment();
```
nunoplopes: you can use a simpler `min(1, i.getDestAlignment()`

PR #116, opened 2019-10-01
fix: clean duplicated default non_poison values in FP bin-ops

zhengyangl:
This is my stupid, I'll be careful next time :)


nunoplopes:
ah, I noticed this as well, and fixed it already :)

PR #118, opened 2019-10-04
Support reading/writing pointer values from memory 

aqjune:
This PR implements reading/writing pointer values from memory.
To do this, a new `Byte` class is implemented, and conversion from `vector<Byte>` to `StateValue` and vice versa (`valueToBytes`, `bytesToValue`) are added.
The definition of Byte follows that of Twin Memory Model, with granularity of _bytes_, not bits, for performance. Also there is no instruction in LLVM that supports manipulation of a pointer bit.
A byte is either a pointer byte, which is a pair (ptr, i) meaning the i'th byte of ptr (0 <= i < 8),
or a non-pointer byte, which has 8-bit numeric digit along with 8-bit poison info. 
```
// Pointer byte's representation:
//   +-+------------------+-------------------+----------------+-------------+
//   |1| non-poison byte? | ptr offset        | block id       | byte offset |
//   | | (1 bit)          | (bits_for_offset) | (bits_for_bid) | (3 bits)    |
//   +-+------------------+-------------------+----------------+-------------+
// Non-pointer byte's representation:
//   +-+--------------------------+--------------------+---------------------+
//   |0| padding(000000000)       | non-poison bits?   | data                |
//   | |                          | (8 bits)           | (8 bits)            |
//   +-+--------------------------+--------------------+---------------------+
```
The semantics of punning between pointer and non-pointer byte also follows the Twin Memory Model. Loading a pointer value as non-pointer yields poison. and loading non-pointer values (integer, float, etc) as pointer yields poison as well.
LLVM unit test failures: 157 -> 173, running time: 278 -> 280 secs.
Besides from Byte implementation, this patch also
- Makes inttoptr raise `Unsupported instruction`, because the pointer returned by `inttoptr` was illformed, causing crash while running unit tests
- Fixes a bug about type of a global variable.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -11,10 +11,60 @@ using namespace util;
 
 namespace IR {
 
-Pointer::Pointer(Memory &m, const char *var_name)
+// A data structure that represents a byte.
+// A byte is either a pointer byte or a non-pointer byte.
+// Pointer byte's representation:
+//   +-+------------------+-------------------+----------------+-------------+
+//   |1| non-poison byte? | ptr offset        | block id       | byte offset |
+//   | | (1 bit)          | (bits_for_offset) | (bits_for_bid) | (3 bits)    |
+//   +-+------------------+-------------------+----------------+-------------+
+// Non-pointer byte's representation:
+//   +-+--------------------------+--------------------+---------------------+
+//   |0| padding(000000000)       | non-poison bits?   | data                |
+//   | |                          | (8 bits)           | (8 bits)            |
+//   +-+--------------------------+--------------------+---------------------+
+class Byte {
+  // Memory, which is needed to get the size of offset and block id.
+  const Memory &m;
+  // The underlying representation.
+  smt::expr p;
+
+public:
+  Byte(Byte &&b): m(b.m), p(std::move(b.p)) {}
+  // Creates a byte with its raw representation.
+  Byte(const Memory &m, smt::expr &&byterepr);
+  // Creates a pointer byte that represents i'th byte of p.
+  // non_poison should be an one-bit vector or boolean.
+  Byte(const Pointer &p, unsigned i, const smt::expr &non_poison);
+  // Creates a non-pointer byte that has data and non_poison.
+  // data and non_poison should have 8 bits.
+  Byte(const Memory &m, const smt::expr &data, const smt::expr &non_poison);
+
+  // Returns a bit that represents whether this byte is a pointer byte.
+  smt::expr is_ptrbyte() const { return p.extract(p.bits() - 1, p.bits() - 1); }
+  // Assuming that this is a pointer byte, returns a single bit that represents
+  // whether the byte is poison.
+  smt::expr ptr_nonpoison() const {
+    return p.extract(p.bits() - 2, p.bits() - 2);
+  }
+  // Returns its value assuming that this is a pointer byte.
+  smt::expr ptr_value() const { return p.extract(p.bits() - 3, 3); }
+  // Assuming that this is a pointer byte, returns its byte offset.
+  smt::expr ptr_byteoffset() const { return p.extract(2, 0); }
+  // Assuming that this is a non-pointer byte, returns 8-bit bitvector.
+  smt::expr nonptr_nonpoison() const { return p.extract(15, 8); };
+  // Assuming that this is a non-pointer byte, returns its value.
+  smt::expr nonptr_value() const { return p.extract(7, 0); }
+
+  const smt::expr& operator()() const { return p; }
+  smt::expr release() const { return std::move(p); }
+};
+
+
+Pointer::Pointer(const Memory &m, const char *var_name)
   : m(m), p(expr::mkVar(var_name, total_bits())) {}
 
-Pointer::Pointer(Memory &m, unsigned bid, bool local) : m(m) {
+Pointer::Pointer(const Memory &m, unsigned bid, bool local) : m(m) {
   expr bid_expr;
   if (local)
     bid_expr = expr::mkUInt((uint64_t)bid << m.bits_for_nonlocal_bid,
@@ -24,7 +74,7 @@ Pointer::Pointer(Memory &m, unsigned bid, bool local) : m(m) {
   p = expr::mkUInt(0, m.bits_for_offset).concat(bid_expr);
 }
 
-Pointer::Pointer(Memory &m, const expr &offset, const expr &local_bid,
+Pointer::Pointer(const Memory &m, const expr &offset, const expr &local_bid,
                  const expr &nonlocal_bid)
   : m(m), p(offset.concat(local_bid).concat(nonlocal_bid)) {}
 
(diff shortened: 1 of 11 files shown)


tests/alive-tv/memory/punning-ptrint.tgt.ll
```diff
@@ -0,0 +1,8 @@
+target datalayout = "e-m:o-i64:64-f80:128-n8:16:32:64-S128"
+
+define i8 @ptr_int_punning(i8** %pptr, i64 %n) {
+  %ptr = call i8* @malloc(i64 1)
+  store i8* %ptr, i8** %pptr
+  ret i8 0
```
nunoplopes: I think ret undef would be better here (since malloc can return null, 0 isn't a particularly good value).

tests/alive-tv/memory/punning-intptr.tgt.ll
```diff
@@ -0,0 +1,11 @@
+target datalayout = "e-m:o-i64:64-f80:128-n8:16:32:64-S128"
+
+define i8* @int_ptr_punning(i8* %ptr, i64 %n) {
+  store i8 1, i8* %ptr
+  %n2 = and i64 %n, 7
+  %n3 = sub i64 0, %n2
+  %ptr_2 = getelementptr i8, i8* %ptr, i64 %n3
+  %pptr = bitcast i8* %ptr_2 to i8**
+  %v = load i8*, i8** %pptr
+  ret i8* null
```
nunoplopes: ret undef (same comment as below)

llvm_util/llvm2alive.cpp
```diff
@@ -160,7 +160,7 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
     case llvm::Instruction::FPToSI:   op = ConversionOp::FPToSInt; break;
     case llvm::Instruction::FPToUI:   op = ConversionOp::FPToUInt; break;
     case llvm::Instruction::PtrToInt: op = ConversionOp::Ptr2Int; break;
-    case llvm::Instruction::IntToPtr: op = ConversionOp::Int2Ptr; break;
+    case llvm::Instruction::IntToPtr:
```
nunoplopes: please remove this diff. It's better to have int2ptr return an invalid expression that skipping a whole function.
Also, this is unrelated with the rest of the PR.

ir/memory.h
```diff
@@ -157,7 +217,15 @@ class Memory {
...
 
+  // Returns the number of bits needed to represent a pointer offset.
   unsigned bitsOffset() const { return bits_for_offset; }
+  // Returns the number of bits needed to represent a block id.
+  unsigned bitsBid() const { return bits_for_local_bid +
+                                    bits_for_nonlocal_bid; }
+  // Returns the number of bits needed to represent a byte.
+  unsigned bitsByte() const { return 1 + 1 + bitsBid() + bitsOffset() + 3; }
+  // Returns the number of bits needed to represent the size of memory.
+  unsigned bitsMemorySize() const { return bits_size_t; }
```
nunoplopes: Memory size is a strange name. Either use ptrSize() or something.
Number of bits of a pointer is not the same as the log(memory-size), since not all archs are byte addressable.

ir/memory.h
```diff
@@ -91,16 +93,74 @@ class Pointer {
...
   // FIXME: these should be tuned per function
   unsigned bits_for_offset = 64;
   unsigned bits_for_local_bid = 8;
   unsigned bits_for_nonlocal_bid = 8;
+  // bits_size_t is equivalent to the size of a pointer.
   unsigned bits_size_t = 64;
 
-  smt::expr blocks_val; // array: (bid, offset) -> StateValue
+  // blocks_val is an array from (bid, offset) to Byte.
+  smt::expr blocks_val;
```
nunoplopes: pls don't change the format

ir/memory.h
```diff
@@ -91,16 +93,74 @@ class Pointer {
 };
 
 
+// A data structure that represents a byte.
```
nunoplopes: This whole Byte class is an implementation detail of the memory. I don't see a need to expose it to users.
I suggest moving it completely to memory.cpp instead.

ir/memory.cpp
```diff
@@ -195,6 +195,125 @@ ostream& operator<<(ostream &os, const Pointer &p) {
 }
 
 
+Byte::Byte(const Memory &m, expr &&byterepr): m(m), p(move(byterepr)) {
+  assert(p.bits() == m.bitsByte());
+}
+
+Byte::Byte(const Pointer &ptr, unsigned i, expr non_poison)
```
nunoplopes: I would pass non_poison as const expr & to avoid copies

ir/memory.cpp
```diff
@@ -195,6 +195,125 @@ ostream& operator<<(ostream &os, const Pointer &p) {
...
+    : m(ptr.getMemory()) {
+  // TODO: support pointers larger than 64 bits.
+  assert(m.bitsMemorySize() <= 64 && m.bitsMemorySize() % 8 == 0);
+  // non_poison should be either bool or 1 bit.
+  if (non_poison.isBool())
+    non_poison = non_poison.toBVBool();
+  assert(non_poison.bits() == 1);
```
nunoplopes: this assert is redundant

ir/memory.cpp
```diff
@@ -195,6 +195,125 @@ ostream& operator<<(ostream &os, const Pointer &p) {
...
+
+  expr byteofs = expr::mkUInt(i, 3);
+  p = expr::mkUInt(1, 1).concat(non_poison).concat(ptr()).concat(byteofs);
+  assert(p.bits() == m.bitsByte());
+}
+
+Byte::Byte(const Memory &m, expr data, expr non_poison): m(m) {
```
nunoplopes: I would turn both expr to const expr &

ir/memory.cpp
```diff
@@ -195,6 +195,125 @@ ostream& operator<<(ostream &os, const Pointer &p) {
...
+  assert(non_poison.bits() == 8);
+
+  unsigned padding = m.bitsByte() - 1 - 8 - 8;
+  p = expr::mkUInt(0, 1).concat(expr::mkUInt(0, padding)).concat(non_poison)
+                        .concat(data);
+}
+
+
+void valueToBytes(const StateValue &val, const Type &fromType,
+                  const Memory &mem, vector<Byte> &bytes) {
```
nunoplopes: you can return the vector instead of passing it as an argument (C++11 way of doing things :) )

ir/memory.cpp
```diff
@@ -195,6 +195,125 @@ ostream& operator<<(ostream &os, const Pointer &p) {
...
+  } else {
+    StateValue bvval = fromType.toBV(val);
+    unsigned bitsize = bvval.bits();
+    unsigned bytesize = divide_up(bitsize, 8);
+
+    bvval = bvval.zext(bytesize * 8 - bitsize);
+
+    for (unsigned i = 0; i < bytesize; ++i) {
+      expr data  = bvval.value.extract((i + 1) * 8 - 1, i * 8);
+      expr np    = bvval.non_poison.extract((i + 1) * 8 - 1, i * 8);
```
nunoplopes: StateValue has extract() now. You can do these 2 lines in one go.

ir/memory.cpp
```diff
@@ -195,6 +195,125 @@ ostream& operator<<(ostream &os, const Pointer &p) {
...
+      auto &b = bytes[i];
+      // Is ptr byte?
+      expr is_ptrbyte = b.is_ptrbyte();
+      // Pointer bytes' offset
+      expr byte_ofs = b.ptr_byteoffset();
+      // Pointer's offset
+      expr ptr_value = b.ptr_value();
+      // Is non-poison byte?
+      expr is_nonpoison = b.ptr_nonpoison();
+
```
nunoplopes: too many comments above; the code is fairly self-explanatory.

ir/memory.cpp
```diff
@@ -195,6 +195,125 @@ ostream& operator<<(ostream &os, const Pointer &p) {
...
+      if (first) {
+        loaded_ptr = move(ptr_value);
+        all_ptrbytes = is_nonpoison & is_ptrbyte;
+        ptr_allequiv = true;
```
nunoplopes: I think you can use a single non_poison var and keep doing &= on that. Also, if you initialize that, most of this moster case split goes away (there's no need to do SMT simplification here).

ir/memory.cpp
```diff
@@ -195,6 +195,125 @@ ostream& operator<<(ostream &os, const Pointer &p) {
...
+      StateValue v(b.nonptr_value(), b.nonptr_nonpoison());
+
+      val = first ? move(v) : v.concat(val);
+      all_nonptr_bytes = first ? ~b.is_ptrbyte() :
+                         (all_nonptr_bytes & ~b.is_ptrbyte());
+      first = false;
+    }
+
+    val = toType.fromBV(val.trunc(bitsize));
+    val.non_poison = val.non_poison && (all_nonptr_bytes == 1);
```
nunoplopes: use &=

ir/memory.cpp
```diff
@@ -195,6 +195,125 @@ ostream& operator<<(ostream &os, const Pointer &p) {
...
+    bool first = true;
+    // All bytes should be non-pointer, because we're loading it as a
+    // non-pointer type.
+    expr all_nonptr_bytes;
+
+    for (auto &b: bytes) {
+      StateValue v(b.nonptr_value(), b.nonptr_nonpoison());
+
+      val = first ? move(v) : v.concat(val);
+      all_nonptr_bytes = first ? ~b.is_ptrbyte() :
```
nunoplopes: please use &= and get rid of the special case.

ir/memory.cpp
```diff
@@ -296,79 +416,70 @@ void Memory::free(const expr &ptr) {
...
+void Memory::store(const expr &p, const StateValue &v, const Type &type,
                    unsigned align) {
-  StateValue val = type.toBV(v);
-  unsigned bits = val.bits();
-  unsigned bytes = divide_up(bits, 8);
-
-  val = val.zext(bytes * 8 - bits);
-
   Pointer ptr(*this, p);
-  ptr.is_dereferenceable(bytes, align);
```
nunoplopes: please don't move this. I like having preconditions first.

ir/memory.cpp
```diff
@@ -296,79 +416,70 @@ void Memory::free(const expr &ptr) {
...
 
-  for (unsigned i = 0; i < bytes; ++i) {
-    // FIXME: right now we store in little-endian; consider others?
-    expr data  = val.value.extract((i + 1) * 8 - 1, i * 8);
-    expr np    = val.non_poison.extract((i + 1) * 8 - 1, i * 8);
-    blocks_val = blocks_val.store((ptr + i).release(), np.concat(data));
+  vector<Byte> bytes;
+  valueToBytes(v, type, *this, bytes);
+  for (unsigned i = 0; i < bytes.size(); ++i) {
```
nunoplopes: i = 0, e = bytes.size()

ir/memory.cpp
```diff
@@ -296,79 +416,70 @@ void Memory::free(const expr &ptr) {
...
   }
+  ptr.is_dereferenceable(bytes.size(), align);
 }
 
-StateValue Memory::load(const expr &p, Type &type, unsigned align) {
-  auto bits = type.bits();
-  unsigned bytes = divide_up(bits, 8);
+StateValue Memory::load(const expr &p, const Type &type, unsigned align) {
   Pointer ptr(*this, p);
-  ptr.is_dereferenceable(bytes, align);
```
nunoplopes: same comment here

ir/memory.cpp
```diff
@@ -296,79 +416,70 @@ void Memory::free(const expr &ptr) {
...
 
-  StateValue val;
-  bool first = true;
+  bool isPtrTy = type.isPtrType();
```
nunoplopes: this variable isn't needed; just inline

ir/memory.cpp
```diff
@@ -296,79 +416,70 @@ void Memory::free(const expr &ptr) {
...
+  auto bitsize = isPtrTy ? bitsMemorySize() : type.bits();
```
nunoplopes: I'm confused why this is needed. Is PtrType.bits() wrong?

ir/memory.cpp
aqjune: This is helpful when reader doesn't know what bitsByte() is, but I don't stick to it, so removed it

ir/memory.cpp
aqjune: I'm fine with any convention, so updated it

ir/memory.cpp
aqjune: I couldn't find it - StateValue's extract is not pushed to master yet?

ir/memory.cpp
aqjune: thanks, updated. :)

ir/memory.cpp
aqjune: Makes sense, I updated it and looks better

ir/memory.cpp
aqjune: done

ir/memory.cpp
aqjune: `PtrType.bits()` returns 64 + 8 + 8, so bitsPtrSize() which returns 64 was needed.

ir/memory.h
aqjune: Makes sense, moved it to memory.cpp

ir/memory.h
aqjune: updated

llvm_util/llvm2alive.cpp
aqjune: Hmm, I'm still a fan of printing Unsupported operation when meeting inttoptr, but anyway it is not related with this patch, so reverted

tests/alive-tv/memory/punning-ptrint.tgt.ll
aqjune: Updated, and added malloc null check as well

aqjune: (This comment was originally attached to BB2's ret, but dislocated after the update)

aqjune:
Fails: 140 -> 162, time: 282 -> 291 secs.
Here are some TODOs in the future:
- Load/store of aggregate values fails, and it is related with its non_poison definition. This requires update of AggregateType's fromBV(), etc.
- Time-out happens when a longer example is given. Using lambda to encode pointer bytes instead of just storing Byte one by one is helpful, performance benefit of which was observed from my Python implementation.

ir/memory.cpp
nunoplopes: ah, never mind. please ignore

Issue #119, opened 2019-10-05
The initial memory cannot contain pointers to local block ids


aqjune:
The initial memory cannot have pointer bytes with block ids newly allocated after beginning.

```
def void @f(i32** %p) {
%q = call malloc(i32 4)
%p2 = load i32*, i32** %p ; p2.bid cannot be q.bid
}
```

This can be supported after pointer-byte thingy (#118) is added.

PR #121, opened 2019-10-06
Let the initial memory have pointers to non-local blocks only

aqjune:
This is a small patch (with a bunch of examples) that addresses #119 .
This patch resolves following 11 llvm unit test fails:

```
Transforms/GVN/rle-nonlocal.ll
Transforms/GVNHoist/pr28626.ll
Transforms/GlobalOpt/malloc-promote-1-no-null-opt.ll
Transforms/GlobalOpt/memset-null.ll
Transforms/InstCombine/all-bits-shift.ll
Transforms/InstCombine/load.ll
Transforms/JumpThreading/phi-known.ll
Transforms/NewGVN/rle-nonlocal.ll
Transforms/SLPVectorizer/X86/crash_gep.ll
Transforms/SimplifyCFG/PR27615-simplify-cond-br.ll
Transforms/SimplifyCFG/PhiEliminate3.ll <- ‘invalid argument’ error disappeared
```
They previously failed because reading a pointer argument yielded a local pointer, as described in #119.

Following 3 tests now fail, but it is expected:
```
Transforms/CorrelatedValuePropagation/icmp.ll: transforms `add` to `add nsw nuw` which is incorrect if input is undef 
Transforms/InstSimplify/result-of-usub-by-nonzero-is-non-zero-and-no-overflow.ll: should deal with nonnull tag.
Transforms/SimplifyCFG/branch-fold-threshold.ll: transforms `false` to `and false, poison`
```
\# of llvm unit test failures: 163 -> 155 , running time: 281 -> 285 secs.


ir/memory.cpp
```diff
@@ -385,6 +385,15 @@ Memory::Memory(State &state) : state(&state) {
...
     blocks_liveness = expr::mkLambda({ bid }, move(val));
   }
+  {
+    // All non-local blocks cannot initially contain pointers to local blocks.
+    Pointer idx(*this, "#idx0");
+    auto byte = Byte(*this, blocks_val.load(idx()));
+    Pointer loadedptr(*this, byte.ptr_value());
+    expr cond = idx.is_local() || byte.is_ptr() == 0 ||
+                byte.ptr_nonpoison() == 0 || !loadedptr.is_local();
+    state.addPre(blocks_val.mkForAll({ idx() }, move(cond)));
```
nunoplopes: mkForAll is a static function.  There's something wrong here.

aqjune:
Updated, running llvm unit tests

aqjune:
Yep, the unit tests result didn't change (155 fails). :)

Issue #122, opened 2019-10-06
"Error: invalid argument"


LebedevRI:
This seems to be recent regression
```
----------------------------------------
Name: D68360 urem expand
  %r = urem i8 %x, %C
  ret i8 %r
=>
  %C_wide = zext i8 %C to i16
  %C_Magic_tmp = udiv i16 -1, %C_wide
  %C_Magic = add i16 %C_Magic_tmp, 1
  %x_wide = zext i8 %x to i16
  %Lowbits = mul i16 %x_wide, %C_Magic
  %Lowbits_wider = zext i16 %Lowbits to i32
  %C_wider = zext i8 %C to i32
  %Result_full = mul i32 %Lowbits_wider, %C_wider
  %Result_high = lshr i32 %Result_full, 16
  %r = trunc i32 %Result_high to i8
  ret i8 %r

Error: invalid argument
```

LebedevRI:
```
fa2f7cd70b733efab5466e0018812d9ff30beda6 is the first bad commit
commit fa2f7cd70b733efab5466e0018812d9ff30beda6
Author: Juneyoung Lee <aqjune@gmail.com>
Date:   Sun Oct 6 23:25:12 2019 +0900

    Let the initial memory have pointers to non-local blocks only (#121)

```

nunoplopes:
I can't repro this with latest master. Are you on latest master?
The commit you point out is unrelated with urem. Shouldn't make any difference

LebedevRI:
I get that with fa2f7cd70b733efab5466e0018812d9ff30beda6 which is latest master,
and don't get that with the commit just before it (1e116db710e8cb5553bd360fb9eba18800ac16de),
so are _you_ one a latest master?

nunoplopes:
Yes, I'm on `fa2f7cd`. Maybe try a clean build?
If it still happens, can you run alive with `-v -smt-verbose` so we see where it crashes?

nunoplopes:
Also, does `ninja check` passes? Because with that commit, I would expect most tests to fail if a rem-related test fails.

LebedevRI:
> Maybe try a clean build?

Sadly, that is **always** the first thing i try. :(

> Also, does `ninja check` pass?

Hm, no, it does not, neither does it pass on 1e116db.
Iet me try with newer z3.

LebedevRI:
http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=941857

Issue #123, opened 2019-10-06
urem UB propagation issues?


LebedevRI:
With #122 "resolved" (by checking-out previous commit), i'm seeing the following:
```
Name: D68360 urem expand
  %r = urem i8 %x, %C
  ret i8 %r
=>
  %C_wide = zext i8 %C to i16
  %C_Magic_tmp = udiv i16 -1, %C_wide
  %C_Magic = add i16 %C_Magic_tmp, 1
  %x_wide = zext i8 %x to i16
  %Lowbits = mul i16 %x_wide, %C_Magic
  %Lowbits_wider = zext i16 %Lowbits to i32
  %C_wider = zext i8 %C to i32
  %Result_full = mul i32 %Lowbits_wider, %C_wider
  %Result_high = lshr i32 %Result_full, 16
  %r = trunc i32 %Result_high to i8
  ret i8 %r

ERROR: Value mismatch for i8 %r

Example:
i8 %x = #x12 (18)
i8 %C = #x00 (0)
i16 %C_wide = #x0000 (0)
i16 %C_Magic_tmp = #xffff (65535, -1)
i16 %C_Magic = #x0000 (0)
i16 %x_wide = #x0012 (18)
i16 %Lowbits = #x0000 (0)
i32 %Lowbits_wider = #x00000000 (0)
i32 %C_wider = #x00000000 (0)
i32 %Result_full = #x00000000 (0)
i32 %Result_high = #x00000000 (0)
Source value: #x12 (18)
Target value: #x00 (0)
```
But `i8 %C = #x00 (0)`, LHS would be undefined.
So why is it complaining?

nunoplopes:
With latest master I get this:
```
Name: D68360 urem expand
  %r = urem i8 %x, %C
  ret i8 %r
=>
  %C_wide = zext i8 %C to i16
  %C_Magic_tmp = udiv i16 -1, %C_wide
  %C_Magic = add i16 %C_Magic_tmp, 1
  %x_wide = zext i8 %x to i16
  %Lowbits = mul i16 %x_wide, %C_Magic
  %Lowbits_wider = zext i16 %Lowbits to i32
  %C_wider = zext i8 %C to i32
  %Result_full = mul i32 %Lowbits_wider, %C_wider
  %Result_high = lshr i32 %Result_full, 16
  %r = trunc i32 %Result_high to i8
  ret i8 %r

ERROR: Value mismatch for i8 %r

Example:
i8 %x = #x02 (2)
[...]

PR #124, opened 2019-10-06
CMake: bump required version of Z3 to 4.8.5

LebedevRI:
Fixes https://github.com/AliveToolkit/alive2/issues/122
References https://github.com/AliveToolkit/alive2/issues/123


LebedevRI:
Is `.5` also considered good?

Issue #120, opened 2019-10-06
alive-tv fails with src/tgt having same code when malloc is involved


aqjune:
```
$ pwd
/Users/juneyounglee/alive2/tests/alive-tv/memory
$ ../../../build/alive-tv malloc-null.src.ll malloc-null.src.ll 

----------------------------------------
define i64 @malloc_null() {
%0:
  %ptr = malloc i64 1
  %i = ptrtoint * %ptr to i64
  ret i64 %i
}
=>
define i64 @malloc_null() {
%0:
  %ptr = malloc i64 1
  %i = ptrtoint * %ptr to i64
  ret i64 %i
}
Transformation doesn't verify!
ERROR: Value mismatch

Example:

Source:
* %ptr = pointer(local, block_id=256, offset=0)
i64 %i = #x0000000000001000 (4096)

Target:
* %ptr = pointer(local, block_id=256, offset=0)
i64 %i = #x0000002000000000 (137438953472)
Source value: #x0000000000001000 (4096)
Target value: #x0000002000000000 (137438953472)
```
The reason is because the addresses of local blocks are not quantified?
Bug found by @Dongjoo-Kim

aqjune:
BTW, what do you think about making `ninja check` run `alive-tv` with same files (e.g. alive-tv <a.ll> <a.ll>)?

nunoplopes:
> BTW, what do you think about making ninja check run alive-tv with same files (e.g. alive-tv <a.ll> <a.ll>)?

That's a good idea, yes!

nunoplopes:
There's this new test failure that might be related: `InstCombine/all-bits-shift.ll`

aqjune:
`InstCombine/all-bits-shift.ll` was related with #119 . I made PR  #121 that resolves it.

Regarding the identity check, I'll make a PR that also contains fix to the `malloc-null.src.ll` bug. :)

nunoplopes:
This is working now. Thanks for discovering this issue!

PR #126, opened 2019-10-08
Let storing to const globals UB, support extern const globals

aqjune:
This PR
- Makes storing to const global UB (by adding `blocks_readonly` array)
- Supports extern const globals (which do not have initial values)
- Adds a few more tests.

Unit tests: 152 -> 153 fails; the failure is llvm-test/Transforms/GlobalOpt/invariant.ll .
What it does is pretty interesting:
```
@object1 = global i32 0
define void @ctor1() {
  store i32 -1, i32* @object1
  %A = bitcast i32* @object1 to i8*
  call void @test1(i8* %A)
  ret void
}
```
=>
```
@object1 = constant i32 -1
define void @ctor1() local_unnamed_addr {
  store i32 -1, i32* @object1
  %A = bitcast i32* @object1 to i8*
  call void @test1(i8* %A)
  ret void
}
```

But I think this is wrong, because LLVM LangRef states that const global var can be placed at read-only section of an executable:
```
A variable may be defined as a global constant, which indicates that the contents of the variable will never be modified (enabling better optimization, allowing the global data to be placed in the read-only section of an executable, etc).
```

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -280,7 +280,8 @@ expr Pointer::is_aligned(unsigned align) const {
 }
 
 // When bytes is 0, pointer is always derefenceable
-void Pointer::is_dereferenceable(const expr &bytes0, unsigned align) {
+void Pointer::is_dereferenceable(const expr &bytes0, unsigned align,
+                                 bool iswrite) {
   expr block_sz = block_size();
   expr offset = get_offset().sextOrTrunc(m.bits_size_t);
   expr bytes = bytes0.zextOrTrunc(m.bits_size_t);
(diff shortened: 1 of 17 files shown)


nunoplopes:
Nice patch! :)
Yes, I agree that test looks fishy. The store might be left there because instcombine/dce will clean it up, but the output of globalopt is incorrect. Do you want to report that to LLVM?

BTW, a potential improvement would be to change the new array to take only a non-local bid, since only that can be const. Then change the Pointer::is_readonly to return is_nonlocal() && load(nonlocal_bid()).
It has 2 advantages: reduces the number of bits in the domain of the array, and 2) it exposes the fact that local objects are never read-only. That should enable simplification of the SMT expressions (it's easier than through the array).

aqjune:
> Do you want to report that to LLVM?

Sure. :) I made a report.

> It has 2 advantages: reduces the number of bits in the domain of the array, and 2) it exposes the > fact that local objects are never read-only. That should enable simplification of the SMT expressions (it's easier than through the array).

Great! I'll include this in a PR for #120 because it seems to be related with quantifying variables like blks_addr for local blocks of src.

PR #140, opened 2019-10-21
do not check declarations

regehr:
do not check functions that contain zero instructions


regehr:
I looked at Function.h and missed the isDeclaration() method, argh!!

PR #141, opened 2019-10-21
Support for insertelement, extractelement and shufflevector

zhengyangl:
There is still some work left for shufflevector. I'll get it done soon :)


regehr:
nice!!
one thing you may want to start doing now is adding tests for the LLVM parser part of alive2, for example by invoking alive-tv, see tests/alive-tv
basically you provide two .ll files, one refining (or not) the other and pass them to this tool

zhengyangl:
Hi @nunoplopes This pull request is ready for review. However, I have a question about poison semantic on insertelement (it is not listed on Langref), which I made assumptions in this PR but I am not sure:

For insertelement: if the index is poison is out-of-bounds or a poison, what should the result be? For both cases,  this PR produces a vector with full poison as the result of the instruction.

regehr:
@zhengyangl the choice you made seems like the most sensible one. the only other possible choice is immediate UB.

ir/instr.cpp
```diff
@@ -1865,4 +1865,121 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
   return make_unique<Memcpy>(*dst, *src, *bytes, align_dst, align_src, move);
 }
 
+vector<Value*> ExtractElement::operands() const {
+  return { a, idx };
```
nunoplopes: I would rename a to v, since it's a vector.

ir/instr.cpp
```diff
@@ -1865,4 +1865,121 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+  os << getName() << " = insertelement" << *a << ", " << *e << ", " << *idx;
+}
+
+StateValue InsertElement::toSMT(State &s) const {
+  auto &[iv, ip] = s[*idx];
+  auto aty = a->getType().getAsAggregateType();
+  expr inbounds = iv.ult(aty->numElementsConst());
+  auto [rv, rp] = aty->update(s[*a], s[*e], iv);
+  return { move(rv), expr::mkIf(ip && inbounds,
+                                move(rp), expr::mkInt(-1, aty->bits())) };
```
nunoplopes: this breaks the abstraction.
I guess the update function should do this, since anyway the case of what happens on idx overflow needs to be considered.

ir/instr.cpp
```diff
@@ -1865,4 +1865,121 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+  os << getName() << " = shufflevector " << *a << ", " << *b << ", " << *m;
+}
+
+StateValue ShuffleVector::toSMT(State &s) const {
+  auto aty = a->getType().getAsAggregateType();
+  auto mty = m->getType().getAsAggregateType();
+  auto ty = getType().getAsAggregateType();
+  vector<StateValue> vals;
+
+  for (unsigned i = 0, e = mty->numElementsConst(); i!= e; ++i) {
```
nunoplopes: missing space in i!=e

ir/instr.cpp
```diff
@@ -1865,4 +1865,121 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+         idx->getType().enforceIntType();
+}
+
+unique_ptr<Instr> InsertElement::dup(const string &suffix) const {
+  return make_unique<InsertElement>(getType(), getName() + suffix, *a, *e, *idx);
+}
+
+
+vector<Value*> ShuffleVector::operands() const {
+  return { a, b, m };
```
nunoplopes: m->mask

ir/instr.cpp
```diff
@@ -1865,4 +1865,121 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+  for (unsigned i = 0, e = mty->numElementsConst(); i!= e; ++i) {
+    auto [iv, ip] = mty->extract(s[*m], i);
+    expr sz = expr::mkUInt(aty->numElementsConst(), 32);
+    auto [av, ap] = aty->extract(s[*a], iv.urem(sz));
+    auto [bv, bp] = aty->extract(s[*b], iv.urem(sz));
+    expr inbounds = iv.ult(sz + sz);
+    expr v = expr::mkIf(iv.ult(sz), move(av), move(bv));
+    expr np = ip && inbounds && expr::mkIf(iv.ult(sz), move(ap), move(bp));
+    StateValue sv = { move(v), move(np) };
+    vals.emplace_back(move(sv));
```
nunoplopes: please combine the last 2 lines to avoid the last move.  emplace_back(move(v), move(np).

ir/instr.cpp
```diff
@@ -1865,4 +1865,121 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+  auto aty = a->getType().getAsAggregateType();
+  auto mty = m->getType().getAsAggregateType();
+  auto ty = getType().getAsAggregateType();
+  vector<StateValue> vals;
+
+  for (unsigned i = 0, e = mty->numElementsConst(); i!= e; ++i) {
+    auto [iv, ip] = mty->extract(s[*m], i);
+    expr sz = expr::mkUInt(aty->numElementsConst(), 32);
+    auto [av, ap] = aty->extract(s[*a], iv.urem(sz));
+    auto [bv, bp] = aty->extract(s[*b], iv.urem(sz));
```
nunoplopes: let's try to avoid urem, since that's expensive. Something like:
If(iv >= sz, extract(b, iv - sz), extract(a, iv))

ir/instr.cpp
```diff
@@ -1865,4 +1865,121 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+    expr inbounds = iv.ult(sz + sz);
+    expr v = expr::mkIf(iv.ult(sz), move(av), move(bv));
```
nunoplopes: Please use ule/uge whenever possible (cheaper in Z3)

ir/type.cpp
```diff
@@ -162,6 +162,14 @@ expr Type::enforceVectorTypeIff(const Type &other) const {
   return !other.enforceVectorType() || (enforceVectorType() && elems);
 }
 
+expr Type::enforceVectorTypeSameChildTy(const Type &other) const {
+  expr elems = false;
+  if (auto agg = getAsAggregateType())
+    if (auto agg2 = other.getAsAggregateType())
+      elems = (agg->getChild(0) == agg2->getChild(0));
```
nunoplopes: return agg-> ...

ir/type.cpp
```diff
@@ -162,6 +162,14 @@ expr Type::enforceVectorTypeIff(const Type &other) const {
...
+  return elems;
```
nunoplopes: return false

ir/type.cpp
```diff
@@ -622,6 +630,30 @@ StateValue AggregateType::extract(const StateValue &val,
                             (val.non_poison << idx).extract(h, l) });
 }
 
+StateValue AggregateType::update(const StateValue &val,
+                                 const StateValue &n,
+                                 const expr &index) const {
+  auto &elementTy = *children[0];
```
nunoplopes: what happens if Agg has no children?  Can't happen with vectors but can happen with structs.
Either this is moved to struct type or it needs to handle the zero children case

ir/type.cpp
```diff
@@ -622,6 +630,30 @@ StateValue AggregateType::extract(const StateValue &val,
...
+  for (unsigned i = 1; i < elements; ++i) {
+    assert(elementTy.bits() == children[i]->bits());
+  }
+
+  unsigned bw_elem = elementTy.bits();
+  unsigned bw_val = val.value.bits();
```
nunoplopes: just bits() (which is equal to val.bits() and equal to val.value.bits()).  Assert below isn't needed

ir/type.cpp
```diff
@@ -622,6 +630,30 @@ StateValue AggregateType::extract(const StateValue &val,
...
+  assert(bw_val == val.non_poison.bits());
+
+  expr shift = (expr::mkUInt(elements - 1, bw_val) - index.zextOrTrunc(bw_val)) *
+    expr::mkUInt(bw_elem, bw_val);
+  expr mask = ~(expr::mkInt(-1, bw_elem).zext(bw_val - bw_elem) << shift);
+
+  StateValue n_bv = elementTy.toBV(n);
+  expr nv_shifted = (n_bv.value.zext(bw_val - bw_elem) << shift);
```
nunoplopes: no () needed around the whole expression

ir/type.cpp
```diff
@@ -622,6 +630,30 @@ StateValue AggregateType::extract(const StateValue &val,
...
+  auto &elementTy = *children[0];
+  for (unsigned i = 1; i < elements; ++i) {
+    assert(elementTy.bits() == children[i]->bits());
+  }
+
+  unsigned bw_elem = elementTy.bits();
+  unsigned bw_val = val.value.bits();
+  assert(bw_val == val.non_poison.bits());
+
+  expr shift = (expr::mkUInt(elements - 1, bw_val) - index.zextOrTrunc(bw_val)) *
```
nunoplopes: you can get rid of the subtraction by doing lshr instead of <<. You need a concat instead of zext, though.

nunoplopes:
> For insertelement: if the index is out-of-bounds or poisonous, what should the result be? For both cases,  this PR produces a vector with full poison as the result.

http://llvm.org/docs/LangRef.html#insertelement-instruction

```If idx exceeds the length of val for a fixed-length vector, the result is a poison value.```

My worry was if it would be UB, but it doesn't seem to be. At least judging by the x64 assembly: https://godbolt.org/z/uhXlX2
(it masks the idx)

File: ir/instr.cpp [hunk 14]
```diff
@@ -1865,4 +1865,121 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+  auto &[iv, ip] = s[*idx];
+  auto aty = a->getType().getAsAggregateType();
+  expr inbounds = iv.ult(aty->numElementsConst());
+  auto [rv, rp] = aty->extract(s[*a], iv);
+  return { move(rv), ip && inbounds && rp };
+}
+
+expr ExtractElement::getTypeConstraints(const Function &f) const {
+  return Value::getTypeConstraints() &&
+         a->getType().enforceVectorType([&](auto &ty) { return ty == getType(); }) &&
```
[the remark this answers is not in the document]
zhengyangl: fixed:)

ir/instr.cpp
```diff
@@ -1865,4 +1865,121 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
   return make_unique<Memcpy>(*dst, *src, *bytes, align_dst, align_src, move);
 }
```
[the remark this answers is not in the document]
zhengyangl: fixed :)

ir/instr.cpp
```diff
@@ -1865,4 +1865,121 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+  expr inbounds = iv.ult(aty->numElementsConst());
+  auto [rv, rp] = aty->update(s[*a], s[*e], iv);
+  return { move(rv), expr::mkIf(ip && inbounds,
+                                move(rp), expr::mkInt(-1, aty->bits())) };
+}
+
+expr InsertElement::getTypeConstraints(const Function &f) const {
+  return Value::getTypeConstraints() &&
+         getType() == a->getType() &&
+         a->getType().enforceVectorType([&](auto &ty) { return ty == e->getType(); }) &&
```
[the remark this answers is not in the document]
zhengyangl: fixed:)

ir/instr.cpp
zhengyangl: fixed :)

ir/instr.cpp
zhengyangl: fixed:)

ir/instr.cpp
zhengyangl: fixed:)

ir/instr.cpp
zhengyangl: fixed:)

ir/instr.cpp
zhengyangl: fixed:)

ir/instr.cpp
zhengyangl: fixed:)

ir/type.cpp
zhengyangl: fixed:)

ir/type.cpp
zhengyangl: fixed:)

ir/type.cpp
zhengyangl: fixed:)

ir/type.cpp
zhengyangl: fixed:)

ir/instr.cpp
zhengyangl: The the extract() from AggregateTypes does not consider idx overflow cases, either. I was just following that.

zhengyangl: If you prefer handle the o.o.b. cases inside update, I can fix that :)

ir/type.cpp
zhengyangl: extract() has the same "always have children" assumptions. I can move both of them if you agree:)

ir/instr.cpp
```diff
@@ -1911,4 +1911,125 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+}
+
+void ExtractElement::print(ostream &os) const {
+  os << getName() << " = extractelement " << *v << ", " << *idx;
+}
+
+StateValue ExtractElement::toSMT(State &s) const {
+  auto &[iv, ip] = s[*idx];
+  auto vty = v->getType().getAsAggregateType();
+  expr inbounds = !iv.uge(vty->numElementsConst());
```
nunoplopes: you can use ult here.

ir/instr.cpp
```diff
@@ -1911,4 +1911,125 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+  vector<StateValue> vals;
+
+  for (unsigned i = 0, e = mty->numElementsConst(); i != e; ++i) {
+    auto [iv, ip] = mty->extract(s[*mask], i);
+    expr sz = expr::mkUInt(vty->numElementsConst(), 32);
+    auto [lv, lp] = vty->extract(s[*v1], iv);
+    auto [rv, rp] = vty->extract(s[*v2], iv - sz);
+    expr v = expr::mkIf(iv.uge(sz), rv, lv);
+    expr np = expr::mkIf(iv.uge(sz), rp, lp);
+    expr inbounds = !iv.uge(sz + sz);
```
nunoplopes: ult

ir/instr.cpp
```diff
@@ -1911,4 +1911,125 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+}
+
+void InsertElement::print(ostream &os) const {
+  os << getName() << " = insertelement" << *v << ", " << *e << ", " << *idx;
+}
+
+StateValue InsertElement::toSMT(State &s) const {
+  auto &[iv, ip] = s[*idx];
+  auto vty = v->getType().getAsAggregateType();
+  expr inbounds = !iv.uge(vty->numElementsConst());
```
nunoplopes: ult

ir/instr.cpp
```diff
@@ -1911,4 +1911,125 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+
+StateValue ShuffleVector::toSMT(State &s) const {
+  auto vty = v1->getType().getAsAggregateType();
+  auto mty = mask->getType().getAsAggregateType();
+  auto ty = getType().getAsAggregateType();
+  vector<StateValue> vals;
+
+  for (unsigned i = 0, e = mty->numElementsConst(); i != e; ++i) {
+    auto [iv, ip] = mty->extract(s[*mask], i);
+    expr sz = expr::mkUInt(vty->numElementsConst(), 32);
```
nunoplopes: this can be hoisted out of the loop

ir/instr.cpp
```diff
@@ -1911,4 +1911,125 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+}
+
+void ShuffleVector::print(ostream &os) const {
+  os << getName() << " = shufflevector " << *v1 << ", " << *v2 << ", " << *mask;
+}
+
+StateValue ShuffleVector::toSMT(State &s) const {
+  auto vty = v1->getType().getAsAggregateType();
+  auto mty = mask->getType().getAsAggregateType();
+  auto ty = getType().getAsAggregateType();
```
nunoplopes: you can inline this in the return statement.

ir/instr.cpp
zhengyangl: > Please use ule/uge whenever possible (cheaper in Z3)

Hi @nunoplopes , On the last round of review, you mentioned that use "ule/uge" whenever possible. Does the performance difference between "equal" and "than" get fixed or am I missed something?

ir/instr.cpp
zhengyangl: fixed :)

ir/instr.cpp
zhengyangl: fixed :)

ir/instr.cpp
nunoplopes: Right, you should use uge/ule when it's possible. I mean, when it's possible to change the formula around such that you don't need to do !uge (which is the same as ult). When you really need to do ult, just use it.

zhengyangl: fixed :) Thanks for the explanation!!!

ir/type.cpp
nunoplopes: Nicely spot; extract (with non-constant idx) has the same problem.  I think it makes sense to move both to VectorType. None of these methods works with structs, and I don't think we need them for arrays(?)

ir/instr.cpp
nunoplopes: Right, for extract things were a bit easier because you get a single element, while for update you can't do the same.
Should we move the OOB check in both cases to their respective functions?  What do you think?
I really don't like exposing the BV representation since we certainly will want to play with different variations and we want everything in one place to avoid bugs.

nunoplopes: Ah, never mind. Better suggestion: you can use Type::getDummyValue(false).non_poison to get your all-poison value.

zhengyangl: I am more concern about what will be the layer for the pluggable UB support work upon? If we're planning to make the pluggable UB works at instr level, then handling OOB outside update() may be more reasonable.

ir/type.cpp
zhengyangl: Okay, I'll move both of them to the VectorType class.

zhengyangl: done

ir/type.cpp
```diff
@@ -764,6 +757,47 @@ VectorType::VectorType(string &&name, unsigned elements, Type &elementTy)
   children.resize(elements, &elementTy);
 }
 
+StateValue VectorType::extract(const StateValue &val,
+                               const expr &index) const {
+  auto &elementTy = *children[0];
+  for (unsigned i = 1; i < elements; ++i) {
+    assert(elementTy.bits() == children[i]->bits());
```
nunoplopes: this is not needed anymore; it's enforced by the vector type

ir/type.cpp
```diff
@@ -764,6 +757,47 @@ VectorType::VectorType(string &&name, unsigned elements, Type &elementTy)
...
+  return elementTy.fromBV({ (val.value << idx).extract(h, l),
+                            (val.non_poison << idx).extract(h, l) });
+}
+
+StateValue VectorType::update(const StateValue &val,
+                              const StateValue &n,
+                              const expr &index) const {
+  auto &elementTy = *children[0];
+  for (unsigned i = 1; i < elements; ++i) {
+    assert(elementTy.bits() == children[i]->bits());
```
nunoplopes: same here

PR #143, opened 2019-10-24
better encodings for funnel shifts

pranavk:
I was playing with the code and realized the existing encodings for funnel shifts times out in some cases for higher bits. The newly added file in this patch for funnel shift tests (same as folds.opt but with 32 bits inputs minus some timeout cases) times out in majority of the cases or takes more time with earlier encodings but works fine with this patch. 

It still takes takes time for simple transformations (for higher bits) involving a constant though. I have omitted those test cases from new fold32.opt. Not much idea why those are taking much time.

### DIFF
--- a/smt/expr.cpp
+++ b/smt/expr.cpp
@@ -576,18 +576,16 @@ expr expr::lshr(const expr &rhs) const {
 
 expr expr::fshl(const expr &a, const expr &b, const expr &c) {
   C2(a);
-  auto nbits = a.bits();
-  expr c_mod_width = c.urem(mkUInt(nbits, nbits)).zext(nbits);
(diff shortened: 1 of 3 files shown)


regehr:
new encoding looks reasonable

I encouraged pranav to measure timeouts on opt-fuzz output and verify that they are reduced

pranavk:
Dataset consisted of 9645 functions generated by `./opt-fuzz  --num-insns=2 --onebinop --oneicmp --cores=8 --num-files=1000000 --width=64`

Before this patch:
7064 timeouts

After this patch:
5393 timeouts

nunoplopes:
The complexity of shift is n log n, so it makes sense to reduce the bw of the shift. It seems that outweighs the additional subtraction.

PR #144, opened 2019-10-24
Let local memory blocks' addresses be quantified, run identical inputs for alive-tv

aqjune:
This patch addresses #120. Addresses of local memory blocks are declared as quantifier through State::addQuant().

I added one more flag to State::addPre which states whether it is a precondition for a quantified variable or not.
This is needed because their representation inside the refinement expression is different from that of preconditions for non-qualifiers. I think preconditions for quantifiers should be represented as follows: 
`forall (quants-for-tgt), precond(quants-for-tgt) -> exists (quants-for-src), precond(quants_for_src) /\ …`

Also, lit test is updated so it runs `alive-tv <input.ll> <input.ll>` (which are identical inputs) for all .ll files.
`-disable-undef-input ` were added to three tests which raised `unknown`.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -201,25 +201,15 @@ expr Pointer::get_offset() const {
 
 expr Pointer::get_address() const {
   expr offset = get_offset().sextOrTrunc(m.bits_size_t);
-  auto local_name = m.mkName("blks_addr");
-  return
-    offset +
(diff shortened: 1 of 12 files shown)


aqjune:
As discussed, I'll split this PR. The first PR will be moving uninterpreted functions from arrays.

PR #149, opened 2019-10-28
Declare blocks_kind / blocks_readonly as UF

aqjune:
This patch changes arrays to UF if they are constant over program execution.
For example, `blks_kind` (which returns whether a memory block is defined at heap or not) has its value never changed after allocated, so it is moved to UF.
In contrary, `blks_liveness` may change during execution (e.g. `free(ptr)`), so it is kept as an array.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -322,7 +322,7 @@ expr Pointer::is_block_alive() const {
 
 expr Pointer::is_at_heap() const {
   // TODO: use bitvector instead of UF if there are few allocations
-  return get_value_from_uf("blks_at_heap", false, LOCAL_BLOCKS);
+  return get_value_from_uf("blks_at_heap", false, BOTH);
 }
 
 expr Pointer::is_readonly() const {
@@ -460,6 +460,7 @@ expr Memory::alloc(const expr &size, unsigned align, BlockKind blockKind,
     // The memory block was initially alive.
     state->addAxiom(mk_liveness_array().load(p.get_bid()));
     state->addAxiom(p.is_readonly() == expr(blockKind == CONSTGLOBAL));
+    state->addAxiom(p.is_at_heap() == expr(false));
   }
 
   return p.release();
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -196,14 +196,19 @@ expr Pointer::get_offset() const {
   return p.extract(m.bitsBid() + m.bits_for_offset - 1, m.bitsBid());
 }
 
+expr Pointer::get_value_from_uf(const char *name, const expr &ret,
+                                MemBlkCategory category) const {
+  auto local_name = m.mkName(name);
+  auto localret = category == NONLOCAL_BLOCKS ? ret :
+                  expr::mkUF(local_name.c_str(), { get_local_bid() }, ret);
+  auto nonlocalret = category == LOCAL_BLOCKS ? ret :
+                     expr::mkUF(name, { get_nonlocal_bid() }, ret);
+  return expr::mkIf(is_local(), localret, nonlocalret);
+}
+
 expr Pointer::get_address() const {
   expr offset = get_offset().sextOrTrunc(m.bits_size_t);
-  auto local_name = m.mkName("blks_addr");
-  return
-    offset +
-      expr::mkIf(is_local(),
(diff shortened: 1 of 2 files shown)


ir/memory.h
```diff
@@ -105,15 +113,12 @@ class Memory {
...
   smt::expr blocks_liveness; // array: bid -> bool
-  smt::expr blocks_kind; // array: bid -> uint(1bit), 1 if heap, 0 otherwise
-  smt::expr blocks_readonly; // array: bid -> bool, true if readonly
 
   std::string mkName(const char *str, bool src) const;
   std::string mkName(const char *str) const;
 
-  smt::expr mk_val_array(const char *name) const;
+  smt::expr mk_val_array() const;
   smt::expr mk_liveness_uf() const;
```
aqjune: As mentioned in this PR text, liveness of a memory block is not constant during execution due to the existence of free() call.
Does this mean declaring liveness as UF has no benefit than having an array? Then, I'd like to change the name of this to `mk_liveness_array`.

aqjune:
Result:

| Description                                      | Before | After |
|--------------------------------------------------|--------|-------|
|  Timeout                                         | 2140   | 2136  |
|  Invalid expr                                    | 1042   | 1042  |
|  program doesn't type check!                     | 681    | 681   |
|  Loops are not supported yet! Skipping function. | 592    | 592   |
|  Source is more defined than target              | 64     | 64    |
|  Value mismatch                                  | 58     | 54    |
|  Target is more poisonous than source            | 19     | 20    |
|  Precondition is always false                    | 10     | 14    |
|  Out of memory; skipping function.               | 5      | 5     |

aqjune:
The +1 `Target is more poisonous than source` was from `GVN/PRE/rle.ll` which uses big-endian. I'll make a separate patch for this.

aqjune:
Some (about half) `Precondition is always false` are coming from type change of a global variable.
This happens when a global variable has a different size in src and tgt, leading to generation of a precondition like
```
(and ..
    (= (concat #b0 (blks_size #x01)) #x0000000000000004)
    (= (concat #b0 (blks_size #x01)) #x0000000000000001)
    ..)
```
.
We can omit optimizations that are known to be global by updating opt-alive.sh, but this is quite coarse grained in the sense that it can filter out innocent local transformations.
Do you want to let alive2 emit `Unsupported global variable change` in this case or some message like that?

ir/memory.h
nunoplopes: right, liveness has to be an array. Renaming sounds good! :)

aqjune:
Conflict resolved

ir/memory.cpp
```diff
@@ -321,11 +321,13 @@ expr Pointer::is_block_alive() const {
 }
 
 expr Pointer::is_at_heap() const {
-  return m.blocks_kind.load(get_bid()) == 1;
+  // TODO: use bitvector instead of UF if there are few allocations
+  return get_value_from_uf("blks_at_heap", false, LOCAL_BLOCKS);
```
nunoplopes: non-local blocks may also be heap allocated. e.g. a pointer received by argument might be a malloc'ed ptr.

nunoplopes:
> Some (about half) Precondition is always false are coming from type change of a global variable.
> This happens when a global variable has a different size in src and tgt, leading to generation of a precondition like
> (and ..
>     (= (concat #b0 (blks_size #x01)) #x0000000000000004)
>     (= (concat #b0 (blks_size #x01)) #x0000000000000001)
>     ..)
> 
> .
> We can omit optimizations that are known to be global by updating opt-alive.sh, but this is quite coarse grained in the sense that it can filter out innocent local transformations.
> Do you want to let alive2 emit Unsupported global variable change in this case or some message like that?

I agree that it would be best to emit an error and skip verification rather than keep increasing the blacklist in opt-alive.sh. If you think this is easy/cheap to check, these please send a separate PR :)

ir/memory.cpp
aqjune: You're right, fixed

aqjune:
> I agree that it would be best to emit an error and skip verification rather than keep increasing the blacklist in opt-alive.sh. If you think this is easy/cheap to check, these please send a separate PR :)

Sure, I'm happy to work on this. :)

PR #151, opened 2019-10-28
Support big endian

aqjune:
This PR adds support for big endian. This PR is orthogonal with #149.
Interestingly `GVN/PRE/rle.ll` (which was mentioned in the #149) still raises `Target is more poisonous than source`, which seems to be related with a bug of alive-tv with undef, because alive-tv is printing that source and targets' values are both `poison`.

### DIFF
--- a/ir/function.h
+++ b/ir/function.h
@@ -56,6 +56,8 @@ class Function final {
   std::unordered_map<std::string, BasicBlock> BBs;
   std::vector<BasicBlock*> BB_order;
 
+  bool little_endian;
+
   // constants used in this function
   std::vector<std::unique_ptr<Value>> constants;
   std::vector<std::unique_ptr<Predicate>> predicates;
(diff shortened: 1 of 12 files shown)


aqjune:
| master | endian | description                                     |
|--------|--------|-------------------------------------------------|
| 2140   | 2142   | Timeout                                         |
| 1042   | 1045   | Invalid expr                                    |
| 681    | 681    | program doesn't type check!                     |
| 592    | 595    | Loops are not supported yet! Skipping function. |
| 64     | 63     | Source is more defined than target              |
| 58     | 57     | Value mismatch                                  |
| 19     | 19     | Target is more poisonous than source            |
| 10     | 10     | Precondition is always false                    |
| 5      | 5      | Out of memory; skipping function.               |

Failures: 168 -> 166; Transforms/SROA/big-endian.ll and Transforms/DeadStoreElimination/simple.ll disappeared

PR #154, opened 2019-10-29
add calloc for alive-tv

Dongjoo-Kim:
Hello Nuno.
This PR implements calloc and adds some basic tests. Support for alive parser (not alive-tv) will be addressed in a separate PR.
I heard that malloc’s semantics will change in the future. calloc’s null-returning semantics is same to that of malloc, so I left TODO comment to the place.


ir/instr.cpp
```diff
@@ -1674,6 +1674,53 @@ unique_ptr<Instr> Malloc::dup(const string &suffix) const {
...
+void Calloc::print(std::ostream &os) const {
+  os << getName() << " = calloc " << *num << " " << *size;
+}
+
+StateValue Calloc::toSMT(State &s) const {
+  auto &[nm, np_num] = s[*num];
+  auto &[sz, np_sz] = s[*size];
+  s.addUB(np_num && np_sz);
+
+  s.addUB(nm.mul_no_uoverflow(sz));
```
nunoplopes: the semantics (according to the man page) is that calloc returns nullptr on overflow

ir/instr.cpp
```diff
@@ -1674,6 +1674,53 @@ unique_ptr<Instr> Malloc::dup(const string &suffix) const {
...
+
+  // TODO: check calloc align.
+  auto p = s.getMemory().alloc(nm * sz, 8, Memory::HEAP);
+
+  s.getMemory().memset(p, { expr::mkUInt(0, 8), true }, nm * sz, 1);
+
+  if (isNonNull)
```
nunoplopes: calloc is never non-null; there's no C++ throwing equivalent. So this functionality can be removed.

ir/instr.cpp
```diff
@@ -1674,6 +1674,53 @@ unique_ptr<Instr> Malloc::dup(const string &suffix) const {
...
+  return { num, size };
+}
+
+void Calloc::rauw(const Value &what, Value &with) {
+  RAUW(num);
+  RAUW(size);
+}
+
+void Calloc::print(std::ostream &os) const {
+  os << getName() << " = calloc " << *num << " " << *size;
```
nunoplopes: should have a comma between num & size.

Dongjoo-Kim:
I reflected your comments, but some modification is needed.

Even calloc's arguments multiplication is overflowed, alloc is executed. (inst.cpp, Line 1705)

I try to set the server for LLVM testing.

ir/instr.cpp
```diff
@@ -1674,6 +1674,50 @@ unique_ptr<Instr> Malloc::dup(const string &suffix) const {
...
+
+StateValue Calloc::toSMT(State &s) const {
+  auto &[nm, np_num] = s[*num];
+  auto &[sz, np_sz] = s[*size];
+  s.addUB(np_num && np_sz);
+
+  // TODO: check calloc align.
+  auto p = s.getMemory().alloc(nm * sz, 8, Memory::HEAP);
+
+  s.getMemory().memset(p, { expr::mkUInt(0, 8), true }, nm * sz, 1);
```
nunoplopes: you can only execute memset if is non-null

ir/instr.cpp
```diff
@@ -1674,6 +1674,50 @@ unique_ptr<Instr> Malloc::dup(const string &suffix) const {
...
+  os << getName() << " = calloc " << *num << ", " << *size;
+}
+
+StateValue Calloc::toSMT(State &s) const {
+  auto &[nm, np_num] = s[*num];
+  auto &[sz, np_sz] = s[*size];
+  s.addUB(np_num && np_sz);
+
+  // TODO: check calloc align.
+  auto p = s.getMemory().alloc(nm * sz, 8, Memory::HEAP);
```
nunoplopes: as you mentioned, this can only be run if there's no overlow

ir/instr.cpp
```diff
@@ -1674,6 +1674,50 @@ unique_ptr<Instr> Malloc::dup(const string &suffix) const {
...
+}
+
+void Calloc::print(std::ostream &os) const {
+  os << getName() << " = calloc " << *num << ", " << *size;
+}
+
+StateValue Calloc::toSMT(State &s) const {
+  auto &[nm, np_num] = s[*num];
+  auto &[sz, np_sz] = s[*size];
+  s.addUB(np_num && np_sz);
```
nunoplopes: I feel that yielding UIB is too strong. Just realized that malloc does the same, but that's not well justified. Malloc won't crash or anything if poison is passed as input.
I think we should just return poison in this case.

tests/alive-tv/memory/calloc-init.src.ll
```diff
@@ -0,0 +1,19 @@
+target datalayout = "e-m:o-i64:64-f80:128-n8:16:32:64-S128"
+
+define i8 @calloc_init() {
+  %ptr = call noalias i8* @calloc(i64 1, i64 1)
+  %i = ptrtoint i8* %ptr to i64
+  %cmp = icmp eq i64 %i, 0
+  br i1 %cmp, label %BB1, label %BB2
+BB1:
+  ret i8 0
```
nunoplopes: maybe return 1 here to test it can go to the other branch

ir/instr.cpp
nunoplopes: I'll fix malloc.

ir/instr.cpp
aqjune: In this code getMemory().alloc() is returning a valid pointer, so memset does not raise UB.

ir/instr.cpp
aqjune: I think this is related with whether calloc can be optimized to malloc + memset in LLVM. If only a portion of allocated bytes of calloc() is read, it is benefial to do memset on only those bytes. memset(src, 0, poison) is currently defined as UB.

ir/instr.cpp
nunoplopes: not if OOM happens

Dongjoo-Kim:
I reflect your comments in the code.
Allocation retuns null pointer if there is overflow.
And the pointer of memset with 0 size can be null pointer, so I make memset size 0 if there is overflow.

ir/instr.cpp
```diff
@@ -8,6 +8,7 @@
 #include "smt/solver.h"
 #include "util/compiler.h"
 #include <functional>
+#include <iostream>
```
nunoplopes: this is probably a debug leftover

Dongjoo-Kim: Thank you for your checking.
I fixed it and removed unused argument in the 'Calloc' class and a test.

ir/instr.cpp
```diff
@@ -1714,6 +1714,51 @@ unique_ptr<Instr> Malloc::dup(const string &suffix) const {
...
+}
+
+void Calloc::print(std::ostream &os) const {
+  os << getName() << " = calloc " << *num << ", " << *size;
+}
+
+StateValue Calloc::toSMT(State &s) const {
+  auto &[nm, np_num] = s[*num];
+  auto &[sz, np_sz] = s[*size];
+  expr calloc_sz = expr::mkIf(nm.mul_no_uoverflow(sz), expr(nm * sz),
```
nunoplopes: no need for expr(.) here

ir/instr.cpp
```diff
@@ -1714,6 +1714,51 @@ unique_ptr<Instr> Malloc::dup(const string &suffix) const {
...
+                              expr::mkUInt(0, sz.bits()));
+
+  // TODO: check calloc align.
+  auto p = s.getMemory().alloc(nm * sz, 8, Memory::HEAP, std::nullopt,
+                               nullptr, nm.mul_no_uoverflow(sz));
```
nunoplopes: pls hoist computation of mul_no_overflow to avoid computing twice

ir/instr.cpp
```diff
@@ -1714,6 +1714,51 @@ unique_ptr<Instr> Malloc::dup(const string &suffix) const {
...
+  
+  // If memset's size is zero, then ptr can be NULL.
+  s.getMemory().memset(p, { expr::mkUInt(0, 8), true }, calloc_sz, 1);
```
nunoplopes: this needs to check if p is null (just make the size=0 if null)

ir/instr.cpp
```diff
@@ -1714,6 +1714,51 @@ unique_ptr<Instr> Malloc::dup(const string &suffix) const {
...
+  s.addQuantVar(flag);
+  return { expr::mkIf(move(flag), nullp.release(), move(p)),
+           expr(np_num && np_sz) };
+}
+
+expr Calloc::getTypeConstraints(const Function &f) const {
+  return Value::getTypeConstraints() &&
+         getType().enforcePtrType() &&
+         num->getType().enforceIntType() &&
+         size->getType().enforceIntType();
```
nunoplopes: num and size need to have same type

ir/memory.cpp
```diff
@@ -462,7 +462,8 @@ pair<expr, vector<expr>> Memory::mkInput(const char *name) {
 }
 
 expr Memory::alloc(const expr &size, unsigned align, BlockKind blockKind,
-                   optional<unsigned> bidopt, unsigned *bid_out) {
+                   optional<unsigned> bidopt, unsigned *bid_out,
+                   expr precond) {
```
nunoplopes: pass this by const expr &

nunoplopes:
btw, can you rebase to make sure the buildbot issues go away?  Thanks!

Dongjoo-Kim:
I fix the code, size of memset in calloc must be 0, when allocated ptr is null.
Also fix the calloc-init test which is missed.

PR #155, opened 2019-10-29
Check global vars' sizes & constness in src and tgt

aqjune:
This PR adds checking of src/tgt's global variables' sizes and constness.
This resolves all 14 `Precondition is always false` error.

### DIFF
--- a/tools/alive-tv.cpp
+++ b/tools/alive-tv.cpp
@@ -209,11 +209,6 @@ static bool compareFunctions(llvm::Function &F1, llvm::Function &F2,
   t.src = move(*Func1);
   t.tgt = move(*Func2);
   TransformVerify verifier(t, false);
-  if (Errors errs = verifier.sync()) {
-    cerr << "Transformation doesn't verify!\n" << errs << endl;
-    ++errorCount;
-    return true;
-  }
 
   t.print(cout, print_opts);
 
--- a/tools/alive-tv.cpp
+++ b/tools/alive-tv.cpp
@@ -209,6 +209,12 @@ static bool compareFunctions(llvm::Function &F1, llvm::Function &F2,
   t.src = move(*Func1);
   t.tgt = move(*Func2);
   TransformVerify verifier(t, false);
+  if (Errors errs = verifier.sync()) {
+    cerr << "Transformation doesn't verify!\n" << errs << endl;
+    ++errorCount;
+    return true;
+  }
+
   t.print(cout, print_opts);
 
   {
--- a/tools/alive-tv.cpp
+++ b/tools/alive-tv.cpp
@@ -209,7 +209,6 @@ static bool compareFunctions(llvm::Function &F1, llvm::Function &F2,
   t.src = move(*Func1);
   t.tgt = move(*Func2);
   TransformVerify verifier(t, false);
-
   t.print(cout, print_opts);
 
   {
--- a/tools/transform.cpp
+++ b/tools/transform.cpp
@@ -306,34 +306,34 @@ TransformVerify::TransformVerify(Transform &t, bool check_each_var) :
 Errors TransformVerify::sync() {
   try {
     t.tgt.syncDataWithSrc(t.src);
-
-    // Check sizes of global variables
-    auto globals_tgt = t.tgt.getGlobalVars();
-    auto globals_src = t.src.getGlobalVars();
-    for (auto GVS: globals_src) {
-      auto I = find_if(globals_tgt.begin(), globals_tgt.end(),
-        [GVS](auto *GV) -> bool { return GVS->getName() == GV->getName(); });
-      if (I == globals_tgt.end())
-        continue;
-
-      auto GVT = *I;
-      if (GVS->size() != GVT->size()) {
-        stringstream ss;
-        ss << "Global variable " << GVS->getName() << " in source and "
-          << "target has different sizes (" << GVS->size() << ", "
-          << GVT->size() << ")";
-        throw AliveException(ss.str(), false);
-      } else if (GVS->isConst() != GVT->isConst()) {
-        stringstream ss;
-        ss << "Global variable " << GVS->getName() << " in source and "
-          << "target has different constness";
-        throw AliveException(ss.str(), false);
-      }
-    }
   } catch (AliveException &ae) {
     return Errors(move(ae));
   }
-  return Errors();
+
+  // Check sizes of global variables
+  auto globals_tgt = t.tgt.getGlobalVars();
+  auto globals_src = t.src.getGlobalVars();
+  for (auto GVS : globals_src) {
+    auto I = find_if(globals_tgt.begin(), globals_tgt.end(),
+      [GVS](auto *GV) -> bool { return GVS->getName() == GV->getName(); });
+    if (I == globals_tgt.end())
+      continue;
+
+    auto GVT = *I;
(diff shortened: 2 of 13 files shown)


tools/transform.cpp
```diff
@@ -301,8 +301,39 @@ TransformVerify::TransformVerify(Transform &t, bool check_each_var) :
       tgt_instrs.emplace(i.getName(), &i);
     }
   }
+}
+
+Errors TransformVerify::sync() {
```
nunoplopes: can't we do this in verify() instead?

tools/transform.cpp
```diff
@@ -301,8 +301,39 @@ TransformVerify::TransformVerify(Transform &t, bool check_each_var) :
...
+        continue;
 
-  t.tgt.syncDataWithSrc(t.src);
+      auto GVT = *I;
+      if (GVS->size() != GVT->size()) {
+        stringstream ss;
+        ss << "Global variable " << GVS->getName() << " in source and "
+          << "target has different sizes (" << GVS->size() << ", "
+          << GVT->size() << ")";
+        throw AliveException(ss.str(), false);
```
nunoplopes: you can return an error right away, there's no need to throw

tools/transform.cpp
```diff
@@ -301,8 +301,39 @@ TransformVerify::TransformVerify(Transform &t, bool check_each_var) :
...
+        [GVS](auto *GV) -> bool { return GVS->getName() == GV->getName(); });
+      if (I == globals_tgt.end())
+        continue;
 
-  t.tgt.syncDataWithSrc(t.src);
+      auto GVT = *I;
+      if (GVS->size() != GVT->size()) {
+        stringstream ss;
+        ss << "Global variable " << GVS->getName() << " in source and "
+          << "target has different sizes (" << GVS->size() << ", "
```
nunoplopes: maybe reshuffle the sentence a bit: Unsupported transformation: global variable foo has different size in source and target (xx vs yy bytes)

tools/transform.cpp
```diff
@@ -301,8 +301,39 @@ TransformVerify::TransformVerify(Transform &t, bool check_each_var) :
...
+}
+
+Errors TransformVerify::sync() {
+  try {
+    t.tgt.syncDataWithSrc(t.src);
+
+    // Check sizes of global variables
+    auto globals_tgt = t.tgt.getGlobalVars();
+    auto globals_src = t.src.getGlobalVars();
+    for (auto GVS: globals_src) {
```
nunoplopes: spacing: auto foo : bar

tools/transform.cpp
```diff
@@ -301,8 +301,39 @@ TransformVerify::TransformVerify(Transform &t, bool check_each_var) :
...
+      if (GVS->size() != GVT->size()) {
+        stringstream ss;
+        ss << "Global variable " << GVS->getName() << " in source and "
+          << "target has different sizes (" << GVS->size() << ", "
+          << GVT->size() << ")";
+        throw AliveException(ss.str(), false);
+      } else if (GVS->isConst() != GVT->isConst()) {
+        stringstream ss;
+        ss << "Global variable " << GVS->getName() << " in source and "
+          << "target has different constness";
```
nunoplopes: Maybe give a more precise message by case splitting, e.g.:
Unsupported transformation: Global variable foo is constant in target but not in source.

The reverse transformation (const->non const) should yield an unsound error (it's a bug!).

tools/transform.cpp
```diff
@@ -301,8 +301,39 @@ TransformVerify::TransformVerify(Transform &t, bool check_each_var) :
...
+        throw AliveException(ss.str(), false);
+      }
+    }
+  } catch (AliveException &ae) {
+    return Errors(move(ae));
+  }
+  return Errors();
```
nunoplopes: return {};

tools/transform.cpp
aqjune: This involves changing `blks_readonly` because the address is currently shared between source and target - having different constness of same global variable in src and tgt will make precondition always false.
Do you prefer having a separate PR that relaxes the condition to nonconst -> const, or address it in this PR?

tools/transform.cpp
aqjune: Updated

tools/transform.cpp
aqjune: Moved sync() contents to verify()

tools/transform.cpp
aqjune: The bug https://bugs.llvm.org/show_bug.cgi?id=43616 was successfully found because previously `blocks_readonly` was an array.
I think I can make `blks_readonly` be defined in source and target separately, so there are two `blks_readonly` UFs.

nunoplopes: Non-const -> Const: we can't verify this
Const -> Non-const: the transformation  is wrong.

We should distinguish these cases.

tools/alive-tv.cpp
```diff
@@ -209,6 +209,7 @@ static bool compareFunctions(llvm::Function &F1, llvm::Function &F2,
   t.src = move(*Func1);
   t.tgt = move(*Func2);
   TransformVerify verifier(t, false);
+
```
nunoplopes: unrelated change

tools/transform.cpp
```diff
@@ -301,11 +301,45 @@ TransformVerify::TransformVerify(Transform &t, bool check_each_var) :
...
+    auto GVT = *I;
+    if (GVS->size() != GVT->size()) {
+      stringstream ss;
+      ss << "Global variable " << GVS->getName() << " has different size in"
+        << " source and target (" << GVS->size() << " vs " << GVT->size()
+        << " bytes)";
+      return { ss.str(), false };
+    } else if (GVS->isConst() && !GVT->isConst()) {
+      stringstream ss;
+      ss << "Transformation is wrong because global variable " << GVS->getName()
```
nunoplopes: let's use incorrect instead of wrong to be nicer :)

tools/transform.cpp
```diff
@@ -301,11 +301,45 @@ TransformVerify::TransformVerify(Transform &t, bool check_each_var) :
...
+  for (auto GVS : globals_src) {
+    auto I = find_if(globals_tgt.begin(), globals_tgt.end(),
+      [GVS](auto *GV) -> bool { return GVS->getName() == GV->getName(); });
+    if (I == globals_tgt.end())
+      continue;
+
+    auto GVT = *I;
+    if (GVS->size() != GVT->size()) {
+      stringstream ss;
+      ss << "Global variable " << GVS->getName() << " has different size in"
```
nunoplopes: Unsupported interprocedural transformation:

tools/transform.cpp
```diff
@@ -301,11 +301,45 @@ TransformVerify::TransformVerify(Transform &t, bool check_each_var) :
...
+        << " bytes)";
+      return { ss.str(), false };
+    } else if (GVS->isConst() && !GVT->isConst()) {
+      stringstream ss;
+      ss << "Transformation is wrong because global variable " << GVS->getName()
+        << " is const in source but not in target";
+      return { ss.str(), true };
+    } else if (!GVS->isConst() && GVT->isConst()) {
+      stringstream ss;
+      ss << "Can't verify this because global variable " << GVS->getName()
```
nunoplopes: Unsupported interprocedural transformation: global variable foo

PR #156, opened 2019-10-29
Let refinement correctly encode precondition on source's quantifier

aqjune:
This is the second PR that splits #144. This is orthogonal with other opened PRs.

This patch let preconditions for source quantifiers be in the existential predicate inside the refinement.
```
forall (quants-for-tgt), precond(quants-for-tgt) -> exists (quants-for-src), precond(quants_for_src) /\ refines(src_output, tgt_output)
```

This hasn't been problematic till now, because there was no case where precondition on quantifiers was added, IIUC.

But this patch is needed for the next splitted PR, which is to set the address of a memory block as a quantifier (as the result of `freeze` does) to resolve issue #120.
Unlike `freeze`'s result, blocks' addresses need preconditions (blocks' disjointness, non-zero-ness). This PR helps them to be correctly encoded.

Note that this is not related with malloc's null semantics, because this is talking about preconditions of addresses of successfully allocated blocks.

### DIFF
--- a/tools/transform.cpp
+++ b/tools/transform.cpp
@@ -271,7 +271,6 @@ static void check_refinement(Errors &errs, Transform &t,
   expr srcpre_quant = src_state.getPreOnQuantifiers();
 
   auto [poison_cnstr, value_cnstr] = type.refines(a, b);
-  expr pre_dom = pre && (dom_a && dom_b);
   expr dom_a_b = dom_a && dom_b;
 
   Solver::check({
@@ -292,7 +291,7 @@ static void check_refinement(Errors &errs, Transform &t,
       }}
   });
 
-  if (check_expr(axioms && preprocess(t, qvars, uvars, pre && srcpre_quant))
+  if (check_expr(axioms && preprocess(t, {}, {}, pre && srcpre_quant))
       .isUnsat()) {
     errs.add("Precondition is always false", false);
   }
--- a/tools/transform.cpp
+++ b/tools/transform.cpp
@@ -266,27 +266,34 @@ static void check_refinement(Errors &errs, Transform &t,
   // so order here matters
   src_state.startParsingPre();
   expr pre = t.precondition ? t.precondition->toSMT(src_state) : true;
-  pre &= src_state.getPre() && tgt_state.getPre();
+  pre &= tgt_state.getPre() && src_state.getPre()
+          && tgt_state.getPreOnQuantifiers();
+  expr srcpre_quant = src_state.getPreOnQuantifiers();
 
   auto [poison_cnstr, value_cnstr] = type.refines(a, b);
   expr pre_dom = pre && (dom_a && dom_b);
+  expr dom_a_b = dom_a && dom_b;
 
   Solver::check({
-    { axioms && preprocess(t, qvars, uvars, pre && dom_a.notImplies(dom_b)),
+    { axioms && preprocess(t, qvars, uvars, pre &&
+                           srcpre_quant.implies(dom_a.notImplies(dom_b))),
       [&](const Result &r) {
         err(r, false, "Source is more defined than target");
       }},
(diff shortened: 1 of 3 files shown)


tools/transform.cpp
```diff
@@ -266,27 +266,34 @@ static void check_refinement(Errors &errs, Transform &t,
   // so order here matters
   src_state.startParsingPre();
   expr pre = t.precondition ? t.precondition->toSMT(src_state) : true;
-  pre &= src_state.getPre() && tgt_state.getPre();
+  pre &= tgt_state.getPre() && src_state.getPre()
+          && tgt_state.getPreOnQuantifiers();
+  expr srcpre_quant = src_state.getPreOnQuantifiers();
 
   auto [poison_cnstr, value_cnstr] = type.refines(a, b);
   expr pre_dom = pre && (dom_a && dom_b);
```
nunoplopes: this variable doesn't seem to be used anymore

tools/transform.cpp
```diff
@@ -266,27 +266,34 @@ static void check_refinement(Errors &errs, Transform &t,
...
-    { axioms && preprocess(t, qvars, uvars, pre_dom && !value_cnstr),
+    { axioms && preprocess(t, qvars, uvars, pre &&
+                           srcpre_quant.implies(dom_a_b && !value_cnstr)),
       [&](const Result &r) {
         err(r, true, "Value mismatch");
       }}
   });
 
-  if (check_expr(axioms && preprocess(t, qvars, uvars, move(pre))).isUnsat()) {
+  if (check_expr(axioms && preprocess(t, qvars, uvars, pre && srcpre_quant))
```
nunoplopes: this one doesn't look correct since in the previous formulas srcpre_quant is allowed to be false (thus it's under implication). Quantified vars need to change I guess (to be existential)?

nunoplopes:
I'm a bit concerned/puzzled with this patch.
We have axioms, which are outside of any quantifier since they usually have their own quantifier.
Then we have preconditions, which can refer to exists/forall vars. For example, if you pass an undef to malloc, the code will add a precondition with a quantified var.
Why is this ok there but not for this new use case?  I guess it's not ok either for the current use case. This needs auditing.

tools/transform.cpp
aqjune: Fixed to make source quantifiers be existential.

tools/transform.cpp
aqjune: Fixed

aqjune:
If I understand the question correctly - the current state(master branch) is not okay with `p = malloc(undef)` as well.
For example, in this example ([malloc-undef.ll.zip](https://github.com/AliveToolkit/alive2/files/3787650/malloc-undef.ll.zip), running `alive-tv malloc-undef.ll malloc-undef.ll` raises `ERROR: Precondition is always false`.
The reason is that generated refinement contains following forall:
```
forall ((undef_1 (_ BitVec 64)) (malloc_isnull!0 Bool))
             (and ..
                  (= undef_1 (concat #b0 (blks_size_src #x01)))
              ..)
```
which makes this `forall` false.

This PR is a step forward addressing such issue. To fully resolve the `malloc(undef)` case, I think `blks_size_src` should be quantified, as #144 quantified `blks_addr_src`. But, quantifying a function isn't well supported in SMT solver, and this can be addressed by quantifying newly allocated sizes / addresses, not quantifying `blks_size_src` function.

tools/transform.cpp
aqjune: How about this? This would address the issue.
```
-  if (check_expr(axioms && preprocess(t, {}, {}, pre && srcpre_quant))
+  if (check_expr(axioms && preprocess(t, qvars, uvars,
+                                      srcpre_quant.implies(pre)))
```

nunoplopes:
I believe this is now fixed for good! :)

aqjune:
Thank you so much! I'll upload PR that adds disjointness of addresses today.

Issue #145, opened 2019-10-26
StateValue's non_poison has an inverted value when converted to BV


aqjune:
```
StateValue Type::toBV(StateValue v) const {
  expr val = toBV(move(v.value));
  auto bw = val.bits();
  return { move(val),
           expr::mkIf(v.non_poison, expr::mkUInt(0, bw), expr::mkInt(-1, bw)) };
}
```

Is this intended? I'll add a comment about this to `non_poison` field in the upcoming PR then.

nunoplopes:
It was intended because being non-poison should be the common case, hence 0. I didn't measure the effect, though
Anyway, the users to StateValue aren't suppose to notice the difference. It's all hidden under these toBV/fromBV APIs. It should, at least.

aqjune:
The non-poison value was linked to Byte's non-poison bits definition because Byte was created from the result of toBV API.
I'll add a comment to Byte's definition instead.

PR #157, opened 2019-11-01
Correctly initialize local memory

aqjune:
Correctly initialize local memory. Previously it was initialized as 0. :/ This is related with #145


nunoplopes:
Thanks for the test case. I've fixed it here: https://github.com/AliveToolkit/alive2/commit/bdb3b2812d44c487e74f3980a34b58226a9d221d

Please disentangle perf improvements from correctness fixes. If you think that swapping values of poison/non-poison is beneficial, please open a separate issue/PR.

PR #158, opened 2019-11-01
Implement memcmp / bcmp

aqjune:
This PR implements memcmp and bcmp.
Poison-ness of memcmp(p, q, size) is defined as follows:
(1) If all bytes in [p, p+size), [q, q+size) are non-poison, it is not poison.
(2) If there exists `i` s.t. all bytes in [p, p+i], [q, q+i] are non-poison && *(p+i) != *(q+i) && forall j<i, *(p+j) == (q+j).
e.g) `01 02 pp pp` and `01 03 pp pp` (`pp` is a poison byte)

I didn't use a recursive function because it did not work well for unknown reason. It didn't stop even if timeout is given, at least.
This PR includes #157. If #157 is accepted, this PR will be updated to not include it.

memcmp can return an arbitrary positive (and negative) number if the bytes aren't the same (`man memcmp` says the returned value is a subtraction of the different bytes, but it says a well-formed program should not depend on it). To guarantee that it returns non-zero number, #156 is needed I believe, so this PR has dependency on #156. This PR implements the non-zeroness of the result with a simple bit mask (it is not fully general), but the unit tests in this PR still work.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1739,7 +1739,7 @@ StateValue Calloc::toSMT(State &s) const {
 
   expr is_null = p == Pointer::mkNullPointer(s.getMemory())();
   expr calloc_sz = expr::mkIf(is_null, expr::mkUInt(0, sz.bits()), size);
-  
+
   // If memset's size is zero, then ptr can be NULL.
   s.getMemory().memset(p, { expr::mkUInt(0, 8), true }, calloc_sz, 1);
 
@@ -1993,12 +1993,49 @@ StateValue Memcpy::toSMT(State &s) const {
 
 expr Memcpy::getTypeConstraints(const Function &f) const {
   return dst->getType().enforcePtrType() &&
-         dst->getType().enforcePtrType() &&
+         src->getType().enforcePtrType() &&
          bytes->getType().enforceIntType();
 }
 
 unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
   return make_unique<Memcpy>(*dst, *src, *bytes, align_dst, align_src, move);
+}
+
+vector<Value*> Memcmp::operands() const {
+  return { ptr1, ptr2, num };
+}
+
+void Memcmp::rauw(const Value &what, Value &with) {
+  RAUW(ptr1);
+  RAUW(ptr2);
+  RAUW(num);
+}
+
+void Memcmp::print(ostream &os) const {
+  os << getName() << " = " << (equalchk ? "bcmp " : "memcmp ") << *ptr1
+     << ", " << *ptr2 << ", " << *num;
+}
+
+StateValue Memcmp::toSMT(State &s) const {
+  auto &[vptr1, np1] = s[*ptr1];
+  auto &[vptr2, np2] = s[*ptr2];
+  auto &[vnum, npn] = s[*num];
+  s.addUB(vnum.ugt(0).implies(np1 && np2));
+  s.addUB(npn);
+
+  unsigned bw = getType().bits();
+  return s.getMemory().memcmp(vptr1, vptr2, vnum, equalchk, bw);
+}
+
+expr Memcmp::getTypeConstraints(const Function &f) const {
+  return ptr1->getType().enforcePtrType() &&
+         ptr2->getType().enforcePtrType() &&
+         num->getType().enforceIntType();
+}
+
+unique_ptr<Instr> Memcmp::dup(const string &suffix) const {
+  return make_unique<Memcmp>(getType(), getName() + suffix, *ptr1, *ptr2, *num,
+                             equalchk);
 }
 
 
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -80,6 +80,22 @@ public:
 
   const expr& operator()() const { return p; }
 
+  expr no_poison_bit() const {
+    IntType ty("", 8);
+    auto v = ty.toBV(ty.getDummyValue(true));
+    return expr::mkIf(is_ptr(), ptr_nonpoison(),
+                      nonptr_nonpoison() == v.non_poison);
+  }
+
+  expr as_int8(optional<function<expr(const expr&)>> p2i = nullopt) const {
+    Pointer p(m, ptr_value());
+    auto three = expr::mkUInt(3, m.bitsPtrSize());
+    return expr::mkIf(is_ptr(),
+      (p.get_address(true, p2i).lshr(
+            ptr_byteoffset().zext(m.bitsPtrSize() - 3) << three).trunc(8)),
+      nonptr_value());
+  }
+
   static Byte mkPoisonByte(const Memory &m) {
     IntType ty("", 8);
     auto v = ty.toBV(ty.getDummyValue(false));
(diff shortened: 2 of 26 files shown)


tests/alive-tv/memory/memcmp-split.tgt.ll
```diff
@@ -0,0 +1,21 @@
...
+  %p_2 = getelementptr i8, i8* %p, i64 2
+  %q_2 = getelementptr i8, i8* %q, i64 2
+  %res2 = call i32 @memcmp(i8* %p_2, i8* %q_2, i64 2)
+  %cmp = icmp eq i32 %res1, 0
+  %b = select i1 %cmp, i32 %res2, i32 %res1
+  ret i32 %b
+}
+
+declare dso_local i32 @memcmp(i8* nocapture, i8* nocapture, i64) local_unnamed_addr #1
+attributes #1 = { nofree nounwind readonly "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "frame-pointer"="none" "less-precise-fpmad"="false" "no-infs-fp-math"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="false" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "unsafe-fp-math"="false" "use-soft-float"="false" }
```
nunoplopes: it's better to remove these attributes since there's plenty of churn there

tests/alive-tv/memory/memcmp-split.src.ll
```diff
@@ -0,0 +1,17 @@
+target datalayout = "e-p:64:64:64"
+; TEST-ARGS: -disable-undef-input
```
nunoplopes: I guess this is because the split increases # uses? If that's the case, please leave a comment

ir/type.cpp
```diff
@@ -337,6 +337,10 @@ pair<expr, vector<expr>> IntType::mkInput(State &s, const char *name) const {
 }
 
 void IntType::printVal(ostream &os, State &s, const expr &e) const {
+  if (!e.isConst()) {
```
nunoplopes: this should never happen. What's the problem?

tests/alive-tv/memory/memcmp-bcmp.src.ll
```diff
@@ -0,0 +1,19 @@
+target datalayout = "e-p:64:64:64"
+target triple = "x86_64-unknown-linux-gnu"
+; TEST-ARGS: -disable-undef-input
```
nunoplopes: is this needed?

tests/alive-tv/memory/memcmp-fail-1.src.ll
```diff
@@ -0,0 +1,18 @@
+target datalayout = "e-m:o-i64:64-f80:128-n8:16:32:64-S128"
+; TEST-ARGS: -disable-undef-input
```
nunoplopes: is this needed?

tests/alive-tv/memory/memcmp-opt-littleendian-fail.src.ll
```diff
@@ -0,0 +1,21 @@
...
+  %q = alloca i64
+  store i64 %x, i64* %p, align 1 ; 20 10 00 00 ..
+  store i64 %y, i64* %q, align 1 ; 30 00 00 00 ..
+  %p8 = bitcast i64* %p to i8*
+  %q8 = bitcast i64* %q to i8*
+  %res = call i32 @memcmp(i8* %p8, i8* %q8, i64 8) ; %res < 0
+  ret i32 %res
+}
+
+; ERROR: Value mismatch
```
nunoplopes: I'm lost with this one. Why is this transformation wrong?

llvm_util/llvm2alive.cpp
```diff
@@ -181,6 +182,26 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
     if (!ty)
       return error(i);
 
+    llvm::LibFunc Func;
+    if (TLI.getLibFunc(*fn, Func) && TLI.has(Func)) {
```
nunoplopes: This code should be moved to known_fns.cpp

ir/instr.cpp
```diff
@@ -1903,12 +1903,50 @@ StateValue Memcpy::toSMT(State &s) const {
...
+}
+
+StateValue Memcmp::toSMT(State &s) const {
+  auto &[vptr1, np1] = s[*ptr1];
+  auto &[vptr2, np2] = s[*ptr2];
+  auto &[vnum, npn] = s[*num];
+  s.addUB(vnum.ugt(0).implies(np1 && np2));
+  s.addUB(npn);
+
+  unsigned bw = dynamic_cast<IntType &>(getType()).bits();
```
nunoplopes: the cast isn't needed

nunoplopes:
Regarding recursive functions, do you still have the code around? It sounds you found a bug in Z3, and we should fix it. The encoding with recursive functions will be much simpler & efficient than with all these quantifiers.

ir/type.cpp
aqjune: This case happened when the result had a quantifier inside; `m.eval()` inside `type.printVal(os, st, m.eval(val.value, true));` failed to evaluate the result.

aqjune:
I have memory.cpp that implements memcmp() using the recursive function. It is not fully correct because it just returns subtracted result & has no poison value encoded, but it is enough for testing.
[memory-recfun.cpp.zip](https://github.com/AliveToolkit/alive2/files/3801168/memory-recfun.cpp.zip)

tests/alive-tv/memory/memcmp-opt-littleendian-fail.src.ll
aqjune: This test shows that it is illegal in little-endian to transform memcmp to integer comparison. The reason is that the least significant byte is positioned at the first byte.
For example, if %x is 0x1020 and %y is 0x30, %x > %y, so the target returns positive number, whereas source returns negative number because 0x20 < 0x30.

tests/alive-tv/memory/memcmp-split.src.ll
aqjune: The main reason (other test files as well) was the timeout. I couldn't make it raise timeout unless -disable-undef-input was given.
I think the tests still make sense without the existence of undef. I'll leave comments about these.

ir/instr.cpp
```diff
@@ -1739,7 +1739,7 @@ StateValue Calloc::toSMT(State &s) const {
 
   expr is_null = p == Pointer::mkNullPointer(s.getMemory())();
   expr calloc_sz = expr::mkIf(is_null, expr::mkUInt(0, sz.bits()), size);
-  
+
```
aqjune: I guess Visual Studio Code is automatically remove the space whenever it is touched. :/
This is a good change anyway; I think convention is that empty line should not have space.

ir/memory.cpp
```diff
@@ -195,19 +211,17 @@ expr Pointer::get_offset() const {
   return p.extract(m.bitsOffset() - 1, 0).sextOrTrunc(m.bits_size_t);
 }
 
-expr Pointer::get_value(const char *name, const FunctionExpr &fn,
-                        const expr &ret_type) const {
-  auto bid = get_short_bid();
-  return expr::mkIf(is_local(), fn(bid), expr::mkUF(name, { bid }, ret_type));
-}
```
aqjune: I made Pointer::get_value a template function so that the type of `fn` is parameteric.

ir/memory.cpp
```diff
@@ -195,19 +211,17 @@ expr Pointer::get_offset() const {
...
-
-expr Pointer::get_address(bool simplify) const {
+expr Pointer::get_address(bool simplify,
+                          optional<function<expr(const expr&)>> local_addrs)
+                          const {
```
aqjune: `local_addrs` is given when `m.local_blk_addr` cannot be used due to the existence of quantifier.
The main motivation here was that, a declaration of recursive function cannot be nested inside forall expression, so this returns a wrong expression if `m.local_blk_addr` contains quantifiers.

aqjune:
I'll close this and start with strlen, which is simpler

PR #161, opened 2019-11-05
LLVM Function comparator in alive-tv

pranavk:
Use the function comparator directly from llvm which does more extensive function comparison than the existing one

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,7 +143,7 @@ target_link_libraries(alive PRIVATE ${ALIVE_LIBS})
 add_library(alive2 SHARED ${IR_SRCS} ${SMT_SRCS} ${TOOLS_SRCS} ${UTIL_SRCS} ${LLVM_UTIL_SRCS})
 
 if (BUILD_LLVM_UTILS OR BUILD_TV)
-  llvm_map_components_to_libnames(llvm_libs support core irreader analysis)
+  llvm_map_components_to_libnames(llvm_libs support core irreader analysis transformutils)
   target_link_libraries(alive2 PRIVATE ${llvm_libs})
   target_link_libraries(alive-tv PRIVATE ${ALIVE_LIBS_LLVM} ${llvm_libs})
 endif()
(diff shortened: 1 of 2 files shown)


regehr:
that is a good cleanup, thanks Pranav!

nunoplopes:
The old & new code have very different purposes. The new code seems to be just an optimization: skip verification of syntactically equal functions. This may or may not be a good idea: it increases the trust base, since now you need to trust this LLVM comparator function to be correct.
The old code was skipping functions that didn't have the same type signature (e.g., same number & type of arguments). Which is a good thing because Alive may crash if signatures don't match. Ideally it wouldn't, and it would flag this as an unsupported transformation (and then the removed code wouldn't be needed at all). But I don't think it will ATM.

Can you confirm if all of the above is correct pls?

regehr:
ah, thanks, good catch Nuno
you're correct (I was getting confused since I had an old version of alive-tv which used this comparator!)

@pranavk you can't delete that type comparison code

@nunoplopes I would expect that trusting this LLVM function is OK since it is used somewhere in the optimizer. the alternative is an alive-level comparator, which would be just fine and could be used elsewhere outside of this function, it's just more work to create.

If @pranavk should do the alive version, then he can work on that

pranavk:
fwiw, the llvm's function comparator (https://github.com/llvm/llvm-project/blob/master/llvm/include/llvm/Transforms/Utils/FunctionComparator.h#L91) always fails conservatively - erring on the side of claiming that two functions are different. I assumed it's never incorrect when it says two functions are equivalent.

nunoplopes:
Well, modulo bugs in the implementation :)
I would err on the safety side and not use that function. We want to find bugs in the compiler after all.
We are able to do some proofs syntactically already; just check the SMT stats and you will see that about half of queries don't reach Z3. Undef complicates things, and reduces the number of syntactic stuff you can do with the current SMT preprocessing. We would need heuristic quantifier instantiation of the SMT formulas to do this (not hard, but needs a tree traversal of the SMT formulas that we don't have ATM).

I'm fine with a simple syntactic check in Alive, as long as it doesn't try to be fancy (like incorporating partial semantics there). More semantic-based syntactic discharges should be done IMHO by folding SMT formulas to false. There's no duplication of implementation of semantics like this.

regehr:
so an alive-level tree comparison, @nunoplopes ? I think this would be a useful check to have.

nunoplopes:
> so an alive-level tree comparison, @nunoplopes ? I think this would be a useful check to have.

agreed!

PR #162, opened 2019-11-06
add support for struct indexing

pranavk:
it's possible that i am overlooking something and there's more to it but following the code around, struct indexing should already work with following change. Can you point to any other issues, if any, that needs to be addressed for this?

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -337,9 +337,14 @@ public:
       if (!op)
         return error(i);
 
-      // TODO: support for struct indexing
-      if (I.getStructTypeOrNull())
-        return error(i);
+      if (auto structTy = I.getStructTypeOrNull()) {
+        llvm::Value *vals[] = { llvm::ConstantInt::getFalse(i.getContext()),
+                                I.getOperand() };
+        auto offset = DL().getIndexedOffsetInType(structTy,
+                llvm::makeArrayRef(vals, 2));
+        gep->addIdx(offset, *make_intconst(1, 64));
+        continue;
+      }
 
       gep->addIdx(DL().getTypeAllocSize(I.getIndexedType()), *op);
     }
--- a/tests/alive-tv/memory/gep-struct1.src.ll
+++ b/tests/alive-tv/memory/gep-struct1.src.ll
@@ -1,0 +1,11 @@
+target datalayout = "e-i32:32-i128:32-i16:16"
+
+%0 = type { i32, i128, i16 }
+
+define i64 @0(%0* nocapture) {
+  %2 = getelementptr inbounds %0, %0* %0, i32 0, i32 2
+  %3 = ptrtoint %0* %0 to i64
+  %4 = ptrtoint i16* %2 to i64
+  %5 = sub i64 %4, %3
+  ret i64 %5
+}
--- a/tests/alive-tv/memory/gep-struct1-fail.src.ll
+++ b/tests/alive-tv/memory/gep-struct1-fail.src.ll
@@ -1,0 +1,13 @@
+target datalayout = "e-i32:32-i128:32-i16:16"
+
+%0 = type { i32, i128, i16 }
+
+define i64 @0(%0* nocapture) {
+  %2 = getelementptr inbounds %0, %0* %0, i32 0, i32 2
+  %3 = ptrtoint %0* %0 to i64
+  %4 = ptrtoint i16* %2 to i64
+  %5 = sub i64 %4, %3
+  ret i64 %5
+}
+
+; ERROR: Value mismatch
--- a/tests/alive-tv/memory/gep-struct1-fail.tgt.ll
+++ b/tests/alive-tv/memory/gep-struct1-fail.tgt.ll
@@ -1,0 +1,7 @@
+target datalayout = "e-i32:32-i128:32-i16:16"
+
+%0 = type { i32, i128, i16 }
+
+define i64 @0(%0* nocapture) {
+  ret i64 24
+}
--- a/tests/alive-tv/memory/gep-struct1.tgt.ll
+++ b/tests/alive-tv/memory/gep-struct1.tgt.ll
@@ -1,0 +1,7 @@
+target datalayout = "e-i32:32-i128:32-i16:16"
+
+%0 = type { i32, i128, i16 }
+
+define i64 @0(%0* nocapture) {
+  ret i64 20
+}
--- a/tests/alive-tv/memory/gep-struct2.src.ll
+++ b/tests/alive-tv/memory/gep-struct2.src.ll
@@ -1,0 +1,15 @@
+target datalayout = "e-i32:32-i128:32-i8:8"
+
+%0 = type [10 x { i32, i128, [5 x { i32, i8, i8 }] }]
+; C representation:
+; struct S1 { X(i32); Y(i8); Z(i8); }
+; struct S2 {A(i32); B(i128); S1[5]; }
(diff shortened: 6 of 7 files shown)


regehr:
interesting. I have not looked at this code. maybe @aqjune can comment.

aqjune:
Hi,
There is an issue with heterogeneous types involved in a struct, for example:
```
$ cat t.ll
%struct.T = type { i8, i32, i8 }

define i8* @f(%struct.T* %0) {
  %2 = getelementptr inbounds %struct.T, %struct.T* %0, i64 0, i32 2
  ret i8* %2
}
```
Running `alive-tv t.ll t.ll` will print this:
```
define * @f(* %0) {
%1:
  %2 = gep inbounds * %0, 12 x i64 0, 1 x i32 2
  ret * %2
}
```
which is problematic because the last index isn't 1 * 2.
Putting exact byte-offset when converting GetElementPtrInst from LLVM IR to GEP of Alive2 IR would work I guess.

pranavk:
Alright! I got it now. Will update it soon.

pranavk:
ok. this latest POC patch should fix the issue but not sure if this is what we want here. For an example like:

```
%struct.T = type { i8, i32, i8 }

define i8* @f(%struct.T* %0) {
  %2 = getelementptr inbounds %struct.T, %struct.T* %0, i64 0, i32 2
  ret i8* %2
}
```
this patch now returns:

```
define * @f(* %0) {
%1:
  %2 = gep inbounds * %0, 12 x i64 0, 8 x i32 0
  ret * %2
}
```

If this approach looks good, I can go ahead beautify the patch and make it ready for merge.

pranavk:
actually wait, i think i got the size thing wrong. I need to check this again. Will do tomorrow

pranavk:
Using a constant value of `1` and sz = offset of the member we want to refer to inside the struct.

```
%struct.T = type { i8, i32, i8 }

define i8* @f(%struct.T* %0) {
  %2 = getelementptr inbounds %struct.T, %struct.T* %0, i64 0, i32 2
  ret i8* %2
}
```
=>
```
define * @f(* %0) {
%1:
  %2 = gep inbounds * %0, 12 x i64 0, 8 x i32 1
  ret * %2
}
```

llvm_util/llvm2alive.cpp
```diff
@@ -332,9 +332,15 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
       if (!op)
         return error(i);
 
-      // TODO: support for struct indexing
-      if (I.getStructTypeOrNull())
-        return error(i);
+      if (auto structTy = I.getStructTypeOrNull()) {
+	auto const0 = llvm::ConstantInt::get(llvm::Type::getInt32Ty(i.getType()->getContext()), 0);
+	auto const1 = llvm::ConstantInt::get(llvm::Type::getInt32Ty(i.getType()->getContext()), 1);
```
aqjune: Can factor out `llvm::Type::getInt32Ty(i.getType()->getContext())` as both lines have the expression.
You'd like to check the indentation as well.

aqjune:
I'd suggest adding tests. I usually write them as an input to alive-tv. It will be great to write bitcode inputs that have `getelementptr`, `ptrtoint`, and `add`.
Other than this (and the comment that has minor suggestions) I have no other special suggestion.
One thing that I have in my mind is that I'm not sure whether the bitvector representation of struct respects padding bits or not, but I guess this is an orthogonal issue.

pranavk:
Yup. It definitely needs some love before we merge. I just quickly hacked it to show the approach. If there are no major issues, I will clean it up.

llvm_util/llvm2alive.cpp
```diff
@@ -332,9 +332,15 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
       if (!op)
         return error(i);
 
-      // TODO: support for struct indexing
-      if (I.getStructTypeOrNull())
-        return error(i);
+      if (auto structTy = I.getStructTypeOrNull()) {
+	auto const0 = llvm::ConstantInt::get(llvm::Type::getInt32Ty(i.getType()->getContext()), 0);
```
nunoplopes: it seems you've tabs here or smth

pranavk: yes, i will clean this up this weekend. It is not yet ready for review. Just hacked up something quick for an experiment.

pranavk: @nunoplopes okay. this is read for review now.

llvm_util/llvm2alive.cpp
pranavk: done.

aqjune:
Could you add a few more negative tests please?

tests/alive-tv/memory/gep-struct1.src.ll
```diff
@@ -0,0 +1,9 @@
+%0 = type { i32, i128, i16 }
```
nunoplopes: these tests need a data layout string otherwise they may on machines with different CPUs and/or ABIs

llvm_util/llvm2alive.cpp
```diff
@@ -332,9 +332,13 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
       if (!op)
         return error(i);
 
-      // TODO: support for struct indexing
-      if (I.getStructTypeOrNull())
-        return error(i);
+      if (auto structTy = I.getStructTypeOrNull()) {
+        llvm::Value *vals[] = { llvm::ConstantInt::getFalse(i.getContext()), I.getOperand() };
```
nunoplopes: these lines are way above 80 chars

llvm_util/llvm2alive.cpp
```diff
@@ -332,9 +332,13 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
+        auto offset = DL().getIndexedOffsetInType(structTy, llvm::makeArrayRef(vals, 2));
+        // FIXME: size bits shouldn't be a constant
```
nunoplopes: you can leave as 64 bits, since that's the return value of getIndexedOffsetInType() anyway. It will be truncated/extended as needed.

pranavk: I am not sure I follow. I am not doing anything to `offset`. The FIXME comment is for the argument to `make_intconst` which creates `1` of width 64.

tests/alive-tv/memory/gep-struct1.src.ll
aqjune: @pranavk Adding `target datalayout = "e-i32:32-i128:32-i16:16"` at the top of gep-struct1.src.ll (and gep-struct2.src.ll as well) will work.

aqjune: To add a negative test, copying the src/tgt files to `gep-struct1-fail.<src/tgt>.ll`, changing `20` in tgt with a bogus number, and adding
```
; ERROR: Value mismatch
```
at the end of `gep-struct1-fail.src.ll` will work.

pranavk: yes, will work on this PR later today or tomorrow.

llvm_util/llvm2alive.cpp
nunoplopes: Right, I mean: the comment can go away. I think it's fine to leave size fixed with 64 bits.

PR #167, opened 2019-11-10
Add disjointness of block addresses, blockaddr+size never overflows

aqjune:
This PR adds disjointness of block addresses as well as precondition stating that blockaddr+size never overflows.

Disjointness between a block and another non-local block that is never explicitly allocated is not enforced because I guess there is no reasonable code that can observe the size of such blocks (at least in standard library functions).

Tests for disjointness of mallocs is not added here, I'll add those after updating malloc's null-ness semantics.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -474,9 +474,15 @@ Memory::Memory(State &state, bool little_endian)
   // The last byte of the memory cannot be allocated because ptr + size
   // (where size is the size of the block and ptr is the beginning address of
   // the block) should not overflow.
-  local_avail_space = expr::mkUInt((1ull << 63) - 1, bitsPtrSize());
+
+  // The initially available space of local area. Src and tgt shares this
+  // value.
+  local_avail_space = expr::mkVar("local_avail_space", bitsPtrSize());
+  auto memsz_half = expr::mkUInt((1ull << 63), bitsPtrSize());
 
   if (state.isSource()) {
+    state.addAxiom(local_avail_space.ult(memsz_half));
+
     // Initialize a memory block for null pointer.
     // TODO: in twin memory model, this is not needed.
     auto nullPtr = Pointer::mkNullPointer(*this);
@@ -488,7 +494,6 @@ Memory::Memory(State &state, bool little_endian)
     unsigned non_local_bid_upperbound = 2;
     // Non-local blocks are disjoint.
     // Ignore null pointer block
-    auto memsz_half = expr::mkUInt((1ull << 63), bitsPtrSize());
     for (unsigned bid = 1; bid <= non_local_bid_upperbound; ++bid) {
       Pointer p1(*this, bid, false);
       for (unsigned bid2 = bid + 1; bid2 <= non_local_bid_upperbound; ++bid2) {
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -208,7 +208,13 @@ expr Pointer::get_address(bool simplify) const {
   if (simplify && (get_bid() == 0).isTrue())
     return offset;
 
-  return offset + get_value("blk_addr", m.local_blk_addr, offset);
+  auto bid = get_short_bid();
+  // Local block area is the upper half of the memory
+  // Non-local block area is the lower half
+  auto retty = expr::mkUInt(0, m.bitsPtrSize() - 1);
+  return offset + expr::mkIf(is_local(),
+      expr::mkUInt(1, 1).concat(m.local_blk_addr(bid)),
+      expr::mkUInt(0, 1).concat(expr::mkUF("blk_addr", { bid }, retty)));
 }
 
 expr Pointer::block_size(bool simplify) const {
@@ -438,7 +444,7 @@ static unsigned last_nonlocal_bid = 1;
 
 Memory::Memory(State &state, bool little_endian)
   : state(&state), little_endian(little_endian),
-    local_blk_addr(expr::mkUInt(0, bits_size_t)),
+    local_blk_addr(expr::mkUInt(0, bits_size_t - 1)),
     local_blk_size(expr::mkUInt(0, bits_size_t - 1)),
     local_blk_kind(expr::mkUInt(0, 2)) {
 
@@ -477,12 +483,9 @@ Memory::Memory(State &state, bool little_endian)
 
   // The initially available space of local area. Src and tgt shares this
   // value.
-  local_avail_space = expr::mkVar("local_avail_space", bitsPtrSize());
-  auto memsz_half = expr::mkUInt((1ull << 63), bitsPtrSize());
+  local_avail_space = expr::mkVar("local_avail_space", bitsPtrSize() - 1);
 
   if (state.isSource()) {
-    state.addAxiom(local_avail_space.ult(memsz_half));
-
     // Initialize a memory block for null pointer.
     // TODO: in twin memory model, this is not needed.
     auto nullPtr = Pointer::mkNullPointer(*this);
@@ -490,7 +493,7 @@ Memory::Memory(State &state, bool little_endian)
     state.addAxiom(nullPtr.block_size(false) == 0);
 
     // TODO: replace the magic number 2 with the result of analysis.
-    // This is just set as 2 to make existing unit tests run without timeout.
+    // This is just set as 2 to make existing unit tests run with less timeout.
     unsigned non_local_bid_upperbound = 2;
     // Non-local blocks are disjoint.
     // Ignore null pointer block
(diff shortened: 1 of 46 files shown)


aqjune:
The reason why only `last_nonlocal_bid` blocks were compared for disjointness of nonlocal blocks was that encoding disjointness using forall quantifier `forall all non_local_bid, ..` raised Z3 unknown.

BTW, I became to think that considering `last_nonlocal_bid` blocks only isn't sound if inttoptr is considered.  
For example,
```
f(i32* %p, i32* %q) {
%r = inttoptr (..) to i32*
store 10, %r
}
```
(Assuming that an integer does not carry provenance), the `store 10, %r` will do something conceptually like:

    If %r falls into %p blocks' addresses, store %r to %p's block
    else if %r falls into %q block's addresses, store %r to %q's block

If %p and %q are pointing to different memory blocks, and %p and %q blocks are *not* disjoint, the semantics of this store will depend on which block to check first, which is strange.
But, in this example, `last_nonlocal_bid` is 0 (because there is no global vars), so the disjointness between %p and %q blocks is never encoded.

To resolve this issue - I think encoding disjointness of `num_def_pointers` memory blocks instead of `last_nonlocal_bid` will work. `num_def_pointers` is the number of pointer variable definitions. In the above example, there are 3 `num_def_pointers` - %p, %q, and %r. I think this makes sense because without loop, we cannot access more than the number of pointer variables. This scheme is also used in the previous Python implementation, and it worked pretty well.
What do you think?

aqjune:
Okay, this is getting pretty complex. :/
If nonlocal blocks are initially filling whole address space, upcoming local blocks' disjoint condition always becomes false. This causes the precondition of source to be always false, leading to spurious result.
To resolve this, I can come up with following two naive solutions:
(1) Restrict the size of nonlocal blocks more, e.g., it cannot be larger than 2^32, which will be enough for 64-bit memory space.
(2) Let `allocated` inside Memory::alloc consider memory layout. But, this should be really tough.
If there's a way to determine whether it is possible to allocate by simply using summation of memory usage, this will be great.

aqjune:
Without considering alignment, if there are n memory blocks using k bytes in total, there is at least one empty place such that it has size (2^64 - k) / (n+1). I'll think about this more.

aqjune:
Dependency: #172

aqjune:
Conflict & unit test failures resolved.

ir/memory.cpp
```diff
@@ -438,7 +438,7 @@ Memory::Memory(State &state, bool little_endian)
   // the block) should not overflow.
   avail_space = expr::mkVar("avail_space", bitsPtrSize());
   if (state.isSource()) {
-    state.addAxiom(avail_space.ult(expr::mkInt(-2, bitsPtrSize())));
+    state.addAxiom(avail_space.ult(expr::mkInt(-2 * max_align, bitsPtrSize())));
```
aqjune: The space for a byte for null pointer & a byte for the end of the memory was needed bigger than 2 due to alignment.

tests/alive-tv/memory/alloca-address-nonnull.src.ll
```diff
@@ -1,3 +1,4 @@
+; TEST-ARGS: -smt-to=9000
```
aqjune: The updated tests usually takes less than 9 seconds (except a few), but conservatively set it to this

tests/unit/fshl/folds32.opt
```diff
@@ -1,4 +1,4 @@
-; TEST-ARGS: -root-only
+; TEST-ARGS: -root-only -smt-to:15000
```
aqjune: Right now this is necessary because memory always encodes disjointness between nonlocal blocks even if the src/tgt never accesses memory; this can be removed after the non_local_bid_upperbound is properly set.

smt/exprs.h
```diff
@@ -57,6 +57,8 @@ class FunctionExpr {
   void add(const FunctionExpr &other);
   void del(const expr &key);
   expr operator()(expr &key) const;
+  const decltype(fn)::const_iterator begin() const { return fn.begin(); }
```
nunoplopes: you can use auto as return value

ir/memory.cpp
nunoplopes: sorry, didn't understand.
Why do we need to reserve more than 2 bytes? (I understand the argument for 2).

ir/memory.cpp
```diff
@@ -463,12 +463,28 @@ Memory::Memory(State &state, bool little_endian)
...
+      for (unsigned bid2 = bid + 1; bid2 <= non_local_bid_upperbound; ++bid2) {
+        Pointer p2(*this, bid2, false);
+        state.addAxiom((p1.is_block_alive() && p2.is_block_alive())
+            .implies(disjoint(p1.get_address(), p1.block_size(),
+                              p2.get_address(), p2.block_size())));
+      }
+      // Non-local blocks' addr + size does not overflow.
+      state.addAxiom(p1.get_address().add_no_uoverflow(p1.block_size()));
+      // Size is less than half of memory
+      state.addAxiom(p1.block_size().extract(bits_size_t-1, bits_size_t-1) == 0);
```
nunoplopes: this is not needed. It's already given for free. See the definition of `block_size()`

ir/memory.cpp
```diff
@@ -537,6 +551,16 @@ expr Memory::alloc(const expr &size, unsigned align, BlockKind blockKind,
     auto size_upperbound = size_zext0 + expr::mkUInt(align - 1, bitsPtrSize());
     allocated &= size_upperbound.ule(avail_space);
 
+    // Disjointness of block's address range with other local blocks
+    for (auto &itm : local_blk_addr) {
+      auto zero = expr::mkUInt(0, bitsOffset());
```
nunoplopes: pls hoist this out of loop

nunoplopes:
After this patch we need to reassess performance. It seems it's going down very quickly.
We need 1) the static analysis, and 2) analyse what Z3 is doing and check if further tactics are needed. Also maybe play with the quantifier weights.
Also, If addresses aren't observed we can't skip most address-related constraints.

ir/memory.cpp
```diff
@@ -519,8 +535,6 @@ expr Memory::alloc(const expr &size, unsigned align, BlockKind blockKind,
   Pointer p(*this, bid, is_local);
   auto short_bid = p.get_short_bid();
   // TODO: If address space is not 0, the address can be 0.
-  // TODO: address of new blocks should be disjoint from other live blocks
-  // FIXME: non-zero constraint can go away when disjoint constr is in place
```
nunoplopes: this last comment wasn't addressed.
It's unclear we want to keep the non-null constraint. Maybe yes, but needs to be evaluated.

aqjune: Null pointer block has size 0, so the non-null condition is still needed. It allows two blocks to be adjacent.

ir/memory.cpp
aqjune: If a big block is allocated, the next block cannot be allocated even if avail_space says it's ok.
ex)
Assume that total memory size is 8. 0x0, 0x7 is initially filled. `avail_space` is 6.
Execute `p = alloc 2, align 1`. Assume that p is allocated at `0x4`. Now 0x1, 0x2, 0x3, 0x6 are only available bytes, `avail_space` is 4.
Execute `q = alloc 1, align 4`. `avail_space >= 4`, so it says `q` can be allocated, but due to fragmentation this cannot.

aqjune:
> After this patch we need to reassess performance. It seems it's going down very quickly.
> We need:
> 
> 1. static analysis. This should disable _all_ memory-related constraints for integer-only functions.
> 2. Analyse what Z3 is doing and check if further tactics are needed. Also maybe play with the quantifier weights.
> 3. If addresses aren't observed we can't skip most address-related constraints.
> 4. Try encoding disjointness of globals with quantifiers.

Yep, it is slowing down quite steeply.
If you want I can start these by making PR for item 1 by Monday or Tuesday.

ir/memory.cpp
aqjune: -> Sorry, I don't think this is a good example. I tried to quickly find better one, but couldn't; Decreasing -2 to something lower was needed to make tests pass. I'll leave here after finding better example.

nunoplopes: I see. But this change doesn't fix the problem either.
This avail_space thing needs to be revised. We need to ensure allocation only succeeds if there's address space left. Checking for avail_space is insufficient.

ir/memory.cpp
nunoplopes: adjacent, yes, but not at address zero?

nunoplopes: never mind, you're right.

ir/memory.cpp
aqjune: Declaring `p = alloc 2, align 1` as a non-local var and `q = alloc 1, align 4` as a local one seems to be the case. `p` is not quantified, but `q` is quantified, so that was the problem.
To guarantee whether there exists a space, I think we can use `avail_space / (1 + # of blocks)` instead of `avail_space`. We can say that there exists at least such contiguous space.

aqjune:
Updated the things discussed today.
`./alive-tv/memory/malloc-null.src.ll` is failing, I'll investigate it tomorrow.

PR #172, opened 2019-11-13
Track available memory space size, malloc does not nondeterministically return null anymore

aqjune:
This implements tracking of memory usage.

This is not complete in the sense that it cannot pass unit tests including malloc -> alloca conversion.


aqjune:
Refactoring of calloc + malloc isn't done yet as well

ir/memory.cpp
```diff
@@ -431,6 +431,36 @@ Memory::Memory(State &state, bool little_endian)
   local_block_liveness
     = expr::mkLambda({ expr::mkVar("#bid0", bitsBid() - 1) }, false);
 
+  // The first byte of the memory cannot be allocated because it is for a null
+  // pointer.
+  // The last byte of the memory cannot be allocated because ptr + size
+  // (where size is the size of the block and ptr is the beginning address of
+  // the block) should not overflow.
+  avail_space = expr::mkUInt(-2, bitsPtrSize());
```
nunoplopes: this should be a variable

tests/alive-tv/memory/mallocs-null.src.ll
```diff
@@ -1,92 +0,0 @@
-; target: 64 bits ptr addr
```
aqjune: This test was deleted because it assumed that malloc can nondeterministically return null.

ir/memory.cpp
aqjune: Is it performance reason or correctness reason? I think members like `local_blk_addr` are also expressions rather than Z3 variables.

aqjune: I made it as a variable so initially arbitrary number of bytes are in use

ir/memory.cpp
```diff
@@ -431,6 +431,39 @@ Memory::Memory(State &state, bool little_endian)
...
+
+    // The required space size should consider alignment
+    auto size_upperbound = p.block_size() + max_align_minus1;
+
+    if (state.isSource()) {
+      state.addAxiom(p.is_block_alive().implies(
+          size_upperbound.ule(avail_space)));
+    }
+    avail_space = avail_space - expr::mkIf(p.is_block_alive(),
+                                           p.block_size(),
```
nunoplopes: this should use size_upperbound instead

ir/memory.cpp
```diff
@@ -510,6 +547,9 @@ expr Memory::alloc(const expr &size, unsigned align, BlockKind blockKind,
                    local_block_liveness.store(short_bid, true),
                    local_block_liveness);
 
+    avail_space = avail_space - expr::mkIf(allocated, size_zext0,
```
nunoplopes: same comment as above re pushing subtraction inside If.

nunoplopes: and for size vs size_upperbound.

ir/memory.cpp
```diff
@@ -431,6 +431,39 @@ Memory::Memory(State &state, bool little_endian)
...
+  // Maximum alignment that a non-local block can have
+  // TODO: get this from data layout?
+  auto max_align_minus1 = expr::mkUInt(8 - 1, bitsPtrSize());
+
+  // Omit null pointer
+  for (unsigned bid = 1; bid <= non_local_bid_upperbound; ++bid) {
+    Pointer p(*this, bid, false);
+
+    // The required space size should consider alignment
+    auto size_upperbound = p.block_size() + max_align_minus1;
```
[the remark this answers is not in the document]
aqjune: Maximum alignment was used here because memory does not store alignment per block - do you think we should store them?

ir/memory.cpp
aqjune: I think block_size is okay here, because smaller blocks can fit into the empty space which was forced by alignment.

nunoplopes: If you need to align this allocation, then you effectively waste the alignment space.
Otherwise this encoding may decide there's still space when there's not. For example:
p = alloc 4
q = alloc 4
r = alloc 4

Initial space: 12. Align to 8.
allocations succeed, because there's enough space. But disjointness constraint fails:
we have:
p = 0
q = 8
r = 12 ?? impossible -> UNSAT

ir/memory.cpp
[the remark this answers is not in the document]
nunoplopes: Right, we only have this information for the allocated globals, not for other non-local stuff.
OK, let's keep this maximum alignment; it's not worth it to complicate further the encoding.
But move this to a field in memory.h so it can be filled out by Target data.

ir/memory.cpp
aqjune: The case is addressed by line 536 (`allocated &= size_upperbound.ule(avail_space);`); the third alloca isn’t allocated because size_upperbound is larger than avail_space.

Using size addresses this case:
P = alloc 4, align 4
Q = alloc 2, align 2
Assuming that initial space is 8, it is possible that p = 0x04 & q = 0x06 , but subtracting size_upperbound would block this.

aqjune: It is line 458, not 536 (`state.addAxiom(p.is_block_alive().implies(size_upperbound.ule(avail_space)))`), but rest of the comment is still valid I think.

nunoplopes: If you don't update the available size with the alignment size, you don't get the compounding effect of multiple alignments. The side constraints you have aren't sufficient because they only check that the summation of the sizes + the last alignment fit.

nunoplopes:
Ah, you also need to update function `Memory::operator<` because of the new field

ir/memory.cpp
aqjune: Understood, so to correctly simulate fragmentation, size_upperbound should be added.

aqjune:
> Ah, you also need to update function `Memory::operator<` because of the new field

I think this is okay because the new field is constant (should be fetched from target data), like `bits_size_t`.

tests/alive-tv/memory/alive-tv
```diff
@@ -0,0 +1 @@
+/Users/juneyounglee/alive2/build/alive-tv
```
nunoplopes: oops?

nunoplopes:
> > Ah, you also need to update function `Memory::operator<` because of the new field
> 
> I think this is okay because the new field is constant (should be fetched from target data), like `bits_size_t`.

I meant the field `avail_space`.

aqjune:
Added avail_space to the operator<.
> I meant the field avail_space.

Removed the alive-tv file as well, sorry.

PR #173, opened 2019-11-14
OOB mask of shufflevector result as undef

zhengyangl:


zhengyangl:
Hi Nuno, this patch is not working yet. Now if we feed O.o.B. as index to shufflevector, the result is undef. But if we feed undef as index, the result is somehow still poison.  I don't know if I implemented correctly. Can you please have a look at my code?

tests/unit/vector/shufflevector1.opt
```diff
@@ -48,10 +48,28 @@ Name: constant-folding-8
...
+%r = <3 x float> <float poison, float undef, float undef>
+
+
+Name: constant-folding-12
+%r = shufflevector <2 x float> <float poison, float 5.0>, <2 x float> <float 4.0, float 6.0>, <4 x i32> <i32 5, i32 undef, i32 0, i32 1>
+  =>
+%r = <4 x float> <float undef, float undef, float poison, float 5.0>
+
+
+Name: constant-folding-13
```
zhengyangl: This test case is failed because undef mask shufflevector still results in poison.

nunoplopes: Ah damn.
When the mask is undef, it can take any value. In particular, if it's 0, it picks the poison argument. So we conclude the result can be either poison, undef, or any of the other arguments.
That's the problem with these semantics that "detect" undef and give some value for that case. That usually doesn't work, and here's another example.
Do you want to try to produce an end-to-end miscompilation using this information? (you should be able to use the same gadgets I've used for the other transformation).
You can also run this patch through the LLVM test suite and see what comes out. You may just find the miscompilation there :)

zhengyangl: I need some more time to figure out a miscompilation :)

zhengyangl: @nunoplopes I have tried to figure out a miscompilation. But for now I think it is impossible after Sanjay's patch.

In order to generate a case containing (undef => poison), I tried to trick the optimizer to synthesize a shufflevector instruction from undef, and then expect some  (undef => normal value) optimization to rewrite the undef in the mask to point to an poison value. See https://godbolt.org/z/E8qFwk for an non-working example.

The reason is that for shufflevector, except for the identity simplification optimization discussed in the bugzilla, there is no other transformations happens on the mask of shufflevector. In the other words, after the (undef => shufflevector %v, undef, undef) transformation is done, there is no other writes on the mask, the poison inside %v will never get exposed.

zhengyangl: I've read a few tests in the InstCombine test suite to find out if any other rewrite that turns undef into normal values so that I can use it to composite a miscompile, but I cannot find one. I've also runned the alive-tv on test suite with this patch again, the results are the same.

zhengyangl: Fixing that identity simplification optimization may make the compilation sound for a moment. But in the longer term this undef mask semantic may not be the best choice. As to make the translation sound, most rewrites on undef masks may not fit.

nunoplopes:
I've fixed the semantics of shufflevector here: https://github.com/AliveToolkit/alive2/commit/fdccba74a3bc358180e32e708dbfc012ffdea201

Issue #176, opened 2019-11-16
GEP constant expression


aqjune:
Among LLVM unit test results, ~1900 unsupported cases are about loads having GEP constexpr as its operand:

```
Unsupported IR features (Top 20):                                                                                                                                                                                                             
3422  phi i32                                                                                                                                                                                                                                 
2843  phi i64                                                                                                                                                                                                                                 
695 getelementptr inbounds                                                                                                                                                                                                                    
654 fpext float                                                                                                                                                                                                                               
401 load i32, i32* getelementptr inbounds <<                                                                                                                                                                                                    
383 getelementptr                                                                                                                                                                                                                             
356 load i64, i64* getelementptr inbounds <<                                                                                                                                                                                                    
296 load i8, i8* getelementptr inbounds <<                                                                                                                                                                                                      
295 load i16, i16* getelementptr inbounds <<                                                                                                                                                                                                    
285 load float, float* getelementptr inbounds <<                                                                                                                                                                                                
259 load double, double* getelementptr inbounds <<                                                                                                                                                                                              
198 tail call i8* @llvm.objc.retain                                                                                                                                                                                                           
189 call i32                                                                                                                                                                                                                                  
146 call i1 @llvm.type.test                                                                                                                                                                                                                   
127 call i8* @llvm.objc.retain                                                                                                                                                                                                                
119 phi i8*                                                                                                                                                                                                                                   
112 call <2 x float> @llvm.pow.v2f32                                                                                                                                                                                                          
107 call token                                                                                                                                                                                                                                
100 call <2 x double> @llvm.pow.v2f64                                                                                                                                                                                                         
100 phi i16                                  
```

This can be addressed by breaking `load gep (..)` into two instructions when lowering to Alive2:
```
%tmp = gep ..
load %tmp
```

Having constant expression in global variable's initialization is a different issue. I guess this should require a bit of work.

I'll make a short PR that resolves this `load gep` problem first.

nunoplopes:
I don't know if it's worth it to specialize for constant exprs in functions, since then we will have to duplicate work for globals. Let's just do it right once.
We should consider if it's possible to reuse current code of instructions for constants. Hopefully it is :)

aqjune:
Good news is that ConstantExpr has `getAsInstruction()` which creates an instruction that is equivalent to the constexpr. Using this, we can reuse existing `llvm2alive_`.
[...]

PR #179, opened 2019-11-18
Let llvm2alive_::run convert constexprs to instructions

aqjune:
This PR converts constexprs to instructions (#176).

constexprs inside global vars are converted into corresponding instructions in Initializer block.

constexprs inside phis are placed just before phis.


aqjune:
Closing this PR to split this into two smaller ones

PR #180, opened 2019-11-20
Fix scripts

aqjune:
A few lines of patch that fixes scripts.

### DIFF
--- a/scripts/stats.php
+++ b/scripts/stats.php
@@ -11,7 +11,7 @@ $stats       = array();
 $knownfns    = array();
 $unsupported = array();
 
-foreach (glob("logs/*.txt") as $f) {
+foreach (glob("$logs/*.txt") as $f) {
   $txt = file_get_contents($f);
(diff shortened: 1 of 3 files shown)


tests/lit/lit/formats/alive2test.py
```diff
@@ -104,8 +104,8 @@ def execute(self, test, litConfig):
       return lit.Test.FAIL, out + err
 
     if exitCode != 0:
-      if expect_err != None and string.find(err, expect_err.group(1)) != -1:
+      if expect_err != None and err.find(expect_err.group(1)) != -1:
```
aqjune: This changes help `ninja check` work with both Python2 and Python3

scripts/stats.php
```diff
@@ -11,7 +11,7 @@
 $knownfns    = array();
 $unsupported = array();
 
-foreach (glob("logs/*.txt") as $f) {
+foreach (glob("$logs/*.txt") as $f) {
```
aqjune: Previous `logs` variable was not used in this file.

PR #182, opened 2019-11-22
Add constants for the number of local/non-local blocks

aqjune:
This adds constants for the number of local/non-local blocks and a simple static analyzer that counts them.
Using this, bits_for_bid is determined as well. With this PR, now ninja check passes again.

An instruction that returns a pointer is added a constraint that checks whether its non-local block id is less than num_nonlocals.

### DIFF
--- a/ir/instr.h
+++ b/ir/instr.h
@@ -113,6 +113,7 @@ public:
   ConversionOp(Type &type, std::string &&name, Value &val, Op op)
     : Instr(type, std::move(name)), val(&val), op(op) {}
 
+  Op getOp() const { return op; }
   std::vector<Value*> operands() const override;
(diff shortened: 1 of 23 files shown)


aqjune:
Updated the PR so
- changes relevant to determining bits_for_bid only remain
- bits_for_offset is moved out as well

What are not addressed:
- Constants are still at `state.h/cpp`; I thought about `config.h` , but configuration means state in language semantics.
- Removal of null-pointer when there's no pointer operation at all is not addressed. This might cause another widespread changes (because it is related with all +1 in block number count) so I like to address that in a separate PR.

There is almost no change in the number of failures (12 -> 11). If my observation is correct, encoding disjointness of nonlocal blocks is one of the main source of timeout (another one is memory refinement).

PR #186, opened 2019-11-24
Let load/stores encode aggregates

aqjune:
This helps correctly loading aggregate's elements.
This resolves 11 failures in LLVM unit tests (243 -> 232).

Previously, first element was at the uppermost bits of aggregate representation, causing load to access the last element rather than the first element.
This updates representation of aggregate value, so the first element comes at the lower-most bits.

This PR works only for little endian; for big endian, bigger change is needed at Memory::store. 

It resolves 8 `Target is more poisonous than source` and 4 `Mismatch in memory` errors, but has a few number of increases in invalid expr / program type check / oom. 147 timeout cases were resolved too, which is interesting.
```
Errors:                                            
3240  Timeout                                      
1242  Invalid expr                                 
632 Loops are not supported yet! Skipping function.
343 program doesn't type check!                    
64  Source is more defined than target             
50  Value mismatch                                 
33  Mismatch in memory                             
29  Target is more poisonous than source           
5 Out of memory; skipping function.                

=>

Errors:                                            
3093  Timeout                                      
1252  Invalid expr                                 
632 Loops are not supported yet! Skipping function.
347 program doesn't type check!                    
64  Source is more defined than target             
47  Value mismatch                                 
29  Mismatch in memory                             
21  Target is more poisonous than source           
8 Out of memory; skipping function.                
```


tests/alive-tv/memory/aggregate-bigendian.src.ll
```diff
@@ -0,0 +1,12 @@
+target datalayout="E-p:64:64:64-p1:16:16:16-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f32:32:32-f64:64:64-v64:64:64-v128:128:128-a0:0:64-s0:64:64"
+
+define i32 @test1({{i32,i8},i32}* %x) {
+  ; {{ 0xDEADBEEF, 0xBA }, 0xCAFEBABE}
+  store {{i32,i8},i32} {{i32,i8} { i32 -559038737, i8 186 }, i32 -889275714 }, {{i32,i8},i32}* %x
+  %y = getelementptr {{i32,i8},i32}, {{i32,i8},i32}* %x, i32 0, i32 0, i32 0
+  %r = load i32, i32* %y
+  ret i32 %r
+}
```
aqjune: This test was exerpted from `test/Transforms/ConstProp/loads.ll` (with minor change to soothe Mismatch in memory).

ir/type.cpp
```diff
@@ -608,7 +608,7 @@ StateValue AggregateType::aggregateVals(const vector<StateValue> &vals) const {
   StateValue v;
   for (unsigned idx = 0; idx < elements; ++idx) {
     auto vv = children[idx]->toBV(vals[idx]);
-    v = idx == 0 ? move(vv) : v.concat(vv);
```
nunoplopes: what do you think about making the representation dependent on the target's endianness. So, keep the old code for big endian, and add your code for little endian.
With this, we avoid byte swaps when loading/storing.

aqjune: The current format is wrong in the big endian too. In big endian, first element should come at the lower address as well. The `test/Transforms/ConstProp/loads.ll` test also guarantees this.

aqjune:
Will work tmr or Friday.
I'd like to leave store/load of vector of non-bytewise element as unsupported. The reason is that it seems unclear to me whether they should be packed, and also what their endian should be.
```
%a = <i12 0xABC, i12 0xDEF>
store %a, ptr // What should *ptr be?
```
What do you think?

nunoplopes:
> Will work tmr or Friday.
> I'd like to leave store/load of vector of non-bytewise element as unsupported. The reason is that it seems unclear to me whether they should be packed, and also what their endian should be.

Sounds good to me.  We need to make sure padding works well before going into non-byte-aligned elements.

aqjune:
I'll be able to update this PR soon

aqjune:
Updated the PR
- AggregateType tracks (byte offset, bytewidth, is_padding) list. For example, {i8, i32} is translated to [(0, 1, false), (1, 3, true), (4, 4, false)].
- Memory::load / Memory::store on aggregate type is updated.
- Couldn't put undef value to padding, because it caused timeout on the added unit tests, which is simple. I put zero here, but I suspect it isn't fully correct. What should we do here?

ir/type.cpp
```diff
@@ -771,6 +782,21 @@ VectorType::VectorType(string &&name, unsigned elements, Type &elementTy)
...
+    elemsz = bits_size_t;
+  else if (elementTy.isIntType() || elementTy.isFloatType())
+    elemsz = elementTy.bits();
+  else {
+    assert("Only pointer / integer / float types are allowed for vector elements"
+           && false);
+  }
+  for (unsigned i = 0; i < elements; ++i)
+    // TODO: elemsz with non-multiple of 8 is not supported.
+    layout[i] = { i * (elemsz / 8), elemsz / 8, false };
```
aqjune: Observed that vector types didn't seem to have paddings: see https://godbolt.org/z/tuPT8R

ir/memory.cpp
```diff
@@ -706,33 +706,62 @@ void Memory::free(const expr &ptr) {
...
-  ptr.is_dereferenceable(bytes.size(), align, !state->isInitializationPhase());
-  for (unsigned i = 0, e = bytes.size(); i < e; ++i) {
-    auto ptr_i = little_endian ? (ptr + i) : (ptr + (e - i - 1));
-    store(ptr_i, bytes[i](), local_block_val, non_local_block_val);
+  bool do_write = !state->isInitializationPhase();
+
+  if (type.isAggregateType()) {
+    auto aty = dynamic_cast<const AggregateType *>(&type);
+    if (deref_check)
+      ptr.is_dereferenceable(aty->bytesize(), align, do_write);
```
aqjune: Check dereferenceable at top level only

aqjune:
LLVM unit tests failure #: 164 -> 150
Two failures appear:
Transforms/SCCP/calltest.ll - this is relevant to assertion failure in known_fn, I'll have a look and make a patch
Transforms/InstCombine/2012-6-7-vselect-bitcast.ll - this seems to be related with bitcast between vectors

ir/type.cpp
```diff
@@ -614,11 +614,22 @@ AggregateType::AggregateType(string &&name, bool symbolic)
   }
 }
 
+AggregateType::AggregateType(string &&name, vector<Type*> &&vchildren,
```
nunoplopes: this design seems slightly complicated.
What do you think about explicitly adding padding fields to structs, so the API wouldn't change much. Each field would have a bool stating whether it's padding or not. Then we just need to detect padding when doing GEPs into structs, but the rest of the code (loads, stores, etc) wouldn't need any change. A padding field is a field like any other. Just with poison data I guess.

aqjune: The `vlayout` parameter exactly does what you've suggested. It takes (beginning offset, field size, padding). Detection of padding is done at utils.cpp . I think load/store still need change because they need to access elementwise (and respect endian of each element while preserving element order).

aqjune:
Removes padding related code; will make a separate patch for that

aqjune:
I'll make array type support PR after this patch is reviewed & merged because it depends on this PR (I think it isn't good to have multiple dependent PRs)

aqjune:
Conflict resolved

PR #187, opened 2019-11-24
Support constexprs in initializers

aqjune:
This PR creates a standalone basic block which is called "initializer block", and let llvm2alive.cpp add constexprs of initializers to it.

This patch helps LLVM's `Transforms/GlobalOpt/array-elem-refs.ll` pass.

The output of `alive-tv constgep.src.ll constgep.tgt.ll` looks like this:

```
{
__globalvars_init:
  %__constexpr_0 = bitcast * @x to *
  store * %__constexpr_0, * @y, align 1
}
define * @f() {
%0:
  %p = load *, * @y, align 8
  ret * %p
}
=>
{
__globalvars_init:
  %__constexpr_1 = bitcast * @x to *
  store * %__constexpr_1, * @y, align 1
}
define * @f() {
%0:
  %p = load *, * @y, align 8
  %a = bitcast * @x to *
  %b = gep inbounds * %a, 1 x i64 0
  ret * %b
}

```

### DIFF
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -73,10 +73,14 @@ void Function::fixupTypes(const Model &m) {
   }
 }
 
-BasicBlock& Function::getBB(string_view name) {
+BasicBlock& Function::getBB(string_view name, bool push_front) {
   auto p = BBs.try_emplace(string(name), name);
-  if (p.second)
-    BB_order.push_back(&p.first->second);
+  if (p.second) {
+    if (push_front)
+      BB_order.insert(BB_order.begin(), &p.first->second);
+    else
+      BB_order.push_back(&p.first->second);
+  }
   return p.first->second;
 }
 
@@ -96,6 +100,14 @@ vector<GlobalVariable *> Function::getGlobalVars() const {
       gvs.push_back(gv);
   }
   return gvs;
+}
+
+vector<string_view> Function::getGlobalVarNames() const {
+  vector<string_view> gvnames;
+  auto gvs = getGlobalVars();
+  transform(gvs.begin(), gvs.end(), back_inserter(gvnames),
+            [](auto itm) { return string_view(itm->getName()).substr(1); });
+  return gvnames;
 }
 
 void Function::addPredicate(unique_ptr<Predicate> &&p) {
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -711,7 +711,7 @@ void Memory::store(const expr &p, const StateValue &v, const Type &type,
   vector<Byte> bytes = valueToBytes(v, type, *this);
 
   Pointer ptr(*this, p);
-  ptr.is_dereferenceable(bytes.size(), align, true);
+  ptr.is_dereferenceable(bytes.size(), align, !state->isInitializationPhase());
   for (unsigned i = 0, e = bytes.size(); i < e; ++i) {
     auto ptr_i = little_endian ? (ptr + i) : (ptr + (e - i - 1));
     store(ptr_i, bytes[i](), local_block_val, non_local_block_val);
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -5,6 +5,7 @@
 #include "llvm_util/known_fns.h"
 #include "llvm_util/utils.h"
 #include "llvm/IR/GetElementPtrTypeIterator.h"
+#include "llvm/IR/GlobalVariable.h"
 #include "llvm/IR/InstVisitor.h"
 #include "llvm/IR/Operator.h"
 #include <utility>
@@ -59,6 +60,7 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
   llvm::Function &f;
   const llvm::TargetLibraryInfo &TLI;
   vector<llvm::Instruction *> i_constexprs;
+  vector<string_view> gvnamesInSrc;
 
   using RetTy = unique_ptr<Instr>;
 
@@ -97,8 +99,9 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
   }
 
 public:
-  llvm2alive_(llvm::Function &f, const llvm::TargetLibraryInfo &TLI)
-      : f(f), TLI(TLI) {}
+  llvm2alive_(llvm::Function &f, const llvm::TargetLibraryInfo &TLI,
+              vector<string_view> &&gvnamesInSrc)
+      : f(f), TLI(TLI), gvnamesInSrc(move(gvnamesInSrc)) {}
 
   ~llvm2alive_() {
     for (auto &inst : i_constexprs) {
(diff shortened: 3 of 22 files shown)


util/symexec.cpp
```diff
@@ -13,17 +13,32 @@ using namespace std;
...
+  // add constants & undefs to State table first
+  for (auto &l : { f.getConstants(), f.getUndefs() }) {
     for (const auto &v : l) {
       s.exec(v);
     }
   }
 
   s.exec(Value::voidVal);
 
+  for (auto &i : initBB.instrs()) {
```
nunoplopes: you could hoist this function into a lambda instead of duplicating the code.

tools/transform.cpp
```diff
@@ -559,12 +559,22 @@ void Transform::print(ostream &os, const TransformPrintOpts &opt) const {
   os << "\n----------------------------------------\n";
   if (!name.empty())
     os << "Name: " << name << '\n';
+  if (!src_init.empty()) {
+    os << "{\n";
```
nunoplopes: you can do consecutive << :)

util/symexec.cpp
```diff
@@ -13,17 +13,32 @@ using namespace std;
 namespace util {
 
 void sym_exec(State &s) {
+  BasicBlock &initBB = const_cast<BasicBlock&>(s.getInitializerBB());
```
nunoplopes: maybe change getInitializerBB so it doesn't return a const

llvm_util/utils.cpp
```diff
@@ -41,10 +41,34 @@ ostream *out;
...
 
+vector<llvm::GlobalVariable *> getUsedGlobalVariables(llvm::Function &F) {
+  llvm::SmallPtrSet<llvm::Value *, 16> Visited;
+  llvm::SmallPtrSet<llvm::GlobalVariable *, 16> GlobalVars;
+  for (auto &BB : F) {
+    for (auto &I : BB) {
+      ::_getUsedGlobalVariables(&I, Visited, GlobalVars);
+    }
+  }
+  return vector<llvm::GlobalVariable *>(GlobalVars.begin(), GlobalVars.end());
```
nunoplopes: you can probably use {begin(), end()}

llvm_util/utils.cpp
```diff
@@ -41,10 +41,34 @@ ostream *out;
...
+  }
+}
+
 }
 
 namespace llvm_util {
 
+vector<llvm::GlobalVariable *> getUsedGlobalVariables(llvm::Function &F) {
+  llvm::SmallPtrSet<llvm::Value *, 16> Visited;
+  llvm::SmallPtrSet<llvm::GlobalVariable *, 16> GlobalVars;
```
nunoplopes: why are you using LLVM datastructures? We don't want to rely on LLVM stuff whenever it's not strictly needed

llvm_util/utils.cpp
```diff
@@ -41,10 +41,34 @@ ostream *out;
...
+      GlobalVars.insert(GV);
+    }
+  }
+}
+
 }
 
 namespace llvm_util {
 
+vector<llvm::GlobalVariable *> getUsedGlobalVariables(llvm::Function &F) {
```
nunoplopes: uhm, why is this analysis implicit when generating alive code? We already traverse the whole code once, why this additional traversal?

llvm_util/llvm2alive.cpp
```diff
@@ -648,6 +649,24 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
       Fn.addInput(move(val));
     }
 
+    // If global variables contain initializer, put them at init block.
+    string initbb_name = ("__globalvars_init");
+    BasicBlock InitBB(move(initbb_name));
```
nunoplopes: inline the move pls.

util/symexec.cpp
```diff
@@ -13,17 +13,32 @@ using namespace std;
...
+
+    if (config::symexec_print_each_value && name[0] == '%')
+      cout << name << " = " << val << '\n';
+  }
+
+  s.finishInitializer();
+
+  // Add inputs
+  for (const auto &v : f.getInputs())
+    s.exec(v);
```
nunoplopes: why is this done after running the init program?
So alloc will run after store? This is not particularly good and it may prevent further optimizations (like propagating sizes of non-locals).

llvm_util/utils.cpp
aqjune: updated the code to use std::set.

llvm_util/utils.cpp
aqjune: The reason was that making get_operand do these thing seemed to make get_operand too complex.
When constexprs are visited at initializers, corresponding instructions are generated at the separate basic block unlike constexprs in a function, and creation of the store to the global variable is needed.
Instead of making get_operand more complex (which was already carrying constexpr_conv around), I put the initializer generation separately before beginning of function translation, which required this analysis.

util/symexec.cpp
aqjune: GlobalVariables are added as constants to the function (through `current_fn->addConstant()`), so things are okay here.

llvm_util/utils.cpp
aqjune: Or alternatively, we can move get_operand to llvm2alive.cpp.
This way, get_operand wouldn't need to carry auxiliaries needed for constexprs.

aqjune:
Alphabetically sorting global variable is not done yet - the main reason is that initializer may reference another global variable, and knowing the full list of used global variables is known only after traversing all global vars.
```
@c = global i8* ... @b ...
@b = global i8* ... @a ...
@a = global i8 0

define void @f() {
  load @a
}
```
Maybe reordering instructions in the initializer block would work?
I'll continue working tomorrow.

ir/function.cpp
```diff
@@ -98,6 +98,14 @@ vector<GlobalVariable *> Function::getGlobalVars() const {
   return gvs;
 }
 
+vector<string> Function::getGlobalVarNames() const {
+  vector<string> gvnames;
+  auto gvs = getGlobalVars();
+  transform(gvs.begin(), gvs.end(), back_inserter(gvnames),
+            [](auto itm) { return string(itm->getName()).substr(1); });
```
nunoplopes: you can use string_view here to avoid copy.

ir/memory.cpp
```diff
@@ -308,8 +308,10 @@ void Pointer::is_dereferenceable(const expr &bytes0, unsigned align,
 
   cond &= is_block_alive();
 
-  if (iswrite)
+  if (iswrite && !m.state->isInitializationPhase()) {
+    // In initialization phase, read-only blocks' values are being updated.
```
nunoplopes: is this the right place to put this?
Why don't we just disable all checks in store, since we know they all pass anyway?

nunoplopes:
> Alphabetically sorting global variable is not done yet - the main reason is that initializer may reference another global variable, and knowing the full list of used global variables is known only after traversing all global vars.
> 
> ```
> @c = global i8* ... @b ...
> @b = global i8* ... @a ...
> @a = global i8 0
> 
> define void @f() {
>   load @c
> }
> ```
> 
> Maybe reordering instructions in the initializer block would work?
> I'll continue working tomorrow.

Nice catch.
Maybe we need to do 2 passes over the globals: one to generate the initialization code, and another to do the stores in right order. Otherwise I'm not seeing how to getting a deterministic order (it doesn't have to be alphabetic; just something deterministic in both programs).

nunoplopes:
I was thinking last night that maybe creating separated BBs for const init isn't worth the pain. It seems we don't get anything out of it.
What do you think about moving this BB to the beginning of the function?  We can still special case its execution in symexec so it doesn't produce UB when writing to const vars.

Also, your previous code to traverse globals in LLVM world might be easier to get a sorted list of used globals. I dunno, pls have a thought :)

ir/memory.cpp
aqjune: Updated so store calls Pointer::is_dereferenceable with appropriate iswrite

aqjune:
- Insert the initializer block at the beginning of a function
- Emit stores to global vars in sorted order

Unit tests error: 206 -> 177
Found out that there are cases where global variables may not have a name. For example:
Transforms/GlobalOpt/2009-03-06-Anonymous.ll
In this case, things doesn't seem to work well. Assigning names to them before running -tv would be the simplest way (`-instnamer` pass), but it would change the bitcode.
I'll make a separate patch that resolves this in the future, as this seems not very urgent issue.

PR #188, opened 2019-11-25
Fix a calloc bug

aqjune:
Let calloc successfully initialize.

`Pointer::Pointer(const Memory &m, const char *var_name)` was creating a non-local pointer, and it was incorrectly used by Memory::memset.
Similar thing was happening in memcpy as well.


aqjune:
The bug found by @Dongjoo-Kim

PR #190, opened 2019-11-26
nullptr_is_used should consider undef pointer too

aqjune:
This PR addresses about 20 "Assertion `!memory_unused()' failed." error from LLVM unit tests.

nullptr_is_used should consider undef pointer because undef pointer is pointing to a null block.


aqjune:
LLVM unit tests failure: 242 -> 226

PR #197, opened 2019-12-01
Optimize encoding of bytes by widening it if possible

aqjune:
A good place to start is transform.cpp, which calculates the new constant bits_byte.
This is written in conservative manner in the sense that if GEP or non-local pointers are met, bails out

LLVM unit tests is running


nunoplopes:
Didn't check the patch carefully, but just a quick comment: you need to check alignment in operations as well, otherwise a 2-byte access may be unaligned and hit 2 memory locations. I would conservatively downgrade those cases to 8-bits byte.

aqjune:
No diff in LLVM unit test failure
Timeout: almost no diff (2141 -> 2143); maybe larger tests are needed to see its effect?

util/compiler.cpp
```diff
@@ -61,4 +61,12 @@ uint64_t divide_up(uint64_t n, uint64_t amount) {
   return (n + amount - 1) / amount;
 }
 
+uint64_t gcd(uint64_t n, uint64_t m) {
```
nunoplopes: uhm, there are more standard algorithms for this, such as: https://en.wikipedia.org/wiki/Greatest_common_divisor#Euclid's_algorithm

tests/alive-tv/undef.ll
```diff
@@ -0,0 +1,7 @@
+define i8* @f(i8 %a) {
```
nunoplopes: it seems this file committed by accident

ir/value.cpp
```diff
@@ -26,6 +27,15 @@ void Value::fixupTypes(const Model &m) {
   type.fixup(m);
 }
 
+bool Value::isInt(int64_t &i) const {
```
nunoplopes: please sort this in the same order as in the header (ie move this up)

tools/transform.cpp
```diff
@@ -379,6 +383,46 @@ static void calculateAndInitConstants(Transform &t) {
...
+        return gcd(align, util::divide_up(ity->bits(), 8));
+      else if (auto fty = dynamic_cast<const FloatType *>(value_ty))
+        return gcd(align, fty->bits() / 8);
+      else if (dynamic_cast<const PtrType *>(value_ty))
+        return gcd(align, bits_size_t / 8);
+      // Aggregate types should consider padding
+      return UNKNOWN;
+    }
+
+    if (dynamic_cast<const BinOp *>(&inst) ||
```
nunoplopes: I think it would be clearer if all the no access instructions would be moved up.

tools/transform.cpp
```diff
@@ -379,6 +383,46 @@ static void calculateAndInitConstants(Transform &t) {
...
+    if (auto st = dynamic_cast<const Store *>(&inst)) {
+      value_ty = &st->getValue()->getType();
+      align = st->getAlign();
+    } else if (auto ld = dynamic_cast<const Load *>(&inst)) {
+      value_ty = &ld->getType();
+      align = ld->getAlign();
+    }
+
+    if (value_ty) {
+      if (auto ity = dynamic_cast<const IntType *>(value_ty))
```
nunoplopes: value_ty->bits() should work for all types, incl pointers

tools/transform.cpp
```diff
@@ -419,6 +465,13 @@ static void calculateAndInitConstants(Transform &t) {
           has_int2ptr |= conv->getOp() == ConversionOp::Int2Ptr;
           has_ptr2int |= conv->getOp() == ConversionOp::Ptr2Int;
         }
+
+        auto accsz = get_access_size(I);
+        if (accsz != NO_ACCESS){
+          min_access_size = accsz == UNKNOWN ? 1 :
+              (min_access_size == UNKNOWN ? accsz :
+                                            gcd(min_access_size, accsz));
```
nunoplopes: to simplify this, you could set UNKNOWN to 1, since that's the alignment when unknown, and set min_access_size to some max alignment, say 512. GCD will take care of bringing that down (it's all powers of 2, so it's fine).

tools/transform.cpp
```diff
@@ -432,7 +485,23 @@ static void calculateAndInitConstants(Transform &t) {
 
   // TODO
   bits_for_offset = 64;
-  bits_size_t = 64;
+
+  // size of byte
+  if (num_nonlocals != 1 || t.src.getType().isPtrType())
+    // Be conservative, run when there are unescaped local blocks only.
```
nunoplopes: what's the problem with functions returning pointers or with global variables or with pointer inputs?
If all memory accesses are aligned, it would be UB if the program receives an unaligned ptr that is dereferenced. So I don't see why this restriction is needed?

ir/memory.cpp
```diff
@@ -16,14 +16,14 @@ namespace {
 // A data structure that represents a byte.
 // A byte is either a pointer byte or a non-pointer byte.
 // Pointer byte's representation:
```
nunoplopes: uhm, comment with seeing the rest of the file: if we expand to multi-byte, then multiple offset bytes may appear in a memory cell.
I don't think it's possible to mix multiple bids because otherwise a single load would access parts of multiple pointers.
We need, however, ranges of byte offsets when going multi-byte?

aqjune: The range is set to be [offset * byte_size, offset * (byte_size+1) - 1) in this PR.
ex) the offset 0 means bytes 0 ~ 3'th byte of the pointer.

tools/transform.cpp
aqjune: Pointer's bits() returns bits_bid + bits_offset (which is 80), so does not work here.

tools/transform.cpp
aqjune: Yes, you're right. This part had to be deleted.

util/compiler.cpp
aqjune: This implements the algorithm :) with additional process when n < m & slightly optimize when n % m is already zero.

PR #202, opened 2019-12-04
Support padding in struct type

aqjune:
This PR depends on #186. The last commit (dc1e58e) is the update.
`is_padding` is defined as a separate member.
Padding is set as poison here.


aqjune:
Conflict resolved

aqjune:
I'll start working on array support.

PR #203, opened 2019-12-05
Initialize global variables too

pranavk:
I may have missed something but is there any reason we are not initializing the global variables now since we have the support for initializing them?


aqjune:
Hi,
The reason is that global variables' value might have already been changed by a caller.
In C, it is something like this:

```
int my_global = 0;

void f() {
  assert(my_global == 0); // This assertion fails if f was called by caller()
}

void caller() {
  my_global = 1;
  f();
}
```

pranavk:
Ah, true, silly me. Closing this.

PR #204, opened 2019-12-06
Support clang plugin

aqjune:
This is a patch that let alive2 run as a clang plugin.
It uses the new pass manager to get callback after each pass. It seems the new pass manager is working pretty well.
To make this PR work, https://reviews.llvm.org/D71086 needs to be reviewed. It is adding just a couple of lines to LLVM's header.

```
clang -fpass-plugin=build/tv/tv.dylib -fexperimental-new-pass-manager -O3 a.c -S -emit-llvm \
  -Xclang -load -Xclang tv/tv.dylib
```
You can give additional arguments with -mllvm (e.g. -mllvm -tv-smt-to=2000)

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -6,10 +6,15 @@
 #include "smt/solver.h"
 #include "tools/transform.h"
 #include "util/config.h"
+#include "llvm/ADT/Any.h"
 #include "llvm/Analysis/TargetLibraryInfo.h"
 #include "llvm/IR/Module.h"
+#include "llvm/IR/LegacyPassManager.h"
 #include "llvm/Pass.h"
+#include "llvm/Passes/PassBuilder.h"
+#include "llvm/Passes/PassPlugin.h"
 #include "llvm/Support/CommandLine.h"
+#include "llvm/Support/raw_ostream.h"
 #include <fstream>
 #include <iostream>
 #include <random>
@@ -98,6 +103,7 @@ unsigned initialized = 0;
 bool showed_stats = false;
 bool report_dir_created = false;
 bool has_failure = false;
+bool is_clangtv = false;
 
 
 struct TVPass final : public llvm::FunctionPass {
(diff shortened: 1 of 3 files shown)


aqjune:
There was a change that was not applied so I push forced, sorry

tv/tv.cpp
```diff
@@ -89,6 +94,7 @@ optional<smt::smt_initializer> smt_init;
 optional<llvm_util::initializer> llvm_util_init;
 TransformPrintOpts print_opts;
 unordered_map<string, pair<Function, unsigned>> fns;
+static llvm::TargetLibraryInfo *TLI;
```
nunoplopes: static isn't needed here; already under anonymous namespace

aqjune:
Rebased (no change other than the static keyword)

aqjune:
I think it will be great if this patch is merged into master, so we can run single file benchmarks after each update.
I added a cmake flag `-DCLANG_PLUGIN=1` which enables compilation of the callback registration. If not set, the code won't compile.

tv/tv.cpp
```diff
@@ -239,4 +262,110 @@ struct TVPass final : public llvm::FunctionPass {
...
+  llvm_unreachable("Unknown IR unit");
+}
+
+static bool do_skip(const llvm::StringRef &ref) {
+  const vector<string_view> pass_list = {
+    "::TVInitPass", "::TVFinalizePass",
+    "ArgumentPromotionPass", "DeadArgumentEliminationPass",
+    "HotColdSplittingPass", "InlinerPass",
+    "GlobalOptPass", "IPSCCPPass"
+  };
```
aqjune: I kept this as a vector to use this as `std::find_if(pass_list.begin(), pass_list.end(), ..)` below.

aqjune:
(this patch is ready to get reviewed. To see how clang emits an error when TV fails, try this example)
```
int f(int *a, int *b) {
  int *tmp = *(int**)a;
  *(int**)b = tmp;
  return 0;
}
```

PR #210, opened 2019-12-07
Add more passes to ignore to opt-alive.sh, fix array printing bug

aqjune:


nunoplopes:
for the passes, can you explain why these should be blacklisted?
 - switch-to-lookup
 - safe-stack
 - pgo-instr-gen
 - loop-extract

Also, are empty arrays allowed in LLVM?

aqjune:
They either create new global variables or new functions.
- switch-to-lookup creates a global constant and uses that if possible (Transforms/SimplifyCFG/X86/switch_to_lookup_table.ll ).
- safe-stack creates a global variable that simulates a distinct stack area  (something like `%unsafe_stack_ptr = load *, * @__safestack_unsafe_stack_ptr, align 8`)
- pgo-instr-gen creates new global variables that store profiling info (something like `@__profc_test_macho`)
- loop-extract extracts a top-level loop into a distinct function.

Empty arrays are allowed in LLVM, according to LangRef (https://llvm.org/docs/LangRef.html#array-type ).

aqjune:
Here are LLVM's tests:
- switch-to-lookup: Transforms/SimplifyCFG/X86/switch_to_lookup_table.ll
- safe-stack: Transforms/SafeStack/X86/phi.ll
- pgo-instr-gen: Transforms/PGOProfile/X86/macho.ll
- loop-extract: Transforms/CodeExtractor/2004-03-13-LoopExtractorCrash.ll

PR #215, opened 2019-12-08
Fix assertion failure at expr::load when optimizing lambda

aqjune:
At some point, the second argument of app can also have form `((_ extract 66 0) (concat #x2 ((_ extract 63 0) (:var 0)))))`.
Test case: Transforms/GVN/pr17732.ll


nunoplopes:
Fixed, thanks!

Issue #209, opened 2019-12-07
Treating a global variable with numeric assigned as its name


aqjune:
Transforms/SimplifyCFG/branch-fold-dbg.ll has a global variable with numeric name assigned, and llvm::GlobalVariable::getName() returns a nonsense value.
This causes llvm::Module::getGlobalVariable() at llvm2alive.cpp fail and return null.

```
@0 = external hidden constant [5 x ...], align 4
...
```

Possible workaround is to not use the Module::getGlobalVariable() & make Alive2's GlobalVariable maintain which LLVM's global variable it was created from. Would this solution work?

nunoplopes:
The problem with unnamed constants is that they may change their number, since the number is just the order in which they appear in the file.
We can try to ignore this and see how it goes. I guess we can just iterate over all globals and count the unnamed ones every time a new GlobalVariable is created.

aqjune:
I see, I'll have a look.

PR #216, opened 2019-12-08
Support anonymous global variables

aqjune:
This resolves issue #209 . It works pretty well.

LLVM unit tests failure: 136 -> 129

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -704,28 +704,22 @@ public:
     auto getGlobalVariable = [this](const string &name) -> llvm::GlobalVariable * {
       auto M = f.getParent();
       // If name is a numeric value, the result should be manually found
-      bool is_numeric = name.size() == 1 && name[0] == '0';
-      unsigned numeric_id = 0;
-      if (!is_numeric) {
-        is_numeric = true;
-        bool first = true;
-        for (char c : name) {
-          if (!(('0' + first) <= c && c <= '9')) {
-            is_numeric = false;
-            break;
-          }
-          first = false;
-          numeric_id = numeric_id * 10 + (c - '0');
-        }
-      }
-      if (!is_numeric)
+      const char *chrs = name.data();
+      char *end_ptr;
+      auto numeric_id = strtoul(chrs, &end_ptr, 10);
+
+      if (end_ptr - chrs != name.size())
         return M->getGlobalVariable(name, true);
       else {
-        auto itr = M->global_begin();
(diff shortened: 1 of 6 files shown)


llvm_util/utils.cpp
```diff
@@ -248,7 +250,23 @@ Value* get_operand(llvm::Value *v,
...
-    string name = '@' + gv->getName().str();
+    string name;
+    if (!gv->hasName()) {
+      unsigned id = 0;
+      auto M = gv->getParent();
+      auto i = M->global_begin(), e = M->global_end();
+      for (; i != e; ++i) {
+        if (&(*i) == gv)
+          break;
+        id++;
```
nunoplopes: ++id

llvm_util/utils.cpp
```diff
@@ -248,7 +250,23 @@ Value* get_operand(llvm::Value *v,
...
+      }
+      assert(i != e);
+      stringstream ss;
```
nunoplopes: use of stringstream here isn't really needed

llvm_util/llvm2alive.cpp
```diff
@@ -694,13 +694,40 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
       }
     }
 
+    auto getGlobalVariable = [this](const string &name) -> llvm::GlobalVariable * {
+      auto M = f.getParent();
+      // If name is a numeric value, the result should be manually found
+      bool is_numeric = name.size() == 1 && name[0] == '0';
+      unsigned numeric_id = 0;
+      if (!is_numeric) {
+        is_numeric = true;
```
nunoplopes: it seems that is_numeric is always true after this. the code seems funny.
Also, please use something like strtoul() instead of doing it by hand

llvm_util/llvm2alive.cpp
```diff
@@ -694,13 +694,40 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
       }
     }
 
+    auto getGlobalVariable = [this](const string &name) -> llvm::GlobalVariable * {
+      auto M = f.getParent();
+      // If name is a numeric value, the result should be manually found
+      bool is_numeric = name.size() == 1 && name[0] == '0';
```
nunoplopes: I think size() is always > 1

aqjune:
Added tests, addressed comments
I was considering that global vars may have names like "0x12" or "01", but checked that bitcode reader rejected them

PR #218, opened 2019-12-10
parser support for array types

pranavk:


tools/alive_lexer.re
```diff
@@ -110,6 +110,11 @@ space+ {
   return VECTOR_TYPE_PREFIX;
 }
 
+"[" space* @tag1 [0-9][0-9]* space* "x" {
```
nunoplopes: this can be simplified to [0-9]+
The rest looks good, thanks!

PR #219, opened 2019-12-10
AggregateValue should not be registered as constant

aqjune:
After conversion of constexpr, AggregateValue may contain registers as elements, but they were registered with function->addConstant(). This causes assertion failure when exec() is run on constants.

To resolve this, this PR creates a copy instruction that assigns AggregateValue to a register (which is similar to how constexpr is converted).


tests/alive-tv/constexpr/aggregate3.tgt.ll
```diff
@@ -0,0 +1,9 @@
+@a = constant i32 0
+@x = constant { i8*, i8* } { i8* bitcast (i32* @a to i8*), i8* bitcast (i32* @a to i8*) }
+@y = constant { i64, i64 } { i64 sub (i64 ptrtoint (i32* @a to i64), i64 1), i64 0 }
```
aqjune: The initialization basic block for this input looks like this:
```
__globalvars_init:
  %__constexpr_2 = bitcast * @a to *
  %__constexpr_3 = bitcast * @a to *
  __copy_2 = {*, *} { %__constexpr_2, %__constexpr_3 }
  %__constexpr_5 = ptrtoint * @a to i64
  %__constexpr_4 = sub i64 %__constexpr_5, 1
  __copy_3 = {i64, i64} { %__constexpr_4, 0 }
  store i32 0, * @a, align 4
  store {*, *} __copy_2, * @x, align 8
  store {i64, i64} __copy_3, * @y, align 8
  br label %0
```

llvm_util/llvm2alive.cpp
```diff
@@ -708,7 +721,7 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
       char *end_ptr;
       auto numeric_id = strtoul(chrs, &end_ptr, 10);
 
-      if (end_ptr - chrs != name.size())
+      if ((unsigned)(end_ptr - chrs) != (unsigned)name.size())
```
aqjune: This was needed to relieve gcc's warning

aqjune:
Resolves 4 unit test errors

ir/instr.cpp
```diff
@@ -593,6 +597,8 @@ StateValue UnaryOp::toSMT(State &s) const {
   case FNeg:
     fn = [](auto v) { return v.fneg(); };
     break;
+  default:
```
nunoplopes: I would prefer to keep Copy explicit and not have any default, since it makes it easier when you add a new opcode.

llvm_util/llvm2alive.cpp
```diff
@@ -93,9 +94,20 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
+    auto val = v.get();
+    BB->addInstr(move(v));
+    return val;
+  }
+
+  auto get_operand(llvm::Value *v, bool enable_copy_inserter = true) {
     return llvm_util::get_operand(v,
-        [this](auto I) { return convert_constexpr(I); });
+        [this](auto I) { return convert_constexpr(I); },
+        [&](auto ag) { return enable_copy_inserter ? copy_inserter(ag) : ag; });
```
nunoplopes: it seems what we want is to introduce copies only when the aggregate is possibly non constant. Can't we detect that instead of special casing shuffle vector?

aqjune:
Addressed comments

PR #220, opened 2019-12-10
Support gep with vector operands, let num_ptrinputs count ptr elems in aggregates

aqjune:
This PR adds supports for gep with vector operands, and lets num_ptrinputs  include ptr elems in aggregates too. The reason is as follows:
```
define <2 x i8*> @f(<2 x i8*> %a) {
  ret <2 x i8*> %a
}
=>
define <2 x i8*> @f(<2 x i8*> %a) {
  ret <2 x i8*> %a
}
```
In this program, %a was not being counted into num_ptrinputs because it was aggregate, but there should have been at least one used memory block when getting formula for refinement of pointer.
To be conservative, I counted all pointer elements in aggregate operands' types as num_ptrinputs.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1826,39 +1826,65 @@ void GEP::print(std::ostream &os) const {
 }
 
 StateValue GEP::toSMT(State &s) const {
-  auto [val, non_poison] = s[*ptr];
-
-  Pointer ptr(s.getMemory(), move(val));
-  if (inbounds)
-    non_poison &= ptr.inbounds();
-
-  for (auto &[sz, idx] : idxs) {
-    auto &[v, np] = s[*idx];
-    auto multiplier = expr::mkUInt(sz, bits_for_offset);
-    auto val = v.sextOrTrunc(bits_for_offset);
-    auto inc = multiplier * val;
-
-    if (inbounds) {
-      non_poison &= multiplier.mul_no_soverflow(val);
-      non_poison &= ptr.add_no_overflow(inc);
-    }
-
-    ptr += inc;
-    non_poison &= np;
+  auto scalar = [&](const StateValue &ptrval,
+                    vector<pair<unsigned, StateValue>> &offsets)
+      -> StateValue {
+    auto non_poison = ptrval.non_poison;
+    Pointer ptr(s.getMemory(), ptrval.value);
 
     if (inbounds)
       non_poison &= ptr.inbounds();
+
+    for (auto &[sz, idx] : offsets) {
+      auto &[v, np] = idx;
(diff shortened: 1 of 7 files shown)


aqjune:
Resolves 4 unit test errors, one new error appear: Transforms/InstCombine/vec_gep_scalar_arg.ll
I think it is a bug.
What it does is:
```
...
%ins2 = insertelement <4 x *> undef, * %t3, i32 0
ret <4 x *> %ins2
=>
%1 = insertelement <4 x *> undef, * %x, i32 0
%2 = gep inbounds <4 x *> %1, 8 x i64 0, 2 x i64 3
ret <4 x *> %2
```
In the source, %ins2's 2/3/4'th element is undef, whereas %2 in the target has poison values.

llvm_util/llvm2alive.cpp
```diff
@@ -10,7 +10,7 @@
 #include "llvm/IR/Operator.h"
 #include <utility>
 #include <vector>
-
+#include <iostream>
```
nunoplopes: debug leftover

tools/transform.cpp
```diff
@@ -370,11 +370,11 @@ static void calculateAndInitConstants(Transform &t) {
...
   for (auto &arg : inputs) {
-    // An argument with aggregate type with pointer member isn't regarded
-    // as a source of non-local pointer, because its member should be extracted
-    // with extractelement / extractvalue instructions, which are also counted
-    // as source of non-local blocks.
+    // An argument with aggregate type with pointer member is dealt separately
     num_ptrinputs += arg.getType().isPtrType();
+
+    if (auto aty = arg.getType().getAsAggregateType())
+      num_ptrinputs += aty->numPointerElements();
```
nunoplopes: What's the reason for this change?  Can you give an example pls?

ir/instr.cpp
```diff
@@ -1826,39 +1826,65 @@ void GEP::print(std::ostream &os) const {
...
 
-    if (inbounds)
-      non_poison &= ptr.inbounds();
+      ptr += inc;
+      non_poison &= np;
+
+      if (inbounds)
+        non_poison &= ptr.inbounds();
+    }
+    return { ptr.release(), expr(non_poison) };
```
nunoplopes: you can move non_poison here

tools/transform.cpp
nunoplopes: I saw now your example in the other page. Can we handle that separately so we try to be less conservative. It's important to keep the number of blocks very low.

aqjune: Personally, I prefer leaving the conservative version here (so it addresses correctness first) and making another patch that optimizes it. It helps the added unit tests work as well. What do you think?

PR #222, opened 2019-12-13
Support llvm.lifetime.start/end for local blocks

aqjune:
This PR adds llvm.lifetime.start/end.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1706,14 +1706,18 @@ void Alloc::rauw(const Value &what, Value &with) {
 
 void Alloc::print(std::ostream &os) const {
   os << getName() << " = alloca " << *size << ", align " << align;
+  if (initially_dead)
+    os << ", dead";
 }
 
 StateValue Alloc::toSMT(State &s) const {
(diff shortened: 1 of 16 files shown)


aqjune:
Turns out that lifetime.start/end can be applied to non-local blocks as well (Transforms/DeadStoreElimination/lifetime.ll).
This means that the non-local blocks' readonly should be modifiable too, but I'm not sure whether it's worth the change. I'll look into it more

aqjune:
- Support lifetime.start/end for local blocks
- If alloca has lifetime.start, it is initially dead (so Memory::alloc()'s precond is set as false)
- `lifetime_start` instruction marks liveness of the pointer to true
- `free` is expanded so it can free alloca if allowed

aqjune:
I can split this PR into smaller ones if you want:
(1) add ilfetime.start/end support which is converted into no-op and raises unsupported if non-local is given
(2) add `lifetime_start` support
(3) add `lifetime_end` support

aqjune:
I'll close this and upload splitted PRs sequentially.

PR #226, opened 2019-12-14
Print 'loops are not supported' instead of 'Unsupported phi'

aqjune:
This makes alive-tv print 'Loops are not supported yet' rather than 'Unsupported phi instruction' when the phi is in a loop.
When phi is in a loop, get_operand() returns none, which led to the unsupported instruction error message.


nunoplopes:
Let's fix this properly, ie let me implement support for loops instead of adding more stuff that I need to revert later :)

PR #242, opened 2019-12-25
Support lifetime.start (2/2)

aqjune:
Key changes are:

1. Alloca now takes 'is_dead' bool. If it is true, the block's liveness is initially set as false. But, the alloca still returns a non-null pointer to allow using the alloca (e.g. bitcast, gep) before lifetime.start().
This involves updating Memory::alloc()'s return type as a pair of expr, which is <a pointer to the block, is allocated?>.

2. llvm2alive.cpp checks whether alloca has lifetime.start uses. If it has, is_dead is set as true.

3. Memory::start_lifetime encodes disjointness as well as updates memory usage, if lifetime is updated.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -886,7 +886,7 @@ pair<expr, expr> Memory::alloc(const expr &size, unsigned align, BlockKind block
   (is_local ? local_blk_kind : non_local_blk_kind)
     .add(short_bid, expr::mkUInt(alloc_ty, 2));
 
-  return { p(), allocated };
+  return { p.release(), move(allocated) };
 }
 
 expr Memory::start_lifetime(const expr &ptr_local) {
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -793,7 +793,7 @@ static expr disjoint_local_blocks(const Memory &m, const expr &addr,
   return disj;
 }
 
-expr Memory::alloc(const expr &size, unsigned align, BlockKind blockKind,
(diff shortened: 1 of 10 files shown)


aqjune:
Needs update, WIP

ir/memory.cpp
```diff
@@ -887,14 +887,34 @@ expr Memory::alloc(const expr &size, unsigned align, BlockKind blockKind,
   (is_local ? local_blk_kind : non_local_blk_kind)
     .add(short_bid, expr::mkUInt(alloc_ty, 2));
 
-  return expr::mkIf(allocated, p(), Pointer::mkNullPointer(*this)());
+  return { p(), allocated };
```
nunoplopes: p.release(), move(allocated)

ir/memory.cpp
```diff
@@ -887,14 +887,34 @@ expr Memory::alloc(const expr &size, unsigned align, BlockKind blockKind,
...
   Pointer p(*this, ptr_local);
+
+  auto align = *local_blk_align(p.get_short_bid());
+  auto size_upperbound = p.block_size() + align.zextOrTrunc(bits_size_t) -
+                         expr::mkUInt(1, bits_size_t);
+  auto allocated = size_upperbound.ule(local_avail_space.zext(1));
+
+  if (observes_addresses()) {
+    // Disjointness of block's address range with other local blocks
+    auto disj = p.get_address().add_no_uoverflow(p.block_size()) &&
```
nunoplopes: I've changed alloc() such that this condition is always added regardless of alloc result. In summary, this can be removed from here.

aqjune:
Addressed comments

PR #245, opened 2019-12-28
Support nocapture

aqjune:
This is a PR for supporting nocapture attribute.

Fully supporting it requires significant job (and its semantics is not clear when poison is involved as discussed).

This patch is WIP; feel free to leave comments. I'll be back and update this soon.

Here is the list of sanity checks of nocapture that I think should be satisfied + sanity checks from LLVM's nocapture tests.

1.
```
define void @f(i8* nocapture %p) {
    store %p, @glb // should be UB
}
```

2. 
```
define void @g(i8* nocapture %p, i8* %q) {
    // p == q can happen unless noalias is given
    store %q, @glb // should not be UB even if p == q.
}
```

3.
```
define void @g(i8* nocapture %ptr) {
  ret ptr // UB. Transforms/Attributor/nocapture-2.ll
}

define void @g(i8* nocapture %ptr) {
  ret (gep ptr, 1) // UB. Transforms/Attributor/nocapture-2.ll
}
```

4.
```
define void @f(i8** %p, i8** %q, i8* nocapture %r) { 
    i8* %tmp = *p; // %tmp can be %r
    *%p = null;
    *%q = %tmp; // this should not be UB.
}
```

The list above does not include tests regarding ptrtoints. nocapture should be applied when the pointer is casted to integer and something dependent on the value happened too, but this should be the next PR (or the future work).

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1640,6 +1640,23 @@ void Return::print(ostream &os) const {
   os << val->getName();
 }
 
+static void addUBForNoCaptureRet(State &s, const expr &bid,
+    const expr &npptr, const StateValue &svret, const Type &t) {
+  auto &[vret, npret] = svret;
+  if (t.isPtrType()) {
+    s.addUB((npptr && npret).implies(
+        Pointer(s.getMemory(), vret).get_bid() != bid));
+    return;
+  }
+
+  if (auto agg = t.getAsAggregateType()) {
+    for (unsigned i = 0, e = agg->numElementsConst(); i != e; ++i) {
+      addUBForNoCaptureRet(s, bid, npptr, agg->extract(svret, i),
+                            agg->getChild(i));
+    }
+  }
+}
+
 StateValue Return::toSMT(State &s) const {
   // Encode nocapture semantics.
   vector<expr> nonpoisons;
@@ -1655,11 +1672,7 @@ StateValue Return::toSMT(State &s) const {
     nonpoisons.emplace_back(np1);
     bids.emplace_back(bid);
 
-    if (val->getType().isPtrType()) {
-      auto &[vret, npret] = s[*val];
-      s.addUB((np1 && npret).implies(
-          Pointer(s.getMemory(), vret).get_bid() != bid));
-    }
+    addUBForNoCaptureRet(s, bid, np1, s[*val], val->getType());
   }
   if (bids.size() != 0)
     s.addUB(s.getMemory().has_noptrbyte(nonpoisons, bids));
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1674,11 +1674,12 @@ StateValue Return::toSMT(State &s) const {
     nonpoisons.emplace_back(np1);
     bids.emplace_back(bid);
   }
+  auto &retval = s[*val];
   if (bids.size() != 0) {
-    addUBForNoCaptureRet(s, nonpoisons, bids, s[*val], val->getType());
+    addUBForNoCaptureRet(s, nonpoisons, bids, retval, val->getType());
     s.addUB(s.getMemory().has_noptrbyte(nonpoisons, bids));
   }
-  s.addReturn(s[*val]);
+  s.addReturn(retval);
 
   return {};
 }
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1640,18 +1640,20 @@ void Return::print(ostream &os) const {
   os << val->getName();
 }
 
-static void addUBForNoCaptureRet(State &s, const expr &bid,
-    const expr &npptr, const StateValue &svret, const Type &t) {
+static void addUBForNoCaptureRet(State &s, const vector<expr> &nonpoisons,
+    const vector<expr> &bids, const StateValue &svret, const Type &t) {
   auto &[vret, npret] = svret;
   if (t.isPtrType()) {
-    s.addUB((npptr && npret).implies(
-        Pointer(s.getMemory(), vret).get_bid() != bid));
+    for (unsigned i = 0, e = nonpoisons.size(); i != e; ++i) {
+      s.addUB((nonpoisons[i] && npret).implies(
+          Pointer(s.getMemory(), vret).get_bid() != bids[i]));
+    }
     return;
   }
 
   if (auto agg = t.getAsAggregateType()) {
     for (unsigned i = 0, e = agg->numElementsConst(); i != e; ++i) {
-      addUBForNoCaptureRet(s, bid, npptr, agg->extract(svret, i),
+      addUBForNoCaptureRet(s, nonpoisons, bids, agg->extract(svret, i),
                             agg->getChild(i));
     }
   }
@@ -1671,11 +1673,11 @@ StateValue Return::toSMT(State &s) const {
     auto bid = Pointer(s.getMemory(), v1).get_bid();
     nonpoisons.emplace_back(np1);
     bids.emplace_back(bid);
-
-    addUBForNoCaptureRet(s, bid, np1, s[*val], val->getType());
   }
-  if (bids.size() != 0)
+  if (bids.size() != 0) {
+    addUBForNoCaptureRet(s, nonpoisons, bids, s[*val], val->getType());
     s.addUB(s.getMemory().has_noptrbyte(nonpoisons, bids));
+  }
   s.addReturn(s[*val]);
 
   return {};
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1641,7 +1641,30 @@ void Return::print(ostream &os) const {
 }
 
 StateValue Return::toSMT(State &s) const {
+  // Encode nocapture semantics.
+  auto noub = expr(true);
+  for (const auto &I: s.getFn().getInputs()) {
+    auto inp = dynamic_cast<const Input *>(&I);
+    if (!inp->getType().isPtrType() ||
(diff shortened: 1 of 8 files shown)


aqjune:
Regarding the case 3:
```
define void @g(i8* nocapture %ptr) { // nocapture keyword was missing in the original text
  ret ptr
}
=>
define void @g(i8* nocapture %ptr) {
  unreachable
}
```
This is incorrect when ptr is poison.
To resolve the case, I'll define returning a nocapture pointer to return poison rather than immediate UB.

aqjune:
After consideration I chose to stick to the simple encoding: raise UB if there exists a pointer byte with nocapture ptr's bid.

This encoding isn't perfect in sense that it causes too much UB. If there already exists nocapture pointer stored in the initial memory, it is always UB. If LLVM wants, it can be exploited for any optimization.
Also, if nocapture ptr and non-nocapture ptr are mixed, UB can silently happen (sanity check 2 and 4).

So, this encoding cannot be used for callee semantics or changes in attributes (e.g. f(%p) -> f(nocapture %p)). For example, memcpy has dst and src marked as nocapture, but we don't want memcpy to raise UB if `*dst` or `*src` contains any of pointer bytes of dst or src itself.

Also, this semantics breaks conversion from poison value to an arbitrary pointer, because nocapture does not raise UB on poison. This can happen when inbounds flag is removed from gep. 
One possible solution is to replace all escaped nocapture pointers with poison, but this will be more expensive than simply encoding UB I guess.
Also, at least, such case wasn't detected from LLVM unit tests.
The # of failures: 77 -> 85, but all of those (except Transforms/MemCpyOpt/form-memset.ll , which is https://bugs.llvm.org/show_bug.cgi?id=44388 ) are from addition of nocapture flags to function arguments.

aqjune:
Running bzip2 and gzip to see performance impact

ir/instr.cpp
```diff
@@ -1641,7 +1641,30 @@ void Return::print(ostream &os) const {
...
+    auto bid = p.get_short_bid();
+    auto cond = np1 && s.getMemory().has_noptrbyte(bid);
+
+    if (val->getType().isPtrType()) {
+      auto &[vret, npret] = s[*val];
+      Pointer pret(s.getMemory(), vret);
+      cond &= npret.implies(pret.get_short_bid() != bid);
+    }
+
+    noub &= move(cond);
```
nunoplopes: just do `s.addUB(...) everywhere`, and let it internally optimize. no need for cond.

ir/instr.cpp
```diff
@@ -1641,7 +1641,30 @@ void Return::print(ostream &os) const {
...
+  for (const auto &I: s.getFn().getInputs()) {
+    auto inp = dynamic_cast<const Input *>(&I);
+    if (!inp->getType().isPtrType() ||
+        !inp->hasAttribute(Input::NoCapture))
+      continue;
+
+    auto &[v1, np1] = s[I];
+    Pointer p(s.getMemory(), v1);
+    auto bid = p.get_short_bid();
+    auto cond = np1 && s.getMemory().has_noptrbyte(bid);
```
nunoplopes: should it be `np1.implies(..)` instead?  If the pointer is poison, then it doesn't escape.

ir/instr.cpp
```diff
@@ -1641,7 +1641,30 @@ void Return::print(ostream &os) const {
...
+
+    if (val->getType().isPtrType()) {
```
nunoplopes: what about aggregate types with ptrs?

ir/memory.cpp
```diff
@@ -1233,6 +1233,23 @@ pair<expr,Pointer> Memory::refined(const Memory &other) const {
...
+  auto res = expr(true);
+  auto ofs = expr::mkFreshVar("ofs", expr::mkUInt(0, bits_for_offset));
+
+  for (unsigned bid = 1; bid < num_nonlocals; ++bid) {
+    Pointer p(*this, expr::mkUInt(bid, bits_for_bid), ofs);
+    Byte b(*this, non_local_block_val.load(p.short_ptr()));
+    Pointer loadp(*this, b.ptr_value());
+    res &= p.is_block_alive().implies(
+      expr::mkForAll({ ofs }, (b.is_ptr() && b.ptr_nonpoison()).implies(
+                                  loadp.get_short_bid() != short_bid)));
```
nunoplopes: you can't use short_bid here; we need to compare the full bid, otherwise this code also excludes the local bid with the same number, which is not what we intend.

ir/instr.cpp
aqjune: You're right, it was incorrect

ir/instr.cpp
aqjune: Aggregate arguments does not support nocapture attr, so it's okay.

aqjune:
Addressed the comments

ir/instr.cpp
aqjune: It was aggregate value return, sorry. Updated

tests/alive-tv/attrs/nocapture.src.ll
```diff
@@ -0,0 +1,32 @@
...
+define i8* @f3(i8* nocapture %p) {
+  ret i8* %p
+}
+
+define i8* @f4(i8* nocapture %p) {
+  %p2 = getelementptr i8, i8* %p, i32 1
+  ret i8* %p2
+}
+
+;define {i8*, i8*} @f5(i8* nocapture %p) {
```
nunoplopes: why are these commented out?

aqjune: Because insertvalue is not supported yet.
```
ERROR: Unsupported instruction:   %v = insertvalue { i8*, i8* } undef, i8* %p, 1
```
These tests are needed for checking whether aggregate value return is working correctly. Should be uncommented after insertvalue support is added

nunoplopes: what if the return value is a vector instead?  Then you can use insertelement

aqjune: Yes, it works. Updated the test

ir/memory.cpp
```diff
@@ -1233,6 +1233,26 @@ pair<expr,Pointer> Memory::refined(const Memory &other) const {
...
+  for (unsigned bid = 1; bid < num_nonlocals; ++bid) {
+    Pointer p(*this, expr::mkUInt(bid, bits_for_bid), ofs);
+    Byte b(*this, non_local_block_val.load(p.short_ptr()));
+    Pointer loadp(*this, b.ptr_value());
+
+    expr c(true);
+    for (unsigned i = 0, e = bids.size(); i < e; ++i)
+      c &= conds[i].implies(loadp.get_bid() != bids[i]);
+    res &= p.is_block_alive().implies(
+      expr::mkForAll({ ofs }, (b.is_ptr() && b.ptr_nonpoison()).implies(c)));
```
nunoplopes: mkforall here is probably not correct.
This is a UB expression, so it's very likely you want state->addQuantVar(). If you add a few negative tests (like replace a store undef -> store capturing a nocapture ptr), it may show the problem

ir/instr.cpp
```diff
@@ -1640,8 +1640,44 @@ void Return::print(ostream &os) const {
...
+    if (!inp->getType().isPtrType() ||
+        !inp->hasAttribute(Input::NoCapture))
+      continue;
+
+    auto &[v1, np1] = s[I];
+    auto bid = Pointer(s.getMemory(), v1).get_bid();
+    nonpoisons.emplace_back(np1);
+    bids.emplace_back(bid);
+
+    addUBForNoCaptureRet(s, bid, np1, s[*val], val->getType());
```
nunoplopes: this is reevaluating the return value many times, including the return stmt below

ir/memory.cpp
aqjune: Could you elaborate a bit please?

ir/instr.cpp
```diff
@@ -1640,8 +1640,46 @@ void Return::print(ostream &os) const {
...
+    auto &[v1, np1] = s[I];
+    auto bid = Pointer(s.getMemory(), v1).get_bid();
+    nonpoisons.emplace_back(np1);
+    bids.emplace_back(bid);
+  }
+  if (bids.size() != 0) {
+    addUBForNoCaptureRet(s, nonpoisons, bids, s[*val], val->getType());
+    s.addUB(s.getMemory().has_noptrbyte(nonpoisons, bids));
+  }
   s.addReturn(s[*val]);
```
nunoplopes: s[*val] is still evaluated twice

ir/memory.cpp
aqjune: If ofs is defined as a quantified variable, UB nondeterministically becomes true or false depending on which ofs was chosen. (It is assumed that ofs in src and tgt are independent, as the result of freeze instruction does)
In this case, source can choose the ofs which contains the nocapture bid, leading src to have UB. So the expressiveness is equivalent to the case where forall is used.

nunoplopes:
I had to revert this PR because I realized it wasn't correct. We can't implement it by checking bid in memory, since a function may have 2 arguments, one nocapture and another without nocapture and receive the same parameter in both arguments.
I guess this requires the same trick as readonly: another bit in pointer to track provenance's attributes.

aqjune:
Makes sense, that will work. Couldn't think that the trick for readonly can be applied to nocapture.
I'll work on this and make PR tomorrow if you want.

PR #253, opened 2020-01-02
Add attribute bits to pointer & support nocapture

aqjune:
Splitted the diffs into two commits for easier review.

LLVM unit test failures: 63 -> 67

I think all 4 added failures are bugs in alive2 - 1 is the floating point bug, 3 are oggenc's DSE example. I'm investigating the reason.
```
Transforms/SLPVectorizer/X86/phi.ll: floating point bug
Transforms/DeadStoreElimination/OverwriteStoreBegin.ll
Transforms/DeadStoreElimination/OverwriteStoreEnd.ll
Transforms/MemCpyOpt/profitable-memset.ll
```

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1233,9 +1233,7 @@ StateValue ICmp::toSMT(State &s) const {
        elem_ty.getAsAggregateType()->getChild(0).isPtrType())) {
     fn = [&](auto &av, auto &bv, Cond cond) {
       Pointer lhs(s.getMemory(), av);
-      lhs.strip_attrs();
       Pointer rhs(s.getMemory(), bv);
-      rhs.strip_attrs();
       switch (cond) {
       case EQ:  return StateValue(lhs == rhs, true);
       case NE:  return StateValue(lhs != rhs, true);
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1233,7 +1233,9 @@ StateValue ICmp::toSMT(State &s) const {
        elem_ty.getAsAggregateType()->getChild(0).isPtrType())) {
     fn = [&](auto &av, auto &bv, Cond cond) {
       Pointer lhs(s.getMemory(), av);
+      lhs.strip_attrs();
       Pointer rhs(s.getMemory(), bv);
+      rhs.strip_attrs();
       switch (cond) {
       case EQ:  return StateValue(lhs == rhs, true);
       case NE:  return StateValue(lhs != rhs, true);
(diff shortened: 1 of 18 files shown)


ir/value.cpp
```diff
@@ -187,6 +189,11 @@ StateValue Input::toSMT(State &s) const {
   if (attributes & NonNull)
     s.addPre(Pointer(s.getMemory(), val).isNonZero());
 
+  if (getType().isPtrType()) {
+    auto nocap = Pointer(s.getMemory(), val).is_nocapture();
+    s.addAxiom((attributes & NoCapture) ? nocap : !nocap);
```
nunoplopes: I think this should be folded in mkInput/mkUndefInput, as it's much more efficient to concatenate another bit than adding an axiom.

tools/transform.cpp
```diff
@@ -538,8 +538,16 @@ static void calculateAndInitConstants(Transform &t) {
...
   bool does_mem_access = false;
   bool has_load = false;
+  // The number of bits needed to encode pointer attributes
+  bits_for_ptrattrs = 0;
 
   for (auto fn : { &t.src, &t.tgt }) {
+    for (auto &v : fn->getInputs()) {
+      auto i = dynamic_cast<const Input*>(&v);
+      if (i && i->hasAttribute(Input::NoCapture))
+        bits_for_ptrattrs = 1;
```
nunoplopes: this pattern makes it harder to add other attributes. You need a has_foo variables (at least locally) to then sum them up.

ir/memory.cpp
```diff
@@ -673,9 +693,15 @@ expr Pointer::is_writable() const {
   return is_local() || non_local;
 }
 
+expr Pointer::is_nocapture() const {
+  if (bits_for_ptrattrs == 0)
+    return false;
+  return get_attrs() == 1;
```
nunoplopes: this is not correct once we have more attributes.  I would do an extract here directly and remove get_attrs()

ir/instr.cpp
```diff
@@ -1233,7 +1233,9 @@ StateValue ICmp::toSMT(State &s) const {
        elem_ty.getAsAggregateType()->getChild(0).isPtrType())) {
     fn = [&](auto &av, auto &bv, Cond cond) {
       Pointer lhs(s.getMemory(), av);
+      lhs.strip_attrs();
```
nunoplopes: this is weird.
I would change operator==/!= instead.

aqjune:
Addressed comments.
Interestingly, the 3 bugs (DSE, MemCpyOpt) disappeared..! Something is happening. I remember they were related with undef values, so something related with precondition + quantifier is happening. #254 may be giving hints..

PR #255, opened 2020-01-03
add realloc

Dongjoo-Kim:
I added realloc instruction.

Realloc is implemented using malloc, memcpy and free.
First, allocate pointer with the size.
If the allocation is success, copy the previous pointer's value to new pointer and free it.
If not, preserve the previous pointer.

However, there is one restriction.
In this implementation, allocation is run before free.
So, the result of realloc is always disjoint with the previous one.
If free precede allocation, then memcpy isn't available.
Could you help me in this matter?

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1870,7 +1870,7 @@ expr Malloc::getTypeConstraints(const Function &f) const {
 
 unique_ptr<Instr> Malloc::dup(const string &suffix) const {
   if (ptr)
-    return make_unique<Malloc>(getType(), getName() + suffix, *size, *ptr);
+    return make_unique<Malloc>(getType(), getName() + suffix, *ptr, *size);
   return make_unique<Malloc>(getType(), getName() + suffix, *size, isNonNull);
 }
 
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1821,21 +1821,27 @@ void Malloc::print(std::ostream &os) const {
 
 StateValue Malloc::toSMT(State &s) const {
   if (!isRealloc) {
-    auto &[sz, np] = s[*size];
+    auto &[sz, np] = s.getAndAddUndefs(*size);
     // TODO: malloc's alignment is implementation defined.
     expr nonnull = expr::mkBoolVar("malloc_never_fails");
     auto [p, allocated] = s.getMemory().alloc(sz, 8, Memory::HEAP, np, nonnull);
 
+    if (isNonNull) {
+      // TODO: In C++ we need to throw an exception if the allocation fails, but
+      // exception hasn't been modeled yet
+      s.addPre(move(allocated));
+    }
+
     return { move(p), expr(np) };
   } else {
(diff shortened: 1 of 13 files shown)


ir/instr.h
```diff
@@ -399,13 +399,17 @@ class Alloc final : public Instr {
...
+  Value *ptr, *size;
   // Is this malloc (or equivalent operation, like new()) never returning
   // null?
   bool isNonNull;
+  bool isRealloc;
+
 public:
-  Malloc(Type &type, std::string &&name, Value &size, bool isNonNull)
-    : Instr(type, std::move(name)), size(&size), isNonNull(isNonNull) {}
+  Malloc(Type &type, std::string &&name, Value &ptr, Value &size,
```
nunoplopes: what about adding two constructors, one for malloc (single value), and another for realloc (2 values)?

tests/alive-tv/memory/realloc.src.ll
```diff
@@ -0,0 +1,31 @@
+target datalayout = "e-m:o-i64:64-f80:128-n8:16:32:64-S128"
+
+define i8 @realloc-check() {
+  %ptr = call noalias i8* @malloc(i64 4)
+  %i = ptrtoint i8* %ptr to i64
```
nunoplopes: please comparison with null, since it makes the test much faster

tests/alive-tv/memory/realloc.src.ll
```diff
@@ -0,0 +1,31 @@
...
+  ret i8 -1
+BB2:
+  store i8 5, i8* %ptr
+  %ptr2 = call noalias i8* @realloc(i8* %ptr, i64 1)
+  %i2 = ptrtoint i8* %ptr2 to i64
+  %cmp2 = icmp eq i64 %i2, 0
+  br i1 %cmp2, label %BB3, label %BB4
+BB3:
+  %a = load i8, i8* %ptr
+  store i8 1, i8* %ptr
```
nunoplopes: can you pls add a negative test that is similar to this one, but where the return value is wrong?

ir/instr.cpp
```diff
@@ -1800,40 +1800,80 @@ unique_ptr<Instr> Alloc::dup(const string &suffix) const {
...
+    return { move(p), expr(np) };
+  } else {
+    auto &[p, np_ptr] = s.getAndAddUndefs(*ptr);
+    auto &[sz, np_size] = s.getAndAddUndefs(*size);
+    s.addUB(np_ptr);
+
+    // If sz > p_sz, it is filled it with poison (local_block_val is
+    // initialized with poison).
+    expr nonnull = expr::mkBoolVar("malloc_never_fails");
+    auto [p_new, allocated] = s.getMemory().alloc(sz, 8, Memory::HEAP, np_size,
```
nunoplopes: some of this code can be shared with malloc; please hoist it out of the if,

Dongjoo-Kim:
I reflected your comments.
Sorry for late response.

PR #262, opened 2020-01-07
Register undef variables used by the size argument of malloc

aqjune:
This helps the removal of alloca test work even when `-disable-undef-input` is not given, by registering the undef Z3 variables used by the size value.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1758,9 +1758,7 @@ void Malloc::print(std::ostream &os) const {
 }
 
 StateValue Malloc::toSMT(State &s) const {
-  auto &[sz, np] = s[*size];
-  for (auto uvar: s.at(*size).second)
-    s.addQuantVar(move(uvar));
+  auto &[sz, np] = s.getAndAddUndefs(*size);
   // TODO: malloc's alignment is implementation defined.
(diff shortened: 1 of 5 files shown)


aqjune:
The motivation behind this patch is that the undef variables of size are not being quantified, and this led to the incorrect result.

nunoplopes:
this looks fishy at 1st glance. undef variables are already quantified in transform.cpp. There's something wrong somewhere else.

aqjune:
This happens when dead code is eliminated. Undef variable is lost at the end of unused instruction.
```
define i8* @f_observed(i64 %size) { ; %size's uvar: undef!2
entry:
  %items.addr = alloca i32, align 8
  %call = call noalias i8* @malloc(i64 %size) ; %call's uvar: undef!2
  %unused = ptrtoint i8* %call to i64 ; %unused's uvar: undef!2
  ret i8* %call ; creates a new undef variable, so undef!2 is not quantified.
}
=>
define i8* @f_observed(i64 %size) {
entry:
  %call = call noalias i8* @malloc(i64 %size)
  ret i8* %call
}
```

nunoplopes:
Ok, sounds good. I would prefer we added a new function to State that is like operator[] but registers undef list vars right away in the quant vars.

aqjune:
Addressed the comment

PR #268, opened 2020-01-13
Let GEP's ptr duplicate undef vars when idx is a vector type

aqjune:
This PR adds constraint to undef aggregate value's pointer elements so they have correct values.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2019,8 +2019,8 @@ StateValue GEP::toSMT(State &s) const {
         else
           offsets.emplace_back(sz, s[*idx]);
       }
-      vals.emplace_back(scalar(ptr_isvect ? aty->extract(ptrval, i) : ptrval,
-                               offsets));
+      vals.emplace_back(scalar(ptr_isvect ? aty->extract(ptrval, i) :
+                               (i == 0 ? ptrval : s[*ptr]), offsets));
     }
     return getType().getAsAggregateType()->aggregateVals(vals);
   }
--- a/ir/value.cpp
+++ b/ir/value.cpp
@@ -43,13 +43,11 @@ void UndefValue::print(ostream &os) const {
   UNREACHABLE();
 }
 
-
 StateValue UndefValue::toSMT(State &s) const {
   auto val = getType().getDummyValue(true);
   expr var = expr::mkFreshVar("undef", val.value);
   s.addUndefVar(expr(var));
-  StateValue undefv(move(var), move(val.non_poison));
-  return undefv;
+  return { move(var), move(val.non_poison) };
 }
 
 
--- a/ir/value.cpp
+++ b/ir/value.cpp
@@ -43,11 +43,13 @@ void UndefValue::print(ostream &os) const {
   UNREACHABLE();
 }
 
+
 StateValue UndefValue::toSMT(State &s) const {
   auto val = getType().getDummyValue(true);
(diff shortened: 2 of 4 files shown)


nunoplopes:
The testcase passes without your patch. Maybe you reduced it too much?

The SMT formula looks good to me as well:
```
(declare-fun undef!4 () (_ BitVec 10))
(declare-fun blk_align_tgt ((_ BitVec 1)) (_ BitVec 8))
(declare-fun blk_align_src ((_ BitVec 1)) (_ BitVec 8))
(declare-fun blk_size ((_ BitVec 1)) (_ BitVec 4))
(assert (let ((a!1 (forall ((undef!2 (_ BitVec 5)) (undef!3 (_ BitVec 5)))
             (not (and (= ((_ extract 9 5) undef!4) undef!2)
                       (= ((_ extract 4 0) undef!4) undef!3))))))
  (and (= (blk_size #b0) #x0)
       (= #x00 (blk_align_src #b0))
       (= #x00 (blk_align_tgt #b0))
       a!1)))
```

nunoplopes:
For reference, here's the original unit test:
```
define <2 x *> @tinkywinky() {
%0:
  %patatino = gep * undef, 1 x <2 x i64> undef
  ret <2 x *> %patatino
}
=>
define <2 x *> @tinkywinky() {
%0:
  ret <2 x *> undef
}
ERROR: Value mismatch

Example:

Source:
<2 x *> %patatino = < undef, undef >

Target:
Source value: < undef, undef >
Target value: < pointer(non-local, block_id=1, offset=0), pointer(non-local, block_id=0, offset=32) >
```

nunoplopes:
Another case to consider:
```
%p = phi(undef, %q)
load %p
=>
load %q
```

nunoplopes:
That one is fine because it has UB in src.
But there's also return value refinement:
```
return foo ? undef : p;
=>
return p;
```

And fn call refinement:
```
call @f(foo ? undef : p)
=>
call @f(p)
```

aqjune:
```
return foo ? undef : p;
=>
return p;
```
If foo is true and p is poison, undef cannot be optimized to p, so this is fine I think.
When p is constant, alive2 successfully verifies, which is good:
```
@z = global 1 bytes, align 1

define * @f(i1 %c) {
%0:
  %g = select i1 %c, * undef, * @z
  ret * %g
}
=>
@z = global 1 bytes, align 1

define * @f(i1 %c) {
%0:
  ret * @z
}
```

nunoplopes:
Verifies it right now, but not sure it does with this patch.

aqjune:
I see..
The core problem is that currently there's no encoding between undef ptr (which points to the null block) and global constants.
But I guess the refinement must be pretty hard to encode, because it needs to identify whether a pointer value is undef or not.
I think we should say that UndefValue::toSMT (that returns arbitrary bit pattern, including bid) is encoding an approximation of undef; otherwise the refinement from undef to global constant is hard to explain.

aqjune:
```
define <2 x *> @tinkywinky() {
%0:
  %patatino = gep * undef, 1 x <2 x i64> undef
  ret <2 x *> %patatino
}
=>
define <2 x *> @tinkywinky() {
%0:
  ret <2 x *> undef
}
ERROR: Value mismatch
```
For this example, this updated patch seems to explain it.
But I'm not sure whether this is worth its change; this will fasten duplication of undef quantified variables n times more.

PR #270, opened 2020-01-16
Pointer::get_value should check validity of p

aqjune:
Sanity check (this couldn't be added to unit test because src identity check should succeed):
```
define i64 @f(i64 %x, i64 %y) {
  %p = inttoptr i64 %x to i64*
  %a = load i64, i64* %p
  ret i64 %a
}
=>
define i64 @f(i64 %x, i64 %y) {
  unreachable
}
```
This example should raise `Invalid expr`.


aqjune:
#271 will let the added unit test succeed by ignoring invalid expr case

PR #276, opened 2020-01-25
Support the case when callee's attribute has byval

aqjune:
This PR deal with the case when the callee's pointer argument has byval attribute.

```
define i8 @f6_2(i8* %p) {
  %a = alloca i8
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %a, i8* %p, i64 1, i1 false)
  %b = call i8 @g2(i8* byval %a)
  ret i8 %b
}
declare i8 @g2(i8* byval)
=>
define i8 @f6_2(i8* %p) {
  %b = call i8 @g2(i8* byval %p)
  ret i8 %b
}
```

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1112,8 +1112,9 @@ unique_ptr<Instr> ExtractValue::dup(const string &suffix) const {
 }
 
 
-void FnCall::addArg(Value &arg) {
+void FnCall::addArg(Value &arg, unsigned flags) {
   args.emplace_back(&arg);
+  argflags.emplace_back(flags);
 }
 
 vector<Value*> FnCall::operands() const {
@@ -1154,18 +1155,19 @@ void FnCall::print(ostream &os) const {
     os << "\t; WARNING: unknown known function";
 }
 
-static void unpack_inputs(State&s, Type &ty, const StateValue &value,
-                          vector<StateValue> &inputs,
-                          vector<StateValue> &ptr_inputs) {
+static void unpack_inputs(State&s, Type &ty, unsigned argflag,
+                          const StateValue &value, vector<StateValue> &inputs,
+                          vector<pair<StateValue, bool>> &ptr_inputs) {
   if (auto agg = ty.getAsAggregateType()) {
     for (unsigned i = 0, e = agg->numElementsConst(); i != e; ++i) {
-      unpack_inputs(s, agg->getChild(i), agg->extract(value, i), inputs,
-                    ptr_inputs);
+      unpack_inputs(s, agg->getChild(i), argflag, agg->extract(value, i),
+                    inputs, ptr_inputs);
     }
   } else if (ty.isPtrType()) {
     Pointer p(s.getMemory(), value.value);
     p.strip_attrs();
-    ptr_inputs.emplace_back(p.release(), expr(value.non_poison));
+    ptr_inputs.emplace_back(StateValue(p.release(), expr(value.non_poison)),
+                            argflag & FnCall::ArgByVal);
   } else {
     inputs.emplace_back(value);
   }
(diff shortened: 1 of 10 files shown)


ir/instr.cpp
```diff
@@ -1100,18 +1101,19 @@ void FnCall::print(ostream &os) const {
     os << "\t; WARNING: unknown known function";
 }
 
-static void unpack_inputs(State&s, Type &ty, const StateValue &value,
-                          vector<StateValue> &inputs,
-                          vector<StateValue> &ptr_inputs) {
+static void unpack_inputs(State&s, Type &ty, unsigned argflag,
+                          const StateValue &value, vector<StateValue> &inputs,
+                          vector<pair<StateValue, bool>> &ptr_inputs) {
```
aqjune: ptr_input carries (pointer value, is byval?) pairs.

ir/memory.cpp
```diff
@@ -643,7 +643,7 @@ expr Pointer::fninput_refined(const Pointer &other) const {
                                            (size2 - off2).uge(size - off),
                                          off2.sgt(size2) && off == off2 &&
                                            size2.uge(size))));
-  local = (other.is_local() || other.is_byval()) && local;
+  local = (other.is_local() || other.is_byval() || is_byval_arg) && local;
```
aqjune: If is_byval_arg is true, `other` pointer is regarded as byval.

ir/memory.cpp
```diff
@@ -1045,16 +1045,18 @@ pair<expr, expr> Memory::mkUndefInput(unsigned attributes) const {
...
+Memory::mkFnRet(const char *name,
+                const vector<pair<StateValue, bool>> &ptr_inputs) const {
   expr var
     = expr::mkFreshVar(name, expr::mkUInt(0, bits_for_bid + bits_for_offset));
   Pointer p(*this, var.concat_zeros(bits_for_ptrattrs));
 
   set<expr> local;
   for (auto &in : ptr_inputs) {
-    Pointer inp(*this, in.value);
+    // TODO: callee cannot observe the bid if this is byval.
```
aqjune: Several decisions that are not directly relevant with the example are made conservatively

ir/state.cpp
```diff
@@ -372,11 +372,14 @@ void State::mkAxioms(State &tgt) {
         for (unsigned i = 0, e = ptr_ins.size(); i != e; ++i) {
           // TODO: needs to take read/read2 as input to control if mem blocks
           // need to be compared
-          expr eq_val = Pointer(mem, ptr_ins[i].value)
-                          .fninput_refined(Pointer(mem2, ptr_ins2[i].value));
+          auto &[ptr_in, is_byval] = ptr_ins[i];
+          auto &[ptr_in2, is_byval2] = ptr_ins2[i];
+          (void)is_byval;
```
aqjune: Assumes that we're only dealing with transformations that don't add byval attributes ; adding such attribute would be interprocedural

aqjune:
Ran unit tests, and it works modulo the case when a global contant is given to byval.
e.g)
```
%p = alloca
memcpy(%p, @glb_const)
f(byval %p)
->
f(byval @glb_const)
```
I'll defer this change: this does not appear in the single file tests.
If properly load-balanced across machines, It takes 40 hrs to run whole single-file tests.

llvm_util/llvm2alive.cpp
```diff
@@ -256,11 +256,20 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
     string fn_name = '@' + fn->getName().str();
     auto call = make_unique<FnCall>(*ty, value_name(i), move(fn_name), flags,
                                     !known);
-    for (auto &arg : i.args()) {
```
nunoplopes: it seems the iterator could be kept as before. There's no benefit in this change AFAICT

ir/instr.h
```diff
@@ -173,17 +173,19 @@ class FnCall final : public Instr {
 public:
   enum Flags { None = 0, NoRead = 1 << 0, NoWrite = 1 << 1, ArgMemOnly = 1 << 2,
                NNaN = 1 << 3 };
+  enum ArgFlags { ArgNone = 0, ArgByVal = 1 << 0 };
 private:
   std::string fnName;
   std::vector<Value*> args;
```
nunoplopes: I think it's preferable to make this a vector<pair<>> instead. Simplifies a bunch of code & ensure the invariant that every arg has an attribute.

ir/state.cpp
nunoplopes: we should assume as little as possible.
This needs to handle the case where byval is added or removed. It can just set refines to false, for example.

aqjune:
Rebased & addressed comments

PR #279, opened 2020-01-27
Add support for fma intrinsic

kevindevos:
This adds support for the fma (fused multiply-add) llvm intrinsic.

I'm unsure if i did the fma.opt unit test correctly, as i added fast math flags to some of the tests. However i do not understand why other floating point tests such as fadd do not fail even when fast math flags are not specified. Shouldn't they fail when one of the inputs to fadd could be NaN?

Overloading fm_poison in instr.cpp was a little tricky, i'm not sure if there is a better solution without code duplication.


tests/unit/fp/fma-error.opt
```diff
@@ -0,0 +1,11 @@
+Name: target more poisonous than source
+; ERROR: Target is more poisonous than source for float %t
+%t = fma float %a, float %b, float %c
+  =>
+%t = poison
+
+Name: value mismatch (2.0 * 4.0) + 1.0 => 8.0
```
nunoplopes: due to limitations in the unit test runner, only 1 failing test per file is supported.

ir/instr.cpp
```diff
@@ -700,52 +721,74 @@ void TernaryOp::print(ostream &os) const {
...
   auto &bv = s[*b];
   auto &cv = s[*c];
-  function<expr(const expr&, const expr&, const expr&)> fn;
+  function<StateValue(const expr&, const expr&, const expr&)> fn;
 
   switch (op) {
   case FShl:
-    fn = [](auto a, auto b, auto c) {
-      return expr::fshl(a, b, c);
+    fn = [&](auto a, auto b, auto c) -> StateValue {
```
nunoplopes: & doesn't seem needed here

ir/instr.cpp
```diff
@@ -700,52 +721,74 @@ void TernaryOp::print(ostream &os) const {
...
+      return { expr::fshl(a, b, c), true };
     };
     break;
+
   case FShr:
-    fn = [](auto a, auto b, auto c) {
-      return expr::fshr(a, b, c);
+    fn = [&](auto a, auto b, auto c) -> StateValue {
```
nunoplopes: idem

ir/instr.cpp
```diff
@@ -700,52 +721,74 @@ void TernaryOp::print(ostream &os) const {
...
+      return { expr::fshr(a, b, c), true };
+    };
+    break;
+
+  case FMA:
+    fn = [&](auto a, auto b, auto c) -> StateValue{
```
nunoplopes: no space after StateValue

ir/instr.cpp
```diff
@@ -700,52 +721,74 @@ void TernaryOp::print(ostream &os) const {
...
+      return fm_poison(s, a, b, c, [](expr &a, expr &b, expr &c) { 
+                                   return expr::fma(a, b, c); }, fmath, false);
     };
     break;
   }
 
   if (getType().isVectorType()) {
-    vector<StateValue> vals;
```
nunoplopes: this change seems unrelated

ir/instr.cpp
```diff
@@ -231,6 +232,14 @@ static StateValue fm_poison(State &s, expr a, expr b,
   return { move(val), move(non_poison) };
 }
 
+static StateValue fm_poison(State &s, expr a, expr b,
+                            function<expr(expr&,expr&)> fn,
+                            FastMathFlags fmath, bool only_input) {
+  return fm_poison(s, a, b, expr(),
```
nunoplopes: expr() doesn't work here. It will make all non_poison expressions above to be null.

ir/instr.cpp
```diff
@@ -680,6 +689,18 @@ unique_ptr<Instr> UnaryOp::dup(const string &suffix) const {
...
 
+TernaryOp::TernaryOp(Type &type, std::string &&name, Value &a, Value &b, Value &c, Op op,
+            FastMathFlags fmath)
+    : Instr(type, std::move(name)), a(&a), b(&b), c(&c), op(op), fmath(fmath) {
+  switch (op) {
+    case FShr:
+    case FShl:
+      assert(fmath.isNone());
+      break;
+    default:
```
nunoplopes: no default pls. That's a hazard waiting to explode in the future

tests/unit/fp/fma-error.opt
kevindevos: Fixed, split into 2 files, fma-error-mismatch.opt and fma-poison-error.opt.

ir/instr.cpp
kevindevos: Fixed

ir/instr.cpp
kevindevos: Fixed

ir/instr.cpp
kevindevos: Changed back to how it was before. This was only a line swap for it to be consistent with BinOp::toSMT().

ir/instr.cpp
kevindevos: Fixed, changed to a case.

ir/instr.cpp
kevindevos: Fixed, but kept this the same. I experimented passing expr(true) but that gets simplified and gives the same result as expr(). Instead the passed expr c is tested with isValid() before doing any operation with it, and if expr() is passed, nothing is added to the non_poison expr.

kevindevos:
Apologies, had to force push since i did something wrong on my side with git and the commit included some code already in master of alive.

ir/instr.cpp
nunoplopes: That doesn't work because c might be null.
You need an extra argument, say "bool is_ternary = true".

nunoplopes: Also, in this line you need move() for a & b.

ir/instr.cpp
```diff
@@ -680,6 +691,18 @@ unique_ptr<Instr> UnaryOp::dup(const string &suffix) const {
   return make_unique<UnaryOp>(getType(), getName() + suffix, *val, op, fmath);
 }
 
+TernaryOp::TernaryOp(Type &type, std::string &&name, Value &a, Value &b, Value &c, Op op,
```
nunoplopes: this line is clearly over 80 chars.
Also, no need for std:: there.

ir/instr.cpp
```diff
@@ -680,6 +691,18 @@ unique_ptr<Instr> UnaryOp::dup(const string &suffix) const {
...
+            FastMathFlags fmath)
+    : Instr(type, std::move(name)), a(&a), b(&b), c(&c), op(op), fmath(fmath) {
```
nunoplopes: nor std:: here

ir/instr.cpp
kevindevos: Thanks!, all fixed, i need to pay more attention.

Issue #212, opened 2019-12-07
Add support for introducing global constants


nunoplopes:
There's no reason we can't support introduction of global constants. e.g., as in the `switch-to-lookup` pass. Example test: Transforms/SimplifyCFG/X86/switch_to_lookup_table.ll

PR #283, opened 2020-02-06
Support introduction of global constants in target (#212)

aqjune:
Addresses #212
Skips refinement check of a block if it is global constant in target but not in source.


aqjune:
```
define i32 @f(i32* %p) {     
  %x = load i32, i32* %p     
  ret i32 %x                 
}                            
=>
@c = constant i32 0          
define i32 @f(i32* %p) {     
  %dummy = load i32, i32* @c 
  %x = load i32, i32* %p     
  ret i32 %x                 
}                            
```

ir/memory.cpp
```diff
@@ -365,12 +365,16 @@ unsigned Pointer::total_bits() {
...
   if (num_locals == 0)
     return false;
-  if (num_nonlocals == 0)
+  if (m.num_nonlocals() == 0)
     return true;
   auto bit = total_bits() - 1;
   return p.extract(bit, bit) == 1;
 }
 
+expr Pointer::nonlocal_wf() const {
```
aqjune: wf stands for wellformedness

tests/alive-tv/glbconstchk3.tgt.ll
```diff
@@ -0,0 +1,5 @@
+@g = constant i32 0
+define i32 @f() {
+  %x = load i32, i32* @g
```
nunoplopes: please add a failing test as well.

ir/memory.cpp
```diff
@@ -1518,7 +1521,19 @@ pair<expr,Pointer> Memory::refined(const Memory &other,
...
-  for (unsigned bid = 1; bid < num_nonlocals; ++bid) {
+  auto is_constglb = [](const Memory &m, unsigned bid) {
+    return
+      find(m.non_local_blk_nonwritable.begin(),
+           m.non_local_blk_nonwritable.end(), bid) !=
+        m.non_local_blk_nonwritable.end();
+  };
+
+  for (unsigned bid = 1; bid < IR::num_nonlocals; ++bid) {
+    if (!is_constglb(*this, bid) && is_constglb(other, bid)) {
```
nunoplopes: this check can be replaced with num_nonlocals_src or something like that?

ir/memory.cpp
nunoplopes: I don't really like this function because it only makes sense for non_local ptrs. It's wrong to call it otherwise.
We can't assert or do something else because it's context sensitive. I would rollback this change; it's unrelated with this PR.

ir/memory.cpp
nunoplopes: this case can only happen between num_nonlocals_src and num_nonlocals_tgt. It needs an assert the very least.

aqjune: Updated

tests/alive-tv/glbconstchk3.tgt.ll
aqjune: tests/alive-tv/refinement/const-fail.src.ll contains the test for mismatch between initial values of constants in src and tgt.

ir/memory.cpp
aqjune: I thought it makes sense because otherwise repeated changes on ule(num_nonlocal) were needed.
Also it is natural to have a function for encoding wellformedness of a pointer; local pointers don't need wellformedness right now, that was why nonlocal pointers were only considered.

The change seemed orthogonal with this PR so I rollbacked it, but I think it will be great if the refactoring is accepted before I work on #284 ; I have an idea for #284, but it will again make repeated changes regarding num_nonlocal.

aqjune:
Moved tests to alive-tv/refinement directory because this patch involved updates to refinement

PR #285, opened 2020-02-07
a few changes, the main one is to allow people to verify an optimization without trying to remember the tricky "opt" invocation

regehr:
- after a CMake update (on Ubuntu 18.04) we now need to explicitly find ZLIB

- some improvements to alive-tv's help output: a more detailed
  summary, removed "Alive:" from in front of the help text since our
  options are clearly in the alive option category, and tell the
  command line library to hide all unrelated command line options

- support running alive-tv with just a single file, in which case
  we run opt -O2 on it and verify refinment of that

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,8 @@ if (CYGWIN)
   set(CMAKE_FIND_LIBRARY_SUFFIXES ".dll")
 endif()
 
+find_package(ZLIB REQUIRED)
+
 find_package(Z3 4.8.5 REQUIRED)
 include_directories(${Z3_INCLUDE_DIR})
 
@@ -150,7 +152,7 @@ target_link_libraries(alive PRIVATE ${ALIVE_LIBS})
 add_library(alive2 SHARED ${IR_SRCS} ${SMT_SRCS} ${TOOLS_SRCS} ${UTIL_SRCS} ${LLVM_UTIL_SRCS})
 
 if (BUILD_LLVM_UTILS OR BUILD_TV)
-  llvm_map_components_to_libnames(llvm_libs support core irreader analysis)
+  llvm_map_components_to_libnames(llvm_libs support core irreader analysis passes)
   target_link_libraries(alive2 PRIVATE ${llvm_libs})
   target_link_libraries(alive-tv PRIVATE ${ALIVE_LIBS_LLVM} ${llvm_libs})
 endif()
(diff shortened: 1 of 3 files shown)


nunoplopes:
I like a lot the idea!
Though we can't really -O2, otherwise there'll be tons of false-positives due to IPO passes.
I also trying to find an LLVM API that would give us only non-IPO passes. I think this one sounds good: `PassBuilder::buildFunctionSimplificationPipeline()` (https://llvm.org/doxygen/classllvm_1_1PassBuilder.html#ac5f76510dbe4c1516309f99cd85d5ce9

What do you think about switching from `-O2` to this function? Looking at the implementation, it seems to run all the relevant passes.

regehr:
Yeah good idea, will look into it and change the code

regehr:
Hi Nuno- this is ready for review again.

optimizeModule() seems like it should be simpler, but I could not get any simpler version to work. The issue is that PassBuilder::registerFunctionAnalyses() does not register enough analyses to support buildFunctionSimplificationPipeline(). Instead of manually adding the missing stuff, I rather used existing library functions to get the rest of the stuff.

regehr:
if you want a test case for the single-file mode, I'm afraid it'll have to be a positive test case, since we can't rely on LLVM continuing to miscompile something, which would be required for a negative test case...

regehr:
@nunoplopes should I do anything else here?

nunoplopes:
Sorry for the delay; was sick this week and a large backlog built in the meantime.

LebedevRI:
> support running alive-tv with just a single file, in which case we run opt -O2 on it and verify refinment of that

@regehr i'm loving this change!
One thing perhaps is to maybe not hardcode `-O2`, but somehow let it be specified?

nunoplopes:
@LebedevRI That's equivalent to running `opt` like:
`~/llvm/build/bin/opt -load $HOME/alive2/build/tv/tv.so -tv -instcombine -tv -o /dev/null foo.ll`

A bit verbose, but a simple shell script should do the trick.

regehr:
Yeah, I didn 't see a way to make it easy to specify a custom opt pipeline here, I think we should just use opt for that.

what I'm trying to do here is to have a couple of very simple modes of using alive-tv that are compatible with compiler explorer, this canned "-O2" one, and then the more recent patch where the functions @src and @tgt are treated specially.

PR #290, opened 2020-02-13
Implement strlen using recfun

aqjune:
This PR implements strlen using Z3's recursive function.


smt/ctx.cpp
```diff
@@ -17,6 +17,7 @@ void context::init() {
   // Disable Z3's use of UFs for NaNs when converting FPs to BVs
   // They generate incorrect formulas when quantifiers are involved
   Z3_global_param_set("rewriter.hi_fp_unspecified", "true");
+  Z3_global_param_set("smt.recfun.depth", "32768");
```
nunoplopes: seems like a huge value. Why?

smt/expr.cpp
```diff
@@ -1579,6 +1609,16 @@ expr expr::mkLambda(const set<expr> &vars, const expr &val) {
   return Z3_mk_lambda_const(ctx(), vars.size(), vars_ast.get(), val());
 }
 
+expr expr::app(const vector<expr> &args) const {
+  C();
+  unique_ptr<Z3_ast[]> args_ast(new Z3_ast[args.size()]);
+  unsigned i = 0;
+  for (auto &v : args) {
+    args_ast[i++] = v();
```
nunoplopes: this needs to do `C2(v)` to check if it's valid.

smt/expr.cpp
```diff
@@ -1447,6 +1447,36 @@ expr expr::mkUF(const char *name, const vector<expr> &args, const expr &range) {
...
+  assert(args.size() == argdefs.size());
+  auto num_args = args.size();
+  vector<Z3_ast> z3_argdefs, z3_args;
+  vector<Z3_sort> z3_sorts;
+  z3_argdefs.reserve(num_args);
+  z3_args.reserve(num_args);
+  z3_sorts.reserve(num_args);
+
+  for (auto &arg : args) {
+    if (!arg.isValid())
```
nunoplopes: C2(arg)

ir/memory.cpp
```diff
@@ -7,6 +7,7 @@
 #include "ir/value.h"
 #include "util/compiler.h"
 #include <string>
+#include <iostream>
```
nunoplopes: debug leftover

ir/memory.h
```diff
@@ -287,6 +287,9 @@ class Memory {
 
   unsigned num_nonlocals() const;
 
+  const auto &get_local_block_val() const { return local_block_val; }
```
nunoplopes: these can't be exposed. These are internal details of the memory model

tools/transform.cpp
```diff
@@ -506,6 +506,16 @@ static uint64_t max_gep(const Instr &inst) {
...
     return 0;
   }
+  if (auto slen = dynamic_cast<const Strlen *>(&inst)) {
+    auto cop = dynamic_cast<ConversionOp *>(slen->getPointer());
+    auto glb = cop && cop->getOp() == ConversionOp::BitCast ?
+        dynamic_cast<GlobalVariable *>(&cop->getValue()) : nullptr;
+    if (glb)
+      max_mem_access = glb->size();
+    else
+      max_mem_access = 32760; //UINT64_MAX;
```
nunoplopes: why 32760?

tools/transform.cpp
```diff
@@ -506,6 +506,16 @@ static uint64_t max_gep(const Instr &inst) {
       max_mem_access = UINT64_MAX;
     return 0;
   }
+  if (auto slen = dynamic_cast<const Strlen *>(&inst)) {
+    auto cop = dynamic_cast<ConversionOp *>(slen->getPointer());
+    auto glb = cop && cop->getOp() == ConversionOp::BitCast ?
+        dynamic_cast<GlobalVariable *>(&cop->getValue()) : nullptr;
```
nunoplopes: maybe hoist this code into a function since it seems it will be helpful for other things beyond strlen.

tools/transform.cpp
```diff
@@ -564,7 +574,8 @@ static uint64_t get_access_size(const Type &ty) {
 static uint64_t get_access_size(const Instr &inst) {
   if (dynamic_cast<const Calloc*>(&inst) ||
       dynamic_cast<const Memcpy*>(&inst) ||
-      dynamic_cast<const Memset*>(&inst)) {
+      dynamic_cast<const Memset*>(&inst) ||
+      dynamic_cast<const Strlen*>(&inst)) {
```
nunoplopes: for now this is ok, but we need a subsequent patch to make it work with bits_byte > 8, since strlen is common.

smt/expr.cpp
```diff
@@ -1579,6 +1609,16 @@ expr expr::mkLambda(const set<expr> &vars, const expr &val) {
   return Z3_mk_lambda_const(ctx(), vars.size(), vars_ast.get(), val());
 }
 
+expr expr::app(const vector<expr> &args) const {
```
nunoplopes: this doesn't seem to be used anywhere?

smt/ctx.cpp
aqjune: This was set to the large number to minimize false positive.
If we can use static analysis, it can be the maximum length of string that is given to strlen.

ir/memory.h
aqjune: They are used at strlen() - do you prefer moving its encoding into Memory class (as memcpy/memset does)?

tools/transform.cpp
aqjune: fixed

tools/transform.cpp
aqjune: okay, understood

smt/expr.cpp
aqjune: It is used by Strlen::toSMT which does `decl.app(next_recfn_args)`

smt/ctx.cpp
aqjune: actually I am fine with setting it with smaller constants first and adding analysis for it.

nunoplopes: ok, I suggest getting rid of this change altogether. This parameter is the minimum initial unrolling; Z3 will bump it automatically.
This can be benchmarked afterward.

aqjune: I guess the number of unrolling can be smaller than this parameter; for example, setting this as 5 makes strlen.src.ll fail, but setting this as 6 makes it succeed.
Instead of 32768, I fixed it as 6 to make the strlen.src.ll test succeed - but it may not be sufficient for existing benchmarks.

smt/expr.cpp
```diff
@@ -1447,6 +1447,35 @@ expr expr::mkUF(const char *name, const vector<expr> &args, const expr &range) {
...
+    z3_sorts.emplace_back(arg.sort());
+  }
+  for (auto &argdef : argdefs) {
+    z3_argdefs.emplace_back(argdef());
+  }
+
+  auto decl = Z3_mk_rec_func_decl(ctx(), Z3_mk_string_symbol(ctx(), name),
+                                  num_args, z3_sorts.data(), range.sort());
+  expr f_app = Z3_mk_app(ctx(), decl, num_args, z3_args.data());
+  expr fbody = funbody(f_app)();
```
nunoplopes: missing C2(fbody)

ir/memory.h
nunoplopes: Not sure strlen() needs to go to memory.cpp. The exported load() is not sufficient?

smt/expr.cpp
```diff
@@ -1447,6 +1447,35 @@ expr expr::mkUF(const char *name, const vector<expr> &args, const expr &range) {
   return Z3_mk_app(ctx(), decl, num_args, z3_args.data());
 }
 
+expr expr::mkRecFnApp(const char *name, const vector<expr> &argdefs,
```
nunoplopes: maybe change the name to mkRecFn, since this doesn't create an application, just the fn decl

smt/ctx.cpp
aqjune: Please ignore my previous comment. 
I agree it seems to be the mininum one. I will remove the change as suggested.

ir/memory.h
aqjune: Nope, it isn't sufficient because the strlen recursive function needs to have arguments for the current memory.
The stored values at local_block_val and non_local_block_val are quantified, and they can be used by recursive functions by passing through arguments only, I think.

aqjune: Or we can declare Strlen as a friend class of Memory and hide these.

nunoplopes: The point is that we really don't want to expose the internals of the memory model.
Instead of quantifying the memory, you can use them directly and create one strlen() function for each call.
We can memoize them by having the memory produce a unique string that deterministically hashes the state. This is easy (just return the Z3 ids of the 2 block val arrays, for example). This will ensure that source/target blocks that are equivalent keep their formulas syntactically equal.

Issue #281, opened 2020-01-31
An error message with src/tgt with a loop


aqjune:
[loop-precond.zip](https://github.com/AliveToolkit/alive2/files/4140496/loop-precond.zip)
alive-tv says transformation is correct, not precondition is false.
If an identical pair is given (`alive-tv loop-precond.src.ll loop-precond.src.ll`), the output is `precondition is false`.
Would this be a bug, or a behavior that is allowed?

The precondition is returning SAT even though the loop is iterating constant times.

nunoplopes:
This is the expected behavior. There's nothing else we can do ATM.

PR #297, opened 2020-02-24
*in_memory() fails if domain is false, add loop-precond example

aqjune:
This patch resolves alive-tv/memory/calloc-overflow.src.ll failure which happens when there is an unreachable block, and adds the loop-precond example from #281


nunoplopes:
The attached test case doesn't crash with current master branch. I'm running the llvm test suite to check for regressions.

aqjune:
To me, it was `Transforms/GVN/PRE/2009-02-17-LoadPRECrash.ll`. Now it succeeds, thanks

PR #298, opened 2020-02-24
Implement strlen

aqjune:
Implements strlen by manually unrolling char comparison constant times & calling UF at last.
Each UF is signed with a unique id that is generated from local_block_val & non_local_block_val's Z3 ast id.


ir/instr.cpp
```diff
@@ -2227,6 +2227,88 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+  os << getName() << " = strlen " << *ptr;
+}
+
+
+// Given (bid, ofs0), encode an expression that returns the next offset ofs2
+// that has a null character.
+// Returns: <found?, ofs2>
+static pair<expr, expr>
+nextNull(const Memory &m, const expr &bid, const expr &ofs0, unsigned i,
+         const string &nextnull_uf) {
```
aqjune: This function can be reused at strcpy implementation.

ir/instr.cpp
```diff
@@ -2227,6 +2227,88 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+
+  const auto &m = s.getMemory();
+  Pointer p(m, eptr);
+
+  string nextnull_uf = "nextnull_" + m.unique_id();
+  expr bid = expr::mkVar("bid", p.get_bid());
+  expr ofs = expr::mkVar("ofs", p.get_offset());
+  s.addAxiom(
+    expr::mkForAll({ bid, ofs },
+                   expr::mkUF(nextnull_uf, {bid, ofs}, ofs).uge(ofs)));
```
nunoplopes: given that the definition is now fully unrolled, I think we don't need UFs.
If 2 strlen() calls are made, they will get the same expression anyway.

ir/instr.cpp
```diff
@@ -2227,6 +2227,88 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+
+  auto [found, ofs2] = nextNull(m, p.get_bid(), p.get_offset(), 0, nextnull_uf);
+
+  s.addUB(found);
```
nunoplopes: this should be addPre(), since we want to ignore executions where strlen isn't unrolled enough.

ir/instr.cpp
```diff
@@ -2227,6 +2227,88 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+    Pointer ptr2(m, bid, ofs2);
+    Byte b = m.load(ptr2);
+    // Approximate the situation by assuming that there is no null between
+    // ofs and ofs2
+    return { ptr2.inbounds(false, true) && !b.is_ptr() && !b.is_poison(false) &&
+             b.is_zero(), move(ofs2) };
+  }
+
+  Pointer ptr(m, bid, ofs);
+  Byte b = m.load(ptr);
```
nunoplopes: why not use the regular load()?
It also helps ensuring that all dereferenciability checks are done properly

ir/instr.cpp
aqjune: There is the case where the size of a block is unknown, e.g:
```
int f(i8* %ptr) {
return strlen(%ptr);
}
```
In this case, the strlen cannot be fully unrolled, so UF is needed

aqjune: I now see that this depends on whether found should be added to addPre or not

ir/instr.cpp
aqjune: What about a string that does not end with null character?
In the case, strlen should raise UB (https://en.cppreference.com/w/cpp/string/byte/strlen )

ir/instr.cpp
aqjune: The result of strlen with unroll_cnt=2 looks conceptually like this:

c1 = load(ptr)
c2 = load(ptr+1)
..
return expr::mkIf(c1 == '\0', 0, mkIf(c2 == '\0', 1, ..))

The second load should check dereferenceability only when c1 != '\0'.
The regular load unconditionally checks dereferenceability, leaving more UBs to the state than required.

ir/instr.cpp
nunoplopes: Using UFs will lead to false positives, since they don't check equality of memory. E.g. this fails verification:
```
%p = alloca
store %p
return strlen(@glb)
=>
return strlen(@glb)
```

These `strlen` calls will get different UFs, hence verification will fail.
There's no way to make this work beyond what was unrolled.

ir/instr.cpp
nunoplopes: Maybe load can get an extra parameter to fill in the UB stuff there rather than doing s.addUB()?

ir/instr.cpp
aqjune: I see. This seems to be a pretty hard problem..
I can discuss this issue with Gil as well. I will share the discussion result.

ir/instr.cpp
aqjune: That sounds good to me. I'll make a separate PR for this.

aqjune: It involves making Pointer::is_dereferenceable do that - do you think it's okay? I think the change seems good because it makes things less stateful.

BTW, the regular load() calls rewriteUndef at the end. I think rewriteUndef should not be called when load() is called from strlen. If unrolled, it will generate huge number of undef quantifiers. What do you think?
```
*p = undef
%a = strlen(p)
%b = strlen(p) // allow %b to be different from %a?
```

ir/instr.cpp
aqjune: Instead we can make it a random quantified variable (as freeze(undef) returns), which should be not greater than (memblock size - offset). What do you think?

nunoplopes: I don't think there's any way of completely getting rid of false-positives. Except for the solution I've proposed, which is to assume that the return value of strlen() is between 0 and max_unroll. i.e. ignore programs executions that go beyond this.
Should be fine since strings should be arbitrary, not constant. For constant strings we can constant fold either when converting LLVM IR to Alive, or here by checking if 'load(p[i]) == 0' is a constant bool.

aqjune:
Finally updated the patch!

PR #302, opened 2020-03-01
this makes three changes to alive-tv:

regehr:
- make sure we ask CMake to look for ZLIB (I thought this was changed,
  did we lose it somehow? anyway, this is mandatory now)

- optionally, send all alive-tv output to a file (this is for Compiler
  Explorer, which sometimes wants to directly exec() our program
  without going through a shell, so redirection doesn't work)

- only return a non-zero process exit code when there's actually an
  error; don't consider TV failure to be an error. again this is mainly
  to make Compiler Explorer happy: we don't want it to think our tool
  crashed when there's a TV failure

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,7 @@ if (CYGWIN)
   set(CMAKE_FIND_LIBRARY_SUFFIXES ".dll")
 endif()
 
+find_package(ZLIB)
 find_package(Z3 4.8.5 REQUIRED)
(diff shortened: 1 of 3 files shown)


regehr:
is there any chance I can get one of you @nunoplopes or @aqjune to implement the correct login in alive2test.py? this pull req simply makes the tests pass again. I tried a couple of things but am not sure I understand what's going on here, hopefully this is easy for someone who knows what they're doing

PR #304, opened 2020-03-02
Implement dominator tree

kevindevos:
This should correctly build dominator trees and print them as .dot files. The main use is to be able to lookup for the dominators or immediate dominator of a given node in O(1). I experimented with building an edge_iterator for it but that would require creating a second map with the relationship dominator->dominated nodes just for iterating over edges. At the moment i don't see any use for this other than quick immediate dominator lookup.

### DIFF
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -289,9 +289,10 @@ void CFG::printDot(ostream &os) const {
   os << "}\n";
 }
 
-
+// Relies on Alive's top_sort run during llvm2alive conversion in order to
+// traverse the cfg in reverse postorder to build dominators.
 void DomTree::buildDominators() {
-  // build predecessor and successor relationships
+  // build predecessors relationship
   for (const auto &[p, b, instr] : cfg) {
     (void)instr;
 
@@ -299,64 +300,50 @@ void DomTree::buildDominators() {
     auto &pred = doms.try_emplace(&p, p).first->second;
     b_node.dominator = nullptr;
     b_node.preds.push_back(&pred);
-    pred.children.push_back(&b_node);
   }
 
-  // JOHN B. KAM AND JEFFREY D. ULLMAN (1976).
-  // Adapted from Algorithm D in article "Global Data Flow Analysis 
-  // and Iterative Algorithms".
-  // This adapted version is almost identical to top_sort in sort.cpp.
-  // The resulting reverse post order works for building dominator trees
-  // even when CFG is cyclic.
-  // Sort a vector of DomTreeNode into reverse postorder.
-  vector<DomTreeNode*> sorted;
-  unordered_map<DomTreeNode*, bool> visited;
-  unsigned reverse_post_order = 1;
-  
-  function<void(DomTreeNode*)> visit = [&](DomTreeNode *node) {
-    if (visited[node])
-      return;
-    visited[node] = true;
-
-    for (auto child : node->children) {
-      visit(child);
-    }
-    // visit parent after visiting all children
-    sorted.push_back(node);
-    node->order = reverse_post_order;
-    // incr instead of decr because sorted is reversed later
-    ++reverse_post_order;
-  };
+  // initialization
+  unsigned i = f.getBBs().size();
+  for (auto &b : f.getBBs()) {
+    auto I = doms.find(b);
+    if (I == doms.end())
+      continue;
+    
+    I->second.order = --i;
+    I->second.dominator = nullptr;
+  }
 
   // assuming first bb is never moved
   auto &entry = doms.at(&f.getFirstBB());
-  visit(&entry);
-  reverse(sorted.begin(), sorted.end());
+  entry.dominator = &entry;
 
   // Cooper, Keith D.; Harvey, Timothy J.; and Kennedy, Ken (2001). 
   // A Simple, Fast Dominance Algorithm
   // http://www.cs.rice.edu/~keith/EMBED/dom.pdf
   // Makes multiple passes when CFG is cyclic to update incorrect initial
   // dominator guesses.
-  entry.dominator = &entry;
-
   DomTreeNode *new_idom;
   bool changed = true;
   while (changed) {
     changed = false;
-    for (auto b : sorted) {
-      if (b->preds.empty())
+    for (auto &b : f.getBBs()) {
+      auto I = doms.find(b);
+      if (I == doms.end())
+        continue;
+
+      auto &b_node = I->second;
+      if (b_node.preds.empty())
         continue;
       
-      new_idom = b->preds.front();
-      for (auto p : b->preds) {
+      new_idom = b_node.preds.front();
+      for (auto p : b_node.preds) {
         if (p->dominator != nullptr) {
           new_idom = intersect(p, new_idom);
         }
       }
 
-      if (b->dominator != new_idom) {
-        b->dominator = new_idom;
+      if (b_node.dominator != new_idom) {
+        b_node.dominator = new_idom;
         changed = true;
       }
     }
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -317,7 +317,7 @@ void DomTree::buildDominators() {
           ++srcIt;
           ++dstIt;
         } else {
-          ++dstIt;
+          break;
         }
       }
       dstDoms.insert(dstDoms.end(), &dst);
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -289,57 +289,94 @@ void CFG::printDot(ostream &os) const {
   os << "}\n";
 }
 
-// Requires that CFG::edge_iterator visits each predecessor of a node before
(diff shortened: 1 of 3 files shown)


LebedevRI:
I don't know the use-case, but can't the llvm domtree infra reused?

nunoplopes:
> I don't know the use-case, but can't the llvm domtree infra reused?

Not really because this will run after the conversion from LLVM IR -> Alive IR. The first use case for this is optimization of formula generation. This will become clearer in the next PRs hopefully.

LebedevRI:
> > I don't know the use-case, but can't the llvm domtree infra reused?
> 
> Not really because this will run after the conversion from LLVM IR -> Alive IR.

Yes, but i wasn't talking about simply reusing IR domtree,
but only reusing the datastructure (`llvm::DominatorTreeBase<>`?).

> The first use case for this is optimization of formula generation. This will become clearer in the next PRs hopefully.

Aha

nunoplopes:
> > > I don't know the use-case, but can't the llvm domtree infra reused?
> > 
> > 
> > Not really because this will run after the conversion from LLVM IR -> Alive IR.
> 
> Yes, but i wasn't talking about simply reusing IR domtree,
> but only reusing the datastructure (`llvm::DominatorTreeBase<>`?).

Oh, I see now what you meant.
I would like to stay as isolated from LLVM as possible since the verification engine is reusable for other tasks beyond just verifying LLVM. It can be used e.g. for MLIR in the future.
Also, it's good to stay away from LLVM because of possible bugs there. Our implementation would have to match LLVM's bugs for a miscompilation to get through.

ir/function.cpp
```diff
@@ -289,4 +289,66 @@ void CFG::printDot(ostream &os) const {
...
+};
+
+void DomTree::printDot(std::ostream &os) const {
+  os << "digraph {\n"
+        "\"" << bb_dot_name(f.getBBs()[0]->getName()) << "\" [shape=box];\n";
+
+  for (auto bb : f.getBBs()) {
+    if (dominators.at(bb).size() > 1) {
+      // print for before last element, last is itself
+      os << '"' << bb_dot_name(dominators.at(bb)[dominators.at(bb).size()-2]
```
nunoplopes: too many .at() lookups

ir/function.h
```diff
@@ -171,4 +171,17 @@ class CFG final {
   void printDot(std::ostream &os) const;
 };
 
+// Does not support loops
+class DomTree final {
+    Function& f;
+    CFG& cfg;
```
nunoplopes: CFG &cfg
+ everywhere else

ir/function.h
```diff
@@ -171,4 +171,17 @@ class CFG final {
...
+    // src -> vec<bb's that dominate src in order of their depth in the CFG>
+    std::unordered_map<const BasicBlock*,
+                      std::vector<const BasicBlock*>> dominators;
```
nunoplopes: including the BB in its own list is probably not very interesting.
Also, do we need to keep all dominators or only the immediate one?

ir/function.cpp
```diff
@@ -289,4 +289,66 @@ void CFG::printDot(ostream &os) const {
...
+      // find intersection between topologically sorted vectors of
+      // predecessor bb's
+      std::vector<const BasicBlock*>::const_iterator srcIt, dstIt;
+      srcIt = dominators[&src].begin();
+      dstIt = dstDoms.begin();
+      while (srcIt != dominators[&src].end() &&
+             dstIt != dstDoms.end()) {
+        if (*srcIt == *dstIt) {
+          dominators[&dst].push_back(*srcIt);
+          srcIt++;
```
nunoplopes: ++foo;

ir/function.cpp
```diff
@@ -289,4 +289,66 @@ void CFG::printDot(ostream &os) const {
   os << "}\n";
 }
 
+void DomTree::buildDominators() {
+  function<void(const BasicBlock*)> ensureExists;
+  ensureExists = [&](auto bb) {
+    if (dominators.find(bb) == dominators.end()) {
+      std::vector<const BasicBlock*> v{bb};
+      dominators.emplace(bb, move(v));
```
nunoplopes: this whole function is equivalent to: `dominators.try_emplace(bb)`
Though it doesn't seem really necessary as you can just have a variable like `auto &src_dom = dominators[&src]`

ir/function.cpp
```diff
@@ -289,4 +289,66 @@ void CFG::printDot(ostream &os) const {
   os << "}\n";
 }
 
+void DomTree::buildDominators() {
```
nunoplopes: what's the algorithm used here?

ir/function.h
kevindevos: The most important dominator is the immediate one which will always be the before last dominator in the vector and i don't think i need any other for the formulas algorithm. However we get all of them as part of the construction because to build dominators for x it needs to intersect the dominators of it's predecessors.

kevindevos: Including the BB itself is part of the algorithm for it to work, it could be removed at the end.

ir/function.cpp
kevindevos: The algorithm builds the dominators such that each dominator dominates itself and also contains the intersection of the dominators of all predecessors. It's just two formulas, it's better described here:
https://cs.stackexchange.com/questions/43105/dominator-tree-for-dag
This does require the CFG to be acyclic and that the edge iterator only visits one node after visiting all it's predecessors.

kevindevos:
The intersection code may have an issue, i need to review it in a moment

kevindevos:
There was no issue, everything should be good.

ir/function.cpp
```diff
@@ -289,57 +289,94 @@ void CFG::printDot(ostream &os) const {
...
+  DomTreeNode *new_idom;
+  bool changed = true;
+  while (changed) {
+    changed = false;
+    for (auto &b : f.getBBs()) {
+      auto &b_node = doms[b];
+      b_node->order = i;
+      if (i == bb_count-1)
+        b_node->dominator = b_node.get();
+      --i;
```
kevindevos: I have not been able to to get the index of a basic block in f.getBBs() from the loop above "for (const auto &[p, b, instr] : cfg) {" with pointer arithmetic, which would allow moving this initialization code to the first loop and improve efficiency for functions with loops.

ir/function.cpp
```diff
@@ -289,4 +289,96 @@ void CFG::printDot(ostream &os) const {
...
+
+      if (b_node->dominator != new_idom) {
+        b_node->dominator = new_idom;
+        changed = true;
+      }
+    }
+  }
 }
+
+DomTree::DomTreeNode* DomTree::intersect(DomTreeNode *b1, DomTreeNode *b2) {
```
nunoplopes: finger1, finger2. no need for the following 2 lines,

ir/function.cpp
```diff
@@ -289,4 +289,96 @@ void CFG::printDot(ostream &os) const {
...
+
+    b_node->preds.emplace_back(pred.get());
+  }
+
+  unsigned i = bb_count-1;
+  DomTreeNode *new_idom;
+  bool changed = true;
+  while (changed) {
+    changed = false;
+    for (auto &b : f.getBBs()) {
```
nunoplopes: is this the algorithm of figure 3?
If so, it says "for all nodes, b, in reverse postorder (except start node)".  Do we guarantee that order?

kevindevos: Yes, the reverse postorder is the topological order of f.getBBs(), both the article "Graph Algorithms with a Functional Flavour" from John Launchbury and a lecture from cornell (https://www.cs.cornell.edu/courses/cs2112/2012sp/lectures/lec24/lec24-12sp.html) say the same, and from what i tested i have not seen any differences.
The order attribute is then needed for intersect. The start node will be skipped as it has no predicates.
In 3hours when the llvm tests are finished i will push the new changes which includes a bugfix for loops.

ir/function.cpp
```diff
@@ -289,4 +289,100 @@ void CFG::printDot(ostream &os) const {
...
+// A Simple, Fast Dominance Algorithm
+// http://www.cs.rice.edu/~keith/EMBED/dom.pdf
+void DomTree::buildDominators() {
+  unsigned bb_count = f.getBBs().size();
+
+  // build predecessor relationships
+  for (const auto &[p, b, instr] : cfg) {
+    (void)instr;
+
+    auto &b_node = doms.try_emplace(&b, DomTreeNode(b)).first->second;
```
nunoplopes: DomTreeNode(b) -> b

ir/function.cpp
nunoplopes: So you rely on top_sort() being run before this function is called.
This should be documented and the the code inspected to guarantee that's the case.

kevindevos:
Ended up implementing the algorithm D in "Global Data Flow Analysis and Iterative Algorithms" by JOHN B. KAM AND JEFFREY D. ULLMAN for the reverse post order.

PR #308, opened 2020-03-08
Add -DTEST_NTHREADS, put alive-tv as check's dependency

aqjune:
Adds `-DTEST_NTHREADS=..` to cmake.
This is helpful for reducing timeout failures by giving small number of cores.

Also this adds alive-tv to `ninja check`'s dependency

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.8)
+include(ProcessorCount)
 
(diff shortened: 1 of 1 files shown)


aqjune:
I tried two alternatives (`ninja check -j4` and `NTHREADS=4 ninja check`) to work but they didn't.
If someone who is better at cmake/ninja can find the solution for it, I'm happy to follow it.

PR #309, opened 2020-03-08
Support noreturn

aqjune:


nunoplopes:
Returning undef doesn't seem correct. Either you add an unreachable instruction after the fn call or the fncall::toSMT() just sets UB = false (seems the easiest way to implement this).

aqjune:
Adding unreachable allows this transformation, which is undesirable:
```
BB:
  call f() noreturn
  unreachable
=>
BB:
  unreachable
```
I think ret undef is okay because it makes noreturn call well-defined. This PR has a few transformations which is (un)desirable with noreturn call.

nunoplopes:
Ok, I agree with you. But I really don't like adding that "return undef" instruction after the call.
First, because you can just do the return from within the fncall::toSMT and 2nd because we don't want any undef.

It's not clear to me what's the right way of implementing this.
Imagine something like:
```
call fn()
return undef
=>
call fn() noreturn
return 42
```

This transformation is incorrect, but I think it is accepted by this patch. This is because the return domain is confused with the domain of "hitting a noreturn function, but not returning".
Either we separate the 2 domains somehow, change how fn calls are recorded, or something else I can't think of right away.

aqjune:
I couldn't find simple & correct method to support noreturn, either.

I have another approach for fully supporting noreturn: we can introduce a special singleton value, say terminator. A terminator cannot refine any other value except itself. And then, we can introduce `ret terminator` after the noreturn call.
An advantage of using this special value is that it naturally encodes refinement between UB and noreturn (because UB can be optimized to a function call with noreturn; e.g. simplifying a branch with one of destination `unreachable` and another noreturn call). Also it wouldn't touch other sensitive algorithms, e.g. sinkDomain().
We can assume that terminator can only appear at src/tgt function's return value, so no extra bit is needed for registers/memory bytes/etc.

nunoplopes:
The problem is that there's no singleton value in SMT. Because the return value is fixed to a given type and there's no unique value you can use that the function can return.
So it seems we need to compute a separate expression for the no return but not UB cases?

aqjune:
I think we can mimick the bitvector representation of poison value. Conceptually, a poison value is a singleton because there is no distinction between integer poison, float poison, etc.
We can have terminator bit which is similar to poison bit for the returned value of src/tgt function.

aqjune:
Another concern is that when we want to support exception handling, adding another domain for exception seems too much; encoding it as a return value seems to simplify things.

nunoplopes:
> I think we can mimick the bitvector representation of poison value. Conceptually, a poison value is a singleton because there is no distinction between integer poison, float poison, etc.
> We can have terminator bit which is similar to poison bit for the returned value of src/tgt function.

Sure, that just means create an expression that represents the cases when the function returned in a no-return state. We don't need to pack it in StateValue though.

aqjune:
Well, I see.
Please give me a bit of time. The domain part seems complex, but it's worth for me to fully understand because it's an important part.

aqjune:
While investigating bugzilla reports, I realized that an arbitrary function may exit (even if it doesn't have `noreturn` attached); `noreturn` simply guarantees that it must exit.
Relevant bug: https://bugs.llvm.org/show_bug.cgi?id=27953
I'll update this PR to consider such case as well.

ir/instr.cpp
```diff
@@ -1247,6 +1249,12 @@ StateValue FnCall::toSMT(State &s) const {
   auto ret = s.addFnCall(fnName_mangled.str(), move(inputs), move(ptr_inputs),
                          out_types, !(flags & NoRead), !(flags & NoWrite),
                          flags & ArgMemOnly, move(returned_val));
+
+  if (flags & NoReturn) {
+    // TODO: Even if a function call doesn't have noreturn, it can possibly
+    // exit. Relevant bug: https://bugs.llvm.org/show_bug.cgi?id=27953
+    s.addNoReturn(expr(false));
```
aqjune: I'll expand this after #323 is reviewed; to do this, addFnCall should be fixed

ir/state.cpp
```diff
@@ -14,17 +14,18 @@ using namespace std;
 
 namespace IR {
 
-expr State::CurrentDomain::operator()() const {
-  return path && UB();
+expr State::CurrentDomain::operator()(bool noret) const {
+  return path && UB() && (noret ? noreturn() : expr(true));
```
aqjune: `noreturn()` will be false after a noreturn fn is called

tests/alive-tv/attrs/noreturn-2.srctgt.ll
```diff
@@ -0,0 +1,16 @@
+declare void @f() noreturn
+
+define void @src(i1 %cond) {
+  br i1 %cond, label %A, label %B
+A:
+  udiv i32 1, 0 ; cannot use unreachable due to a bug; it does not call addReturn
+  ret void
```
aqjune: I observed that this can be resolved by making Assume::toSMT() call State::addReturn(), but I guess it wouldn't be optimal for performance. What would be the best way to resolve this?

tools/transform.cpp
```diff
@@ -358,11 +358,21 @@ static void check_refinement(Errors &errs, Transform &t,
...
       [&](const Result &r) {
         err(r, [](ostream&, const Model&){},
             "Source is more defined than target");
       }},
+    { mk_fml(ub_a && dom_a && !dom_b),
+      [&](const Result &r) {
+        err(r, [](ostream&, const Model&){},
+            "Target may never return");
+      }},
+    { mk_fml(ub_a && !dom_a && dom_b),
```
nunoplopes: well, let's compress these 2 queries in one to reduce overhead.

nunoplopes:
I'm not convinced about the correctness of this patch. I think removing the noreturn cases from the return domain isn't ok.
e.g.:
```
call f() noreturn
=>
unreachable
```

Isn't the domain the same with your patch?
What about cases with loops where edges aren't reachable due to unrolling? The return domain may be the same when comparing a noreturn fn call with an unreachable BB.
I feel we need a return domain & a noreturn domain. Both of which can be a refinement, not equality.
Thoughts?

aqjune:
Hi, sorry for delay

> ```
> call f() noreturn
> =>
> unreachable
> ```
> 
> Isn't the domain the same with your patch?

In this example, returnDomain() will be the same whereas returnUB() wouldn't be the same. Maybe returnUB() is the one that you wanted to say.
check_refinement() also takes two kinds of domains - one that considers noreturn, and another that represents a domain without considering noreturn at the end of the block.

> What about cases with loops where edges aren't reachable due to unrolling? The return domain may be the same when comparing a noreturn fn call with an unreachable BB.

Actually this part was why I took pretty long time for responding to your question.
I have two questions about this:

(1) Why should we distinguish the two cases?
Is it because we want to separate impossible execution paths case due to insufficient loop unroll from the truly impossible execution path?
Then, when can the information be used?

(2) Wouldn't good unrolling heuristics always can make two domains different?
If our unrolling heuristic is good enough, unreachable blocks due to insufficient unrolling will finally be reachable.
There must be some nontrivial cases, such as:
```
for (int i = 0; i < 100; i++) {
  if (i >= 50) { printf("a"); }
  ..
}
```
To make the `printf` reachable, it should be unrolled 50 times, but I assumed that there must be some clever ideas (from good amount of previous works) for this.

aqjune:
I just realized that not separating it can resurrect https://github.com/AliveToolkit/alive2/issues/281 . :(
```
loop:
  br (i < 4), loop, exit
=>
loop:
  call f()
  br (i < 4) loop, exit
```
I'll update the patch to separate those two.

ir/instr.cpp
aqjune: For fixing addFnCall, I'll make a separate PR after this is merged

aqjune:
> > What about cases with loops where edges aren't reachable due to unrolling? The return domain may be the same when comparing a noreturn fn call with an unreachable BB.
> 
> Actually this part was why I took pretty long time for responding to your question.
> I have two questions about this:

Please ignore my questions here. They are resolved and relevant issues are fixed by following commits.

tools/transform.cpp
aqjune: Oh this wasn't intended. I thought I merged the two queries but it didn't. I'll update this

smt/exprs.h
```diff
@@ -25,6 +25,8 @@ class AndExpr {
   void add(const AndExpr &other);
   void del(const AndExpr &other);
   void reset();
+  const auto begin() const { return exprs.begin(); }
```
nunoplopes, line 28: these don't seem to be used?

ir/instr.cpp
nunoplopes: that's why you leave a condition there? you want to pass a non-deterministic boolean in the remaining cases?

ir/state.cpp
```diff
@@ -156,7 +157,7 @@ void State::addJump(const BasicBlock &dst0, expr &&cond) {
   auto &data = predecessor_data[dst][current_bb];
   data.second.add(memory, cond);
   data.first.UB.add(domain.UB(), cond);
-  data.first.path.add(move(cond));
+  data.first.path.add(move(cond) && domain.noreturn());
```
nunoplopes, line 160: This shouldn't be needed here. Path only tracks the paths constraints, i.e., the branching conditions

nunoplopes:
Ok, I've a suggestion on how to improve the encoding.
We need 2 domains: domain of the function, and domain for which the function returns. The latter is a subset of the former, obviously.
So every time there's a no return function call, we add the current domain to the function domain, and set UB=false.
For refinement, we just need to check that domain src -> domain tgt and domain return src -> domain return tgt.

In terms of implementation, this means not doing any change to addUB(), and change the addnoreturn.

ir/instr.cpp
aqjune: We should maintain a boolean var saying whether the function call exited or not, and mkAxiom() should encode a relation saying that if inputs are the same their exited-ness are equivalent.

aqjune:
A function without noreturn tag can still exit, so the encoding of UB should consider it:
```
  f()
  UB
=>
  UB
  f()
```
Even if f does not have noreturn tag, this transformation is invalid (https://bugs.llvm.org/show_bug.cgi?id=27953 ).

But.. this is a pretty big issue. Dealing with this in this single PR seems a bit intractable for review.

I'll assume that a function with explicit noreturn only exits first & update this PR. In this assumption, the suggested encoding will work.

aqjune:
Fixes Transforms/SimplifyCFG/noreturn-call.ll .

ir/instr.cpp
```diff
@@ -1331,6 +1333,12 @@ StateValue FnCall::toSMT(State &s) const {
   auto ret = s.addFnCall(fnName_mangled.str(), move(inputs), move(ptr_inputs),
                          out_types, !(flags & NoRead), !(flags & NoWrite),
                          flags & ArgMemOnly, move(returned_val));
+
+  if (flags & NoReturn) {
+    // TODO: Even if a function call doesn't have noreturn, it can possibly
```
nunoplopes, line 1338: can you open an issue in Alive2 with a test case that we miss miscompilation pls?

PR #311, opened 2020-03-10
Optimize strlen(p) when p is a constant char array

aqjune:


llvm_util/known_fns.cpp
```diff
@@ -48,6 +48,20 @@ known_call(llvm::CallInst &i, const llvm::TargetLibraryInfo &TLI,
...
+      auto BI = llvm::dyn_cast<llvm::BitCastInst>(i.getArgOperand(0));
+      auto G = llvm::dyn_cast<llvm::GlobalVariable>(BI->getOperand(0));
+      if (G && G->isConstant() && G->hasInitializer()) {
+        auto C = llvm::dyn_cast<llvm::ConstantDataArray>(G->getInitializer());
+        if (C && C->isCString()) {
+          auto sz = llvm::ConstantInt::get(i.getType(), C->getAsCString().size());
+          RETURN_KNOWN(make_unique<UnaryOp>(*ty, value_name(i),
+              *get_operand(sz, [](auto x){ return nullptr; },
+                               [](auto x){ return nullptr; }),
+              UnaryOp::Copy));
```
aqjune: Turns out that calling `get_operand` from `known_call` isn't pretty.
What do you think about moving `known_call` into llvm2alive?

nunoplopes: you don't need to create an llvm constant. There's an Alive function to create (Alive) integer constants.
Also, we should use hasDefinitiveInitializer. I just realized it's wrong in other places as well.

aqjune:
Reflected the comment

PR #315, opened 2020-03-14
cmake: search for additional subdirs for z3 include file

pranavk:
this should prevent the need to specify z3_include_dir

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,9 +42,6 @@ cmake -GNinja -DCMAKE_BUILD_TYPE=Release ..
 ninja
 ```
 
-If CMake cannot find the Z3 include directory (or finds the wrong one) pass
-the ``-DZ3_INCLUDE_DIR=/path/to/z3/include`` argument to CMake
-
 
 Building and Running Translation Validation
 --------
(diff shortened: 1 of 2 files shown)


pranavk:
i tested it on linux only. I think it should work fine on mac as well which was the reason @regehr introduced this readme line in the first place. let me know if it doesn't.

nunoplopes:
@regehr does this patch solve the issue you had?

regehr:
I still can't use z3 from a non-default location with or without this patch. I get the output below. 

Adding  NO_DEFAULT_PATH to the find_library command for z3 doesn't help.

This error happens even when I remove all other z3 installs from my system, so I don't know what is happening here.

```
-- Configuring done
CMake Warning at CMakeLists.txt:180 (add_library):
  Cannot generate a safe runtime search path for target alive2 because files
  in some directories may conflict with libraries in implicit directories:

    runtime library [libz3.so] in /usr/lib may be hidden by files in:
      /home/ce/z3-install/lib

  Some of these libraries may not be found correctly.


-- Generating done
```

nunoplopes:
Ok, if this doesn't help anyone let's not merge it.

PR #322, opened 2020-03-17
Add scaffolding for imprinting alive2 git version into executables. Fixes #321

### LINKED ISSUE #321: version indication needed
We should have a way for compiled alive2 binaries to report the git commit they're based on. AFAIK there's no particularly elegant or painless way to do this. I can look into it if nobody else has time.

LebedevRI:
This requires some git tag to be present in the repository.
I'm not sure github pr is smart enough to propagate the tag i added,
if not then the first commit in the repo should be marked as v0.0

```
$ ./alive
Usage: alive2 <options> <files.opt>
version v0.0-801-g3781c4d-dirty

<...>
```

```
$ ./alive-tv --help
OVERVIEW: Alive2 stand-alone translation validator:
version v0.0-801-g2b7965f-dirty
see alive-tv --version  for LLVM version info,

<...>
```

Fixes https://github.com/AliveToolkit/alive2/issues/321


CMakeLists.txt
```diff
@@ -32,6 +32,28 @@ if (CYGWIN)
...
+  COMMAND "${CMAKE_COMMAND}" -E echo_append "#define ALIVE_VERSION_MACRO " > "${CMAKE_BINARY_DIR}/version_gen.h.tmp"
+  COMMAND "${GIT_EXECUTABLE}" describe --tags --dirty >> "${CMAKE_BINARY_DIR}/version_gen.h.tmp"
+  WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
+  VERBATIM
+)
+add_custom_command(
+  DEPENDS "${CMAKE_BINARY_DIR}/version_gen.h.tmp"
+  OUTPUT "${CMAKE_BINARY_DIR}/version_gen.h"
+  COMMAND "${CMAKE_COMMAND}" -E copy_if_different "${CMAKE_BINARY_DIR}/version_gen.h.tmp" "${CMAKE_BINARY_DIR}/version_gen.h"
+  COMMAND "${CMAKE_COMMAND}" -E remove -f "${CMAKE_BINARY_DIR}/version_gen.h.tmp"
```
LebedevRI: This indirection is needed so that the version is auto-updated each time it can be updated,
without the need to re-running cmake.
This does of course mean that second `ninja` in `ninja && ninja` will not be a NOP.

CMakeLists.txt
```diff
@@ -32,6 +32,28 @@ if (CYGWIN)
   set(CMAKE_FIND_LIBRARY_SUFFIXES ".dll")
 endif()
 
+find_package(Git REQUIRED)
+add_custom_command(
+  OUTPUT "${CMAKE_BINARY_DIR}/version_gen.h.tmp"
+  COMMAND "${CMAKE_COMMAND}" -E echo_append "#define ALIVE_VERSION_MACRO " > "${CMAKE_BINARY_DIR}/version_gen.h.tmp"
+  COMMAND "${GIT_EXECUTABLE}" describe --tags --dirty >> "${CMAKE_BINARY_DIR}/version_gen.h.tmp"
```
LebedevRI: `git describe` adds newline at the end, and storing result into a macro seems like the least ugly solution to end up with a version string with no newline at the end

tools/alive-tv.cpp
```diff
@@ -314,8 +315,21 @@ static llvm::Function *findFunction(llvm::Module &M, const std::string FName) {
...
+  llvm::sys::PrintStackTraceOnErrorSignal(argv[0]);
+  llvm::PrettyStackTraceProgram X(argc, argv);
+  llvm::EnableDebugBuffering = true;
+  llvm::llvm_shutdown_obj llvm_shutdown; // Call llvm_shutdown() on exit.
+  llvm::LLVMContext Context;
+
+  std::string Usage =
+      R"EOF(Alive2 stand-alone translation validator:
+version )EOF";
+  Usage += alive_version;
```
LebedevRI: It is a really good idea to use the C string here, not a macro, so that version change only causes relinking but not recompilation.

tools/alive.cpp
```diff
@@ -19,19 +20,21 @@ using namespace std;
...
-    " -smt-to:x\t\tTimeout for SMT queries in ms\n"
-    " -max-mem:x\t\tMax memory consumption in MB (aprox)\n"
-    " -smt-verbose\t\tPrint all SMT queries\n"
-    " -skip-smt\t\tSkip all SMT queries\n"
-    " -disable-poison-input\tAssume input variables can never be poison\n"
-    " -disable-undef-input\tAssume input variables can never be undef\n"
-    " -h / --help\t\tShow this help\n";
+  cerr << "Usage: alive2 <options> <files.opt>\n"
+          "version "
+       << alive_version
```
LebedevRI: It is a really good idea to use the C string here, not a macro, so that version change only causes relinking but not recompilation.

util/version.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include "version_gen.h"
+
+#define xstr(s) str(s)
+#define str(s) #s
+
+namespace util {
+
+// clang-format off
+const char alive_version[] = {
+  xstr(ALIVE_VERSION_MACRO)
```
LebedevRI: since `git describe` adds newline at the end i can't just wrap it's output in cmake into `"`,
so we need this macro stringification ugliness..

regehr:
yikes, I forgot that alive does its own command line processing since it doesn't want LLVM as a dependency, argh!!!

my CMake is weak but this all LGTM.

thanks a lot Roman!

context for this request is that Nuno noticed that alive-tv on our CE was a bit out of date. this kind of observation would be much nicer if he (or whoever) could check the exact commit that is being used.

LebedevRI:
yes, i saw that disscussion.

LebedevRI:
@nunoplopes note that one needs to add some tag for this to work.
I'd suggest marking eed8d017ea40228b928fe77c03478f483baaa94e as v0.0

nunoplopes:
> @nunoplopes note that one needs to add some tag for this to work.
> I'd suggest marking [eed8d01](https://github.com/AliveToolkit/alive2/commit/eed8d017ea40228b928fe77c03478f483baaa94e) as v0.0

Done, thanks!

PR #327, opened 2020-03-21
Add -func param to alive-tv

aqjune:
This PR adds -func param to alive-tv. Given -func=f, alive-tv only compares function f from input. alive-tv can take multiple -func parameters.


tests/alive-tv/params/func-fail.src.ll
```diff
@@ -0,0 +1,13 @@
...
+define i32 @a() {
+  ret i32 0
+}
+
+define i32 @b() {
+  ret i32 1
+}
+
+define i32 @c() {
+  ret i32 2
```
nunoplopes: this test was supposed to fail. Is it failing? if yes, it needs the ERROR: thing

aqjune: Yes, I added the missing ERROR check.
BTW, I realized that the lit testing isn't checking absence of ERROR when dealing with src.ll/tgt.ll with multiple functions. Maybe I can look into this.

PR #328, opened 2020-03-21
alive2test.py should check absence of 'ERROR: ..' too

aqjune:
This also updates XFAIL to check stdout too.

Realized that we were missing several fails, including alive-tv/memory/lifetime-1.src.ll ; it was raising `Precondition is false` error.


aqjune:
Regarding the `precondition is false` error, I'll have a look

PR #336, opened 2020-03-29
Allow a global var to be constant in tgt (reproduces pr43616 again)

aqjune:
This PR allows a global var to be constant in tgt again.
This helps Alive2 reproduce https://bugs.llvm.org/show_bug.cgi?id=43616, which was found by Alive2 in the past but became unsupported at some point.


aqjune:
Will create a new PR with the tests only

PR #338, opened 2020-03-30
fix some seg faults

pranavk:
* extractvalue: don't access out of bounds item
* don't access uninitialized array type

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1108,7 +1108,7 @@ expr ExtractValue::getTypeConstraints(const Function &f) const {
     type = &ty->getChild(idx);
 
     c &= ty->numElements().ugt(idx);
-    if (++i == idxs.size())
+    if (++i == idxs.size() && !c.isFalse())
       c &= ty->getChild(idx) == getType();
   }
   return c;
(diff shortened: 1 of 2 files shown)


nunoplopes:
Can you please provide test cases?
These fixes don't look obvious to me at all.

pranavk:
For the extractvalue one:
```
Name: ev-out-of-bounds
%r = extractvalue [2 x i32] [i32 5, i32 10], 10
  =>
%r = 5
```

For the uninitialized one:
```
Name: insertvalue-1
%r = insertvalue [3 x i3] [i3 0, i3 0, i3 0], i3 0, 1
  =>
%r = [3 x i3] [i3 1, i3 0, i3 2]
```

If the second one doesn't crash, you can put an `assert(i < is_padding.size())` in `AggregateType::isPadding()`.

regehr:
add the tests to the PR :)

pranavk:
This would require some changes to the test infra. Eg: I can add extractvalue test case above but for that I need to change "Doesn't type check" ==> "ERROR: Doesn't type check" or change lit.py a bit. First option may break things with Compiler Explorer, etc. So I guess I will change lit.py a bit to support these typechecking failure tests.

nunoplopes:
Thanks! I've fixed those crashes here: https://github.com/AliveToolkit/alive2/commit/4d159a7a68a244039fe20cac36a14774c35e7872
Your first fix didn't work with non-constant vectors.

PR #339, opened 2020-03-30
add support for insertvalue

pranavk:
both alive IR and LLVM IR support


nunoplopes:
Did you run LLVM's test suite? What are the new test failures (if any)?

pranavk:
Not yet. Will report back.

ir/instr.cpp
```diff
@@ -1123,6 +1123,102 @@ unique_ptr<Instr> ExtractValue::dup(const string &suffix) const {
...
+  for (auto idx : idxs) {
+    auto ty = type->getAsAggregateType();
+    if (!ty) {
+      c = false;
+      break;
+    }
+    type = &ty->getChild(idx);
+
+    c &= ty->numElements().ugt(idx);
+    if (++i == idxs.size() && !c.isFalse())
```
nunoplopes: please don't use `c.isFalse()` as it doesn't work in general.
Use the same fix I did for extractvalue: `idx < ty->numElementsConst()`

ir/type.cpp
```diff
@@ -737,6 +737,30 @@ StateValue AggregateType::extract(const StateValue &val, unsigned index) const {
...
+  expr idx_v = expr::mkUInt(total_value, bits());
+  expr fill_v = expr::mkUInt(0, bits() - bw_elt);
+  expr mask_v = ~expr::mkInt(-1, bw_elt).concat(fill_v).lshr(idx_v);
+  expr nv_shifted = children[index]->toBV(elt).value.concat(fill_v).lshr(idx_v);
+
+  unsigned bw_np_elt = children[index]->np_bits();
+  expr idx_np = expr::mkUInt(total_np, np_bits());
+  expr fill_np = expr::mkUInt(0, np_bits() - bw_np_elt);
+  expr mask_np = ~expr::mkInt(-1, bw_np_elt).concat(fill_np).lshr(idx_np);
+  expr np_shifted = children[index]->toBV(elt).non_poison.concat(fill_np).lshr(idx_np);
```
nunoplopes: this line seems over 80 chars

ir/instr.cpp
```diff
@@ -1123,6 +1123,102 @@ unique_ptr<Instr> ExtractValue::dup(const string &suffix) const {
...
+  os << getName() << " = insertvalue " << *val << ", " << *elt;
+  for (auto idx : idxs) {
+    os << ", " << idx;
+  }
+}
+
+StateValue InsertValue::toSMT(State &s) const {
+  auto sv = s[*val];
+
+  Type *type = &val->getType();
```
nunoplopes: this seems like a lot of repeated code here.
Since indexes are constant, I suggest you extract elements & construct a new value with aggregateVals. You only need to go deep on the indexes path; the remaining ones you keep top-level. i.e. extract idx = 0, append to vector, extract idx=1, but since it's an update, go deeper, etc. then aggregate back.
With a scheme like this there's zero need for new bit-manipulating code, which we don't really want, as it limits opportunities to change representation. And code below is not really possible in this file. Internals like bit-level representation aren't supposed to be exposed in the semantics (this file).

tools/alive_parser.cpp
```diff
@@ -822,6 +822,23 @@ static unique_ptr<Instr> parse_extractvalue(string_view name) {
   return instr;
 }
 
+static unique_ptr<Instr> parse_insertvalue(string_view name) {
+  auto &type = parse_type();
+  auto &val = parse_operand(type);
+  parse_comma();
+  auto &elt_ty = parse_type();
+  auto &elt = parse_operand(elt_ty);
+  auto instr = make_unique<InsertValue>(get_sym_type(), string(name), val, elt);
```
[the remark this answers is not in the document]
pranavk: You are right. Used the template from extractvalue and forgot to change it.

ir/instr.cpp
pranavk: true. Following your suggestion, made it much simpler. Thanks. I got inspired by logic in InsertElement::update earlier.

pranavk:
> Did you run LLVM's test suite? What are the new test failures (if any)?

Yes, I ran it on LLVM test suite. New test that fails after adding this insertvalue support is this:
`test/Transforms/InstCombine/with_overflow.ll`

After minimizing the test case, we found out that the actual bug doesn't really depend on insertvalue. Eg: Below is the minimized/tweaked src file which when optimized with LLVM's instcombine results in a code which is not verified by alive.

```
define i8 @sadd_always_overflow(i8 %x) nounwind {
  %c = icmp sgt i8 %x, 100
  %y = select i1 %c, i8 %x, i8 100
  %a = call { i8, i1 } @llvm.sadd.with.overflow.i8(i8 %y, i8 28)
  %r = extractvalue {i8, i1} %a, 0
  ret i8 %r
}
declare { i8, i1 } @llvm.sadd.with.overflow.i8(i8, i8) nounwind readnone
```

Here's the resulting alive output
```
define i8 @sadd_always_overflow(i8 %x) {
%0:
  %c = icmp sgt i8 %x, 100
  %y = select i1 %c, i8 %x, i8 100
  %a = sadd_overflow {i8, i1} %y, 28
  %r = extractvalue {i8, i1} %a, 0
  ret i8 %r
}
=>
define i8 @sadd_always_overflow(i8 %x) {
%0:
  %c = icmp sgt i8 %x, 100
  %y = select i1 %c, i8 %x, i8 100
  %a = add nuw i8 %y, 28
  ret i8 %a
}
ERROR: Target is more poisonous than source

Example:
i8 %x = undef

Source:
i1 %c = undef
i8 %y = undef
{i8, i1} %a = { undef, undef }
i8 %r = undef

Target:
i1 %c = #x1 (1)
i8 %y = #xe4 (228, -28)
i8 %a = poison
Source value: undef
Target value: poison
```

nunoplopes:
Nice. I don't think we have reported that bug.
Though be aware that bug would go away if undef didn't exist. I haven't been reporting these bugs for now.

pranavk:
oops, i just noticed that some of the tests i added fails typechecking phase. Let me fix them

ir/instr.cpp
```diff
@@ -1125,6 +1125,89 @@ unique_ptr<Instr> ExtractValue::dup(const string &suffix) const {
...
+    } else {
+      vals.emplace_back(v);
+    }
+  }
+
+  return ty->aggregateVals(vals);
+}
+
+StateValue InsertValue::toSMT(State &s) const {
+  auto sv = s[*val];
```
nunoplopes, line 1171: use & in these

ir/instr.cpp
```diff
@@ -1125,6 +1125,89 @@ unique_ptr<Instr> ExtractValue::dup(const string &suffix) const {
...
+  auto elem = s[*elt];
+
+  Type *type = &val->getType();
+  vector<unsigned> idxs_reverse = idxs;
```
nunoplopes, line 1175: vector has a constructor that can take iterators. so you can do vector(idxs.rbegin(), rend())

ir/instr.cpp
```diff
@@ -1125,6 +1125,89 @@ unique_ptr<Instr> ExtractValue::dup(const string &suffix) const {
...
+}
+
+void InsertValue::print(ostream &os) const {
+  os << getName() << " = insertvalue " << *val << ", " << *elt;
+  for (auto idx : idxs) {
+    os << ", " << idx;
+  }
+}
+
+StateValue update_repack(Type *type,
```
nunoplopes, line 1148: static

ir/instr.cpp
```diff
@@ -1125,6 +1125,89 @@ unique_ptr<Instr> ExtractValue::dup(const string &suffix) const {
...
+                         const StateValue &val,
+                         const StateValue &elem,
+                         vector<unsigned> &indices) {
+  auto ty = type->getAsAggregateType();
+  unsigned cur_idx = indices.back();
+  indices.pop_back();
+  vector<StateValue> vals;
+  for (unsigned i = 0; i < ty->numElementsConst(); ++i) {
```
nunoplopes, line 1156: i= 0, e = ty->numeleme..(); i < e

pranavk:
okay. all should be good now. CI passes again and i addressed your comments above.

aqjune:
Nit: I think the tests should not be at `constexpr` dir. They are not checking whether encoding of `insertvalue` is correctly done when it is in a constant expression. `insertvalue` constexpr is something like this: https://godbolt.org/z/cePn2c .
If you have a time, could you make a PR so they are moved out from this dir & `insertvalue` constexpr tests are added too?

regehr:
@pranavk if you do this, you should also consider using the .srctgt.ll extension which halves the number of test files

PR #341, opened 2020-04-01
Let memcpy -> load/store i8 raise memory mismatch

aqjune:
.. because load i8 yields poison if it was a pointer byte.
This implies that we don't have 'unsigned char' type in the twin memory model right now.
To recap: making i8 as unsigned char again causes all arithmetic optimizations on i8  unsound/controversial again, so this was our design choice.


aqjune:
With undef disabled, no difference in failures on my machine (it is 72)

PR #347, opened 2020-04-04
Support vector load/store with element size not a multiple of 8

### LINKED ISSUE #277: Vector load/store broken with element size not a multiple of 8
Encoding of the following is broken:
```
define i32 @fn(<4 x i4>* %p ) {
    store <4 x i4> <i4 3, i4 5, i4 2, i4 3>, <4 x i4>* %p
    ret i32 0
}
```

aqjune:
This resolves #277 by making int vectors be encoded as primitive values at valueToBytes / bytesToValue.
For structs (e.g. {i1, ..}), this doesn't work, because each element occupy its own byte.
Fixes one failure (Transforms/SROA/vector-promotion-different-size.ll).

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -273,7 +273,7 @@ static vector<Byte> valueToBytes(const StateValue &val, const Type &fromType,
     for (unsigned i = 0; i < bytesize; ++i)
       bytes.emplace_back(p, i, val.non_poison);
   } else {
-    assert(!fromType.isAggregateType() || isIntVector(fromType));
+    assert(!fromType.isAggregateType() || isNonPtrVector(fromType));
     StateValue bvval = fromType.toInt(*s, val);
     unsigned bitsize = bvval.bits();
     unsigned bytesize = divide_up(bitsize, bits_byte);
@@ -326,7 +326,7 @@ static StateValue bytesToValue(const Memory &m, const vector<Byte> &bytes,
              move(non_poison) };
 
   } else {
-    assert(!toType.isAggregateType() || isIntVector(toType));
+    assert(!toType.isAggregateType() || isNonPtrVector(toType));
     auto bitsize = toType.bits();
     assert(divide_up(bitsize, bits_byte) == bytes.size());
 
@@ -1457,7 +1457,7 @@ unsigned Memory::getStoreByteSize(const Type &ty) {
     return divide_up(bits_program_pointer, 8);
 
   auto aty = ty.getAsAggregateType();
-  if (aty && !isIntVector(ty)) {
+  if (aty && !isNonPtrVector(ty)) {
     unsigned sz = 0;
     for (unsigned i = 0; i < aty->numElementsConst(); ++i)
       sz += getStoreByteSize(aty->getChild(i));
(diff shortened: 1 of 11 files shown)


ir/memory.cpp
```diff
@@ -1512,7 +1516,8 @@ Memory::load(const expr &p, const Type &type, unsigned align) {
   auto ubs = ptr.is_dereferenceable(bytecount, align, false);
 
   StateValue ret;
-  if (auto aty = type.getAsAggregateType()) {
+  auto aty = type.getAsAggregateType();
+  if (aty && !isVectorWithNonByteInts(type)) {
```
nunoplopes, line 1520: can we use the code for all vectors (not just non-byte aligned)?
This would allow to widen all vector load/stores.

nunoplopes: I guess non-ptr vectors only.

ir/type.cpp
```diff
@@ -856,7 +856,27 @@ expr AggregateType::fromInt(expr v) const {
...
+  for (unsigned i = 0; i < elements; ++i) {
+    unsigned idx = little_endian ? i : elements - 1 - i;
+    auto *child = children[idx];
+    if (child->bits() == 0)
+      continue;
+
+    auto bits2 = bits + child->bits();
+    auto np_bits2 = np_bits + child->np_bits();
+    child_vals[idx]
+      = children[i]->fromInt({ v.value.extract(bits2 - 1, bits),
```
nunoplopes, line 874: AggregateType has an extract function. That should work I think, no?

aqjune: extract calls fromBV, so may not work in general.

ir/memory.cpp
aqjune: Updated

ir/memory.cpp
```diff
@@ -1454,7 +1456,8 @@ unsigned Memory::getStoreByteSize(const Type &ty) {
   if (ty.isPtrType())
     return divide_up(bits_program_pointer, 8);
 
-  if (auto aty = ty.getAsAggregateType()) {
+  auto aty = ty.getAsAggregateType();
+  if (aty && !isIntVector(ty)) {
```
nunoplopes, line 1460: why only int? we only distinguish between ptr and non-ptr in memory.

ir/type.cpp
nunoplopes: right. then maybe add a flag to extract to pick either fromBV/fromInt.
I really want to keep the number of functions that know about encoding minimal, so it's easier to experiment with different encodings. Also, it reduces opportunities for bugs to creep in.

ir/memory.cpp
aqjune: Rename `isIntVector` to `isNonPtrVector` and expand its coverage too

ir/type.cpp
aqjune: Updated

aqjune:
A few tests failing, investigating

nunoplopes:
> A few tests failing, investigating

Ah, didn't see this message in time. Please submit a new PR.
If this is not a quick fix, I can just revert this.

aqjune:
Having a look, this seems to happen previously too, maybe I can make a fix

aqjune:
One regression is fixed by #348 
Another new failure on my machine is Transforms/InstCombine/load.ll , which transforms load/store of <4 x i8> to load/store of i32, so it should fail.

Issue #345, opened 2020-04-04
Refinement from undef ptr to undef int


aqjune:
https://web.ist.utl.pt/~nuno.lopes/alive2/index.php?hash=b636d1f9d0bd9caf&test=Transforms%2FMemCpyOpt%2Ffca2memcpy.ll

In src,

```
store {*, i8, i24, i32} undef, * %dst, align 8
```

stores undef pointer at %dst.
After transformation,

```
memset * (bitcast %dst) align 8, i8 undef, i64 16
```

This problem can be resolved by adding a refinement between pointer byte to null and integer byte.
More specifically, we can say that a pointer byte ((nullblock, i), ofs) is refined by an integer ((i >> ofs) & 0xFF).

nunoplopes:
We decided this is not really correct, since a subsequent load would read an int like a pointer, yielding poison rather than undef.

Issue #236, opened 2019-12-18
nullptr byte and zero byte should be equivalent 


aqjune:
Transforms/MemCpyOpt/fca2memcpy.ll 's `@destroysrc()` converts
```
store {*, i8, i24, i32} { null, 0, poison, 0 }, * %src, align 8
```
into memset 0.

PR #349, opened 2020-04-06
Support store ptr undef -> memset i8 undef (#345)

aqjune:
This resolves issue #345.

Transformations of interest:
```
  //   store ty* null  <-> memset i8 0
  //   store ty* undef  -> memset i8 undef
```

Only extracting the last bits_byte bits from pointer offset & comparing it seemed enough.
Running unit tests..

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -811,11 +811,10 @@ expr Pointer::blockValRefined(const Pointer &other) const {
                 val.ptrByteoffset() == val2.ptrByteoffset() &&
                 val.ptr().refined(val2.ptr());
   }
-  // allow null ptr <-> zero
   // If ptr points to a null block with ofs, its integer address is compared.
-  // Relevant transformations:
-  //   store ty* null  <-> memset i8 0
-  //   store ty* undef  -> memset i8 undef
+  // This explains two transformations:
+  //   store ty* null  -> memset i8 0
+  //   store ty* undef -> store ty* (nullblk, undef) -> memset i8 undef
   expr intval;
   if (val.isValid()) {
     expr ofs = val.ptr().getOffset();
@@ -825,15 +824,13 @@ expr Pointer::blockValRefined(const Pointer &other) const {
                 .zextOrTrunc(bits_byte);
   }
   // val is (null, ofs), val2 is ofs
-  expr refine_ptrint = val.ptr().isNull(false) && !val2.isPoison(false) &&
+  expr refine_ptrint = val.ptr().isNull(false) &&
+                       val2.nonptrNonpoison() == 0 &&
                        intval == val2.nonptrValue();
-  // val is 0, val2 is nullptr
-  expr refine_intptr =
-    !val2.isPoison(false) && val.nonptrValue() == 0 && val2.ptr().isNull();
   return val.isPoison() ||
-         expr::mkIf(is_ptr,
-            expr::mkIf(is_ptr2, ptr_cnstr, refine_ptrint),
-            expr::mkIf(is_ptr2, refine_intptr, int_cnstr));
+         expr::mkIf(is_ptr == is_ptr2,
+                    expr::mkIf(is_ptr, ptr_cnstr, int_cnstr),
+                    is_ptr && move(refine_ptrint));
 }
 
 expr Pointer::blockRefined(const Pointer &other) const {
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -815,11 +815,28 @@ expr Pointer::blockValRefined(const Pointer &other) const {
                 val.ptrByteoffset() == val2.ptrByteoffset() &&
                 val.ptr().refined(val2.ptr());
   }
+  // allow null ptr <-> zero
+  // If ptr points to a null block with ofs, its integer address is compared.
+  // Relevant transformations:
+  //   store ty* null  <-> memset i8 0
+  //   store ty* undef  -> memset i8 undef
+  expr val_i8 =
+    expr::mkIf(is_ptr,
+      (val.ptr().getOffset().lshr(val.ptrByteoffset() * bits_byte))
+          .zextOrTrunc(bits_byte),
+      val.nonptrValue());
+  expr val2_i8 =
+    expr::mkIf(is_ptr2,
+      (val2.ptr().getOffset().lshr(val.ptrByteoffset() * bits_byte))
+          .zextOrTrunc(bits_byte),
+      val2.nonptrValue());
+  expr null_punning = (!is_ptr2 || val2.ptr().isNull(false)) &&
+                      (!is_ptr || val.ptr().isNull(false)) &&
+                      !val2.isPoison() && val_i8 == val2_i8;
   return val.isPoison() ||
          expr::mkIf(is_ptr == is_ptr2,
                     expr::mkIf(is_ptr, ptr_cnstr, int_cnstr),
-                    // allow null ptr <-> zero
-                    val.isZero() && !val2.isPoison() && val2.isZero());
+                    move(null_punning));
 }
 
 expr Pointer::blockRefined(const Pointer &other) const {
@@ -906,10 +923,13 @@ Pointer Pointer::mkNullPointer(const Memory &m) {
   return { m, 0, false };
 }
 
-expr Pointer::isNull() const {
+expr Pointer::isNull(bool check_offset) const {
   if (!has_null_block)
     return false;
-  return *this == mkNullPointer(m);
+  auto np = mkNullPointer(m);
+  if (!check_offset)
+    return this->getBid() == np.getBid();
+  return *this == np;
 }
 
 expr Pointer::isNonZero() const {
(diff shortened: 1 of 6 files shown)


aqjune:
Resolves one failure (Transforms/MemCpyOpt/fca2memcpy.ll)
Unit tests timeout: 997 -> 1010

ir/memory.cpp
```diff
@@ -814,11 +810,25 @@ expr Pointer::block_val_refined(const Pointer &other) const {
...
+  //   store ty* undef  -> memset i8 undef
+  expr val_i8 = expr::mkIf(is_ptr,
+                           val.ptr().get_offset().zextOrTrunc(bits_byte),
+                           val.nonptr_value());
+  expr val2_i8 = expr::mkIf(is_ptr2,
+                            val2.ptr().get_offset().zextOrTrunc(bits_byte),
+                            val2.nonptr_value());
+  expr null_punning = (!is_ptr2 || val2.ptr().isNull(false)) &&
+                      (!is_ptr || val.ptr().isNull(false)) &&
+                      !val2.is_poison() && val_i8 == val2_i8;
```
nunoplopes, line 827: When this code is used, either is_ptr is true or is_ptr2 is true, but not both.
So I think we can make this constraint simpler, e.g.:
```
!val2.is_poison() &&
  mkif(is_ptr,
       val.ptr().isNull(false) && val.ptr().get_offset().zextOrTrunc(bits_byte) == val2.nonptr_value(),
       val2.ptr().isNull(false) && val.nonptr_value() == val2.ptr().get_offset().zextOrTrunc(bits_byte))
```

Though this feels a bit strange to me. Why are we comparing an 8-bit value with a 64-bit offset truncated to 8 bits? Seems wrong. And this is not detecting undef at all, right?
This will allow something like (undesired):
```
%p = gep i8, null, 42
store %p, %q
=>
store 42, %q
store 42, %q+1
store 42, %q+2
store 42, %q+...
```

aqjune: Detecting undef seemed hard because the generated formula looked like this:
(select (store .. (store init_mem undef) undef) |idx|)
This cannot be constant folded because idx isn't constant.

Regarding the truncation, it was my suggested approximation. Otherwise we need to shift the corresponding bits, which seemed too expensive for resolving the single case from undef.

aqjune: To supplement: the added refinement here is nullptr -> int, and undefptr -> nullptr is already allowed, so undefptr -> nullptr -> int is now allowed.

nunoplopes: right, but we need to make sure we don't introduce false negatives.
Let's think a bit more in a better solution. I don't think the solution you're proposing is a good tradeoff. Better 1 false positive than introducing false-negatives IMHO.

aqjune:
Rebased & fixed it to extract Byteoffset bits of the pointer offset and compare, rather than just comparing the last bits.

Timeouts: 1065 -> 1083

As a follow-up, I can prepare a patch that simplifies the refinement in certain input.
LLVM unit tests say that this refinement is helpful only when there is a pointer store in src & memset undef in tgt, so we can simply disable this in all other cases.

ir/memory.cpp
```diff
@@ -815,11 +811,29 @@ expr Pointer::blockValRefined(const Pointer &other) const {
                 val.ptrByteoffset() == val2.ptrByteoffset() &&
                 val.ptr().refined(val2.ptr());
   }
+  // allow null ptr <-> zero
+  // If ptr points to a null block with ofs, its integer address is compared.
+  // Relevant transformations:
+  //   store ty* null  <-> memset i8 0
+  //   store ty* undef  -> memset i8 undef
+  expr intval;
+  if (val.isValid()) {
```
nunoplopes, line 820: isvalid shouldn't be needed

ir/memory.cpp
```diff
@@ -815,11 +811,29 @@ expr Pointer::blockValRefined(const Pointer &other) const {
                 val.ptrByteoffset() == val2.ptrByteoffset() &&
                 val.ptr().refined(val2.ptr());
   }
+  // allow null ptr <-> zero
+  // If ptr points to a null block with ofs, its integer address is compared.
+  // Relevant transformations:
+  //   store ty* null  <-> memset i8 0
```
nunoplopes, line 817: Although we allow reading 0 null, we currently don't allow reading null as 0.
Was there any transformation that required that?  If so, we need to change `bytesToValue` as well..

ir/memory.cpp
```diff
@@ -815,11 +811,29 @@ expr Pointer::blockValRefined(const Pointer &other) const {
...
+                       intval == val2.nonptrValue();
+  // val is 0, val2 is nullptr
+  expr refine_intptr =
+    !val2.isPoison(false) && val.nonptrValue() == 0 && val2.ptr().isNull();
   return val.isPoison() ||
-         expr::mkIf(is_ptr == is_ptr2,
-                    expr::mkIf(is_ptr, ptr_cnstr, int_cnstr),
-                    // allow null ptr <-> zero
-                    val.isZero() && !val2.isPoison() && val2.isZero());
+         expr::mkIf(is_ptr,
```
nunoplopes, line 834: I feel the poison constraint for val2 needs to be hoisted. Right now it appears 4x times. Not good.
Also, don't use `isPoison` as it brings a big formula for no reason.

ir/memory.cpp
```diff
@@ -815,11 +811,29 @@ expr Pointer::blockValRefined(const Pointer &other) const {
...
+  // If ptr points to a null block with ofs, its integer address is compared.
+  // Relevant transformations:
+  //   store ty* null  <-> memset i8 0
+  //   store ty* undef  -> memset i8 undef
+  expr intval;
+  if (val.isValid()) {
+    expr ofs = val.ptr().getOffset();
+    expr shift_amount = val.ptrByteoffset() *
+                        expr::mkUInt(bits_byte, val.ptrByteoffset().bits());
+    intval = ofs.lshr(shift_amount.zextOrTrunc(ofs.bits()))
```
nunoplopes, line 824: I don't really understand this.
You have `intval = offset >> byteOffset * bits_byte`

This formula has no meaning. It's mixing ptr offset & the byte offset that can't really be mixed. What's this trying to achieve?

aqjune: The goal here is to interpret a pointer byte (0, ofs) as a 64-bit integer of ofs.
When loading a byte, the corresponding 8 bits should be extracted, so it is shifted by `byteOffset * bits_byte` and extracted.

aqjune:
Rebased & addressed comments (f4bb799 is the only added commit)

ir/memory.cpp
aqjune: I ran LLVM unit tests to check whether refinement from 0 to nullptr is needed, and it did not raise any new failure.
The motivation was also about nullptr -> 0 (https://github.com/AliveToolkit/alive2/issues/236 ), so I removed the refinement.

ir/memory.cpp
aqjune: Updated to use nonptrNonpoison instead.
Is hoisting the poison check for val2 still needed? It is simpler, compared to the before. And to me hoisting it seemed nontrivial...

ir/memory.cpp
nunoplopes: This doesn't seem correct. See this example:
```
p = malloc
p' = gep p, 0x700000
store i8* p, i8** q
=>
store i8* p, i8** q
store 0, (bitcast p to i8)
```

Your code will shift the offset by at least one. So the shifted offset is 0, and thus refinement will hold for the above program. Which it shouldn't.

aqjune: I think in that case p isn't null block, so refinement shouldn't hold.
Here is the program that I tested with - [zero-null2.zip](https://github.com/AliveToolkit/alive2/files/4640248/zero-null2.zip)

nunoplopes: ok, then this:
```
p' = gep null, 0x700...000
store i8* p, i8** q
=>
store i8* null, i8** q
store 0, (bitcast p to i8)
```

I really don't understand the direction you are taking with this patch. It compares 2 completely different types: internal Alive2 pointer representation with an `i8` LLVM value. Doesn't type check. Can't see why it would ever be correct.

aqjune: In that case refinement also doesn't hold, if I understand the example correctly: https://gist.github.com/aqjune/d3d76604f28a71f8e4af565856b83eac

There are two optimizations that fails if we allow refinement on bytes of a same type only:
`store * null, %ptr   -> memset i8 0` (#236)
`store * undef, %ptr -> memset i8 undef` (#345)
I was trying to give a relation between a pointer byte and i8 to explain these. The first one was addressed by #237, the second one is on the fly.
An approach that I suggest is that we can treat a pointer on a null block as something special so it is compatible with an integer. I think this is a natural expansion of the previous patch (#237) as well as supports these two optimizations.

aqjune:
Determined to leave the optimization as unsupported, as discussed

PR #350, opened 2020-04-07
AggregateType maintains padding elements, and *_overflow should consider this too

aqjune:
`tests/alive-tv/umul_overflow.srctgt.ll` raises type checking failure, and the reason is that in Alive2 umul_overflow's return type is not {i32, i1}, but {i32, i1, i24} because llvm2alive automatically inserts paddings.
This resolves this issue by inserting dummy values at such padding when encoding the result of umul_overflow.

This resolves type checking failures at LLVM unit tests & gzip single file benchmarks.
Exact numbers for LLVM unit tests - I'm running the tests

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -540,7 +540,7 @@ StateValue BinOp::toSMT(State &s) const {
                                     bi->non_poison));
       }
     }
-    return ty->aggregateVals(vals);
+    return ty->aggregateVals(vals, true);
   }
 
   if (vertical_zip) {
@@ -548,7 +548,7 @@ StateValue BinOp::toSMT(State &s) const {
     auto [v1, v2] = zip_op(a.value, a.non_poison, b.value, b.non_poison);
     vals.emplace_back(move(v1));
     vals.emplace_back(move(v2));
-    return getType().getAsAggregateType()->aggregateVals(vals);
+    return getType().getAsAggregateType()->aggregateVals(vals, true);
   }
   return scalar_op(a.value, a.non_poison, b.value, b.non_poison);
 }
@@ -567,8 +567,8 @@ expr BinOp::getTypeConstraints(const Function &f) const {
                   lhs->getType() == rhs->getType();
 
     if (auto ty = getType().getAsStructType()) {
-      instrconstr &= ty->numElements() == 2 &&
-                     ty->getChild(0) == lhs->getType() &&
+      // Note that ty->numElements() may not be 2 due to paddings
+      instrconstr &= ty->getChild(0) == lhs->getType() &&
                      ty->getChild(1).enforceIntOrVectorType(1) &&
                      ty->getChild(1).enforceVectorTypeEquiv(lhs->getType());
     }
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -567,8 +567,9 @@ expr BinOp::getTypeConstraints(const Function &f) const {
                   lhs->getType() == rhs->getType();
 
     if (auto ty = getType().getAsStructType()) {
-      // Note that ty->numElements() may not be 2 due to paddings
-      instrconstr &= ty->getChild(0) == lhs->getType() &&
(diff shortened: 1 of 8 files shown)


ir/instr.cpp
```diff
@@ -567,8 +567,8 @@ expr BinOp::getTypeConstraints(const Function &f) const {
                   lhs->getType() == rhs->getType();
 
     if (auto ty = getType().getAsStructType()) {
-      instrconstr &= ty->numElements() == 2 &&
-                     ty->getChild(0) == lhs->getType() &&
+      // Note that ty->numElements() may not be 2 due to paddings
```
nunoplopes, line 570: this is dangerous. can't we have a solution like the number of elements is two modulo padding?

ir/instr.cpp
```diff
@@ -540,15 +540,15 @@ StateValue BinOp::toSMT(State &s) const {
                                     bi->non_poison));
       }
     }
-    return ty->aggregateVals(vals);
+    return ty->aggregateVals(vals, true);
```
nunoplopes, line 543: is true needed here?

ir/type.cpp
```diff
@@ -701,19 +701,36 @@ expr AggregateType::numElements() const {
...
       continue;
+    }
 
-    auto vv = children[idx]->toBV(vals[idx]);
+    StateValue vv;
+    if (needsPadding && isPadding(idx))
+      vv = children[idx]->getDummyValue(true);
+    else
+      vv = vals[val_idx++];
+    vv = children[idx]->toBV(vv);
```
nunoplopes, line 733: move(vv)

ir/instr.cpp
aqjune: We have the number of padding as constant only so I didn't use the condition.
Would it be helpful to have the guard still?

ir/instr.cpp
aqjune: Yes, because `vals` does not contain padding values.

aqjune:
`program doesn't type check!`: 317 -> 23

4 failures appear (with disable-undef-byte set):
```
Transforms/InstCombine/sadd-with-overflow.ll
Transforms/InstCombine/ssub-with-overflow.ll
Transforms/InstCombine/uadd-with-overflow.ll
Transforms/PhaseOrdering/unsigned-multiply-overflow-check.ll
```
Some of them seems to be a bug in Alive2; sadd_overflow with vector inputs does not correctly propagate poison values to overflow bits.

ir/instr.cpp
aqjune: The check revived

aqjune:
BTW, I had to set the padding as poison to match it with get_operand's behavior (sadd_overflow.srctgt.ll ).

ir/instr.cpp
nunoplopes: but these don't return structures that need padding. only vectors, I think

ir/instr.cpp
```diff
@@ -567,7 +567,8 @@ expr BinOp::getTypeConstraints(const Function &f) const {
                   lhs->getType() == rhs->getType();
 
     if (auto ty = getType().getAsStructType()) {
-      instrconstr &= ty->numElements() == 2 &&
+      auto cnt = ty->numElements();
+      instrconstr &= cnt - expr::mkInt(ty->numPaddingsConst(), cnt) == 2 &&
```
nunoplopes, line 571: can you create a function in aggregateType for this, e.g. numElementsExcludingPadding().
Seems useful in general.

ir/instr.cpp
aqjune: It can happen because sadd.overflow can take vectors as well:
```
call { <2 x i32>, <2 x i1> } @llvm.sadd.with.overflow.v2i32(..)
```

ir/instr.cpp
aqjune: Updated

PR #352, opened 2020-04-08
llvm.maxnum, llvm.minnum support

zhengyangl:


zhengyangl:
Oops, it seems that fast math flags can be attached with these two intrinsics. I'll update the patch.

zhengyangl:
after fastmath flags is plugged, we got this, seems a bug to me.

```
define float @minnum_f32_1_minnum_p0_val_nnan_ninf(float %x) {
%0:
  %y = fminnum float 0.000000, %x
  %z = fminnum nnan ninf float %y, 1.000000
  ret float %z
}
=>
define float @minnum_f32_1_minnum_p0_val_nnan_ninf(float %x) {
%0:
  %1 = fminnum nnan ninf float %x, 0.000000
  ret float %1
}
ERROR: Target is more poisonous than source

Example:
float %x = NaN

Source:
float %y = #x00000000 (+0.0)
float %z = #x00000000 (+0.0)

Target:
float %1 = poison
Source value: #x00000000 (+0.0)
Target value: poison
```
It is this test case `llvm/test/Transforms/InstCombine/minnum.ll`

nunoplopes:
BTW, can we rename this to fmin/fmax? LLVM's langref states that these are similar to libc's fmin/fmax.

nunoplopes:
> after fastmath flags is plugged, we got this, seems a bug to me.
> 
> ```
> define float @minnum_f32_1_minnum_p0_val_nnan_ninf(float %x) {
> %0:
>   %y = fminnum float 0.000000, %x
>   %z = fminnum nnan ninf float %y, 1.000000
>   ret float %z
> }
> =>
> define float @minnum_f32_1_minnum_p0_val_nnan_ninf(float %x) {
> %0:
>   %1 = fminnum nnan ninf float %x, 0.000000
>   ret float %1
> }
> Transformation doesn't verify!
> ERROR: Target is more poisonous than source
> 
> Example:
> float %x = NaN
> 
> Source:
> float %y = #x00000000 (+0.0)
> float %z = #x00000000 (+0.0)
> 
> Target:
> float %1 = poison
> Source value: #x00000000 (+0.0)
> Target value: poison
> ```
> 
> It is in this test case `llvm/test/Transforms/InstCombine/minnum.ll`

I agree, it's a bug. Please report it :)

ir/instr.cpp
```diff
@@ -482,6 +486,28 @@ StateValue BinOp::toSMT(State &s) const {
...
+                            expr::mkNumber("-0", a));
+      
+      auto v = [&](expr &a, expr &b) {
+        expr z = a.isFPZero() && b.isFPZero();
+        // (+, -) or (-, +)
+        expr s = (a.isFPNeg() && !b.isFPNeg()) || (!a.isFPNeg() && b.isFPNeg());
+        expr cmp = (op == FMinNum) ? a.fole(b) : a.foge(b);
+        return expr::mkIf(a.isNaN(), b,
+                          expr::mkIf(b.isNaN(), a,
+                                     expr::mkIf(z && s, ndz,
```
nunoplopes, line 505: from my reading of langref, fmax(-0, -0) can return +0. If you agree, then `s` isn't needed here

zhengyangl: let me run TV on transforms suite again without s, and let's see if it is breaking anything.

zhengyangl: just to avoid spurious errors.

zhengyangl: yeah, allowing non-det +0, -0 for input (+0, +0) and (-0, -0) eliminates three Unexpected Failures when running TV on Transform suite.

PR #354, opened 2020-04-21
Add an command-line argument for omitting long array initializers

aqjune:
This PR adds a command-line option that that makes an array initializer omitted if it has more than the given number of elements.


llvm_util/llvm2alive.cpp
```diff
@@ -998,6 +1003,9 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
 
 namespace llvm_util {
 
+unsigned omit_array_size = -1;
+
```
aqjune, line 1007: This is declared here rather than at ir/globals.h because constants at globals.h are determined after an Alive2 function is analyzed.
Also, moving this to globals.h introduces dependency to ir/globals.h , which did not exist before.

aqjune:
A few tests regarding global variables are moved to `globals` directory too. They were small changes, so embedded here

nunoplopes:
This is a bit of a hack.. Temporarily is ok, but we need to work on an algorithm to automatically decide which initializers to keep.

PR #358, opened 2020-04-22
Resolve regression after #356

aqjune:
Resolve a crash of Transforms/ConstProp/extractvalue.ll .


aqjune:
I'll leave LLVM unit test result here after it finishes.

aqjune:
Failures: 75 -> 74, no new failure added

PR #359, opened 2020-04-22
Let Memory-related class use camel case for names of member functions

aqjune:
This PR makes classes in Memory.h use camel case for member functions.
All classes in the ir/ directory except those are already using camel cases, so I think it is natural to use the convention in them as well.


aqjune:
Running tests

aqjune:
Would this PR make sense?
Unit test result: no regression added

PR #361, opened 2020-04-23
Fix build with llvm trunk - llvm::Type::VectorTyID is now FixedVectorTyID

LebedevRI:
Since https://reviews.llvm.org/D77587 / https://github.com/llvm/llvm-project/commit/2dea3f129878e929e5d1f00b91a622eb1ec8be4e
Do we need to somehow uglify this to make building with older LLVM still possible?


LebedevRI:
Thanks!

PR #366, opened 2020-04-24
ptr_deref_within does not need nondeterministic comparison

aqjune:
https://github.com/AliveToolkit/alive2/pull/313 changed Pointer::slt/sle/.. to return nondeterministic result when pointers to different blocks are given.

This affected how ptr_deref_within works, which only needs the comparison between offsets because equality of block ids are compared separately.

This patch brings performance improvement by not using the Pointer::slt/slt/.., hence not introducing a quantified variable.

LLVM unit tests timeout: 1083 -> 1073


aqjune:
FYI, this patch is rolling back ptr_deref_within to the state before #313 was merged, IIUC.

PR #367, opened 2020-04-26
Use non-local bid 0 as a plain block id if there is no null block

aqjune:
Use non-local bid 0 as a plain block id if there is no null block.

I added has_null_block which represents whether either src or tgt needs the existence of the null block.


aqjune:
Will leave a comment after unit test is done

aqjune:
No additional regression, timeout: 1073 -> 1056

PR #368, opened 2020-04-26
Let pointer comparison compare absolute addresses

aqjune:
*NOTE: This description is out-of-date; see the comments below

---

This has two changes:

- Let icmp ptr1, ptr2 compare their absolute addresses if either points to the null block.

- Add refinement between ptr and null if ptr.getAddress() == 0. This allows
```
if (ptr == null) use(ptr);
=>
if (ptr == null) use(null);
```

Both are based on the twin memory semantics.

Unfortunately, this patch isn't sufficient to support Transforms/InstSimplify/and-or-icmp-nullptr.ll that was reported by @LebedevRI ... :( Currently icmp ult/ugt... does not consider block sizes, so it still says it's wrong.

BTW, It is amazing that precisely defining pointer comparison is a pretty hard job.

---

### DIFF
--- a/tests/alive-tv/ptrcmp/lifetime-cmp.srctgt.ll
+++ b/tests/alive-tv/ptrcmp/lifetime-cmp.srctgt.ll
@@ -1,0 +1,58 @@
+define void @src() {
+	%p = alloca i32
+	%p8 = bitcast i32* %p to i8*
+	%q = alloca i32
+	%q8 = bitcast i32* %q to i8*
+	call void @llvm.lifetime.start.p0i8(i64 4, i8* %p8)
+	call void @llvm.lifetime.end.p0i8(i64 4, i8* %p8)
+
+	call void @llvm.lifetime.start.p0i8(i64 4, i8* %q8)
+
+	%ip = ptrtoint i32* %p to i64
+	%iq = ptrtoint i32* %q to i64
+	%guard = icmp eq i64 %ip, %iq
+	br i1 %guard, label %COMPARE, label %EXIT
+
+COMPARE:
+	%c0 = icmp eq i8* %p8, %q8
+	%qgep1 = getelementptr inbounds i8, i8* %q8, i64 1
+	%c1 = icmp eq i8* %p8, %qgep1
+	call void @g(i1 %c0) ; false
+	call void @g(i1 %c1) ; false
+	br label %EXIT
+
+EXIT:
+	call void @llvm.lifetime.end.p0i8(i64 4, i8* %q8)
+	ret void
+}
+
+define void @tgt() {
+	%p = alloca i32
+	%p8 = bitcast i32* %p to i8*
+	%q = alloca i32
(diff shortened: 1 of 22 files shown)


LebedevRI:
Thank you for looking into this!

tests/alive-tv/ptrcmp/lifetime-cmp.srctgt.ll
```diff
@@ -0,0 +1,58 @@
...
+
+	%ip = ptrtoint i32* %p to i64
+	%iq = ptrtoint i32* %q to i64
+	%guard = icmp eq i64 %ip, %iq
+	br i1 %guard, label %COMPARE, label %EXIT
+
+COMPARE:
+	call void @g(i1 false)
+	call void @g(i1 false)
+	br label %EXIT
```
aqjune, line 47: This should fail but succeeds due to #385 .

aqjune:
(As discussed with Nuno) we chose to simply compare the integer addresses of pointers rather than adopting the definition of the twin memory paper.

The reason is as follows:

1. If we make comparison of pointers to different blocks return nondeterministic value (as defined in the twin memory paper), the comparison with null becomes problematic:
 
    - If comparison with null just compares underlying integer addresses, it breaks the and-or-icmp-null optimization. In target, comparison with null is removed, so the result becomes nondeterministic.

    - If comparison with null returns nondeterministic value, it does not explain nonnull attribute; If `icmp ne ptr, null` holds in a branch condition, (int)ptr should not be zero.
 
So the comparison of two different blocks should yield a deterministic value.
 
2. The twin memory's pointer comparison semantics does not support icmp (gep p, i), (gep q, i) to icmp p, q.

aqjune:
Resolved LLVM unit test failures:

```
Transforms/InstCombine/and-or-icmp-nullptr.ll
Transforms/InstCombine/assume.ll
Transforms/InstCombine/cast_ptr.ll
Transforms/InstCombine/icmp.ll
Transforms/InstCombine/or.ll
Transforms/InstSimplify/and-or-icmp-nullptr.ll
```

So finally `and-or-icmp-nullptr.ll` is supported..!

Added failures:
```
Transforms/GVN/condprop.ll : propagation of pointer equality 
Transforms/CorrelatedValuePropagation/phi-common-val.ll : replaces null with p s.t. (int)p == 0
Transforms/InstCombine/gep-custom-dl.ll : icmp (gep p, i), (gep q, i) => icmp p, q is incorrect in some data layout
```

Timeouts: 868 -> 876

To resolve some regressions, other than updating pointer comparison, I had to add two things:

- Refinement from p to null if (int)p is zero. This is allowed in the twin memory paper
- free(p) is no-op if (int)p is zero. This was needed to hoist free(p) out of if(p) branch.

aqjune:
One regression: alive-tv/memory/calloc-overflow.src.ll raises precondition is false error, but I guess this is an error in the encoding of calloc...

PR #369, opened 2020-04-27
Improving how UB smt formulas are constructed by using ite's.

kevindevos:
This PR changes the way UB formulas are built by traversing the CFG similar to a DFS. This requires using dominators and manually building a post-dominator relationship between a merge and its dominator.

The core building of UB is in the buildUB function, it first traverses in a topdown manner in order to guarantee child-before parent order. In the second traversal, when a block is visited twice, the ub for this bb will be constructed as an ite of all the direct sucessors of this block, thus containing its own ub and the ub of all sucessors (until the leafs). 

The algorithm may also move an UB expression of a merge to its dominator either when:
- the merge post-dominates its dominator
- the merge partly post-dominates its dominator and the cost of duplicating a path formula is estimated to be lower than the cost of duplicating UB. 
The more paths there are to a merge, the more often its UB would be duplicated without this change, and in some cases it may not pay off to move the expression of a merge to its dominator if the estimated size of UB is small and the number of paths is large, which would lead to much larger formulas.
The path formula is constructed in buildPostdomBreakingBBs and is a ite formula that is true for each block that breaks the post-dominator relationship. The expression moved in this case is no longer the ub of the merge, but not(breakingPostDomPath) => mergeUB.

In certain cases there are blocks that wrongly prevent the optimization from taking place, these blocks include:
- blocks with only back-edges
- blocks with unreachable
- blocks that once reached, inevitably reach a basic block of one of the two above types.
Identifying these blocks is done with the propagateNoRetBB function, which stores them in a set. Later when the postdominance relationship is tested, the blocks in this set will be ignored and will soften the restrictions to consider post-dominance to satisfy.

Finally, the algorithm also calculates an estimate of the size of UB for a given block (which includes all UB of its sucessors until all leafs), and the size of the potential path expression used to guard the implies when applying the above transformation. The path size is simply the number of jumps between a block and the entry of the program. UB size is the number of expressions added to the given block and all its sucessors.
Moving the merge UB to its dominator only compensates when the estimated size of the UB duplicated is larger than the estimated size of the added path formula used as a guard for the implies of this transformation.


regehr:
hey @zhengyangl, can you look into Transforms/LoopVectorize/start-non-zero.ll?

zhengyangl:
sure I'll have a look.

zhengyangl:
@kevindevos both master and your branch (rebased) are showing timeout on the Transforms/LoopVectorize/start-non-zero.ll. I am wondering what does the log look like on your side.

aqjune:
This is great..!
Transforms/SimplifyCFG/PhiEliminate3.ll might be affected by 19cb7000cd1c906dc347a45659d90b8b989a2ebb , but I'm not fully sure.
I tried to see the value mismatch error, but my alive2 raised timeout too. Could you please share the value mismatch error of PhiEliminate3.ll as well?

kevindevos:
I will have a look on both tests, it is also possible that it now gives timeout on PhiEliminate3.ll due to the changes made on alive since the version i tested on.

kevindevos:
@zhengyangl  @aqjune  Both tests seem to pass for me on the master and on the rebase with llvm-lit. I will have to rerun the test suite for master and this rebase to check for the rest (if llvm allows me to as version 11 is making my memory usage explode). Sorry for the confusion, i guess the version i tested on is a bit too old and results changed since then.

smt/exprs.h
```diff
@@ -102,6 +102,7 @@ class DisjointExpr {
   auto begin() const { return vals.begin(); }
   auto end() const   { return vals.end(); }
   auto size() const  { return vals.size(); }
+  auto empty() const { return vals.size() == 0; }
```
nunoplopes, line 105: vals.empty()

ir/state.h
```diff
@@ -66,23 +66,24 @@ class State {
   // dst BB -> src BB -> (domain data, memory)
```
aqjune, line 66: Could you update this comment as well?

ir/state.h
```diff
@@ -107,6 +108,16 @@ class State {
   };
   std::map<std::string, std::map<FnCallInput, FnCallOutput>> fn_call_data;
 
+  // src -> <(dst, cond), isolated ub>
+  std::unordered_map<const BasicBlock*, std::pair<std::vector<std::pair<
+                                                              const BasicBlock*,
+                                                              smt::expr>>,
+                                                              std::optional<
+                                                              smt::expr>>>
```
aqjune, line 116: What about declaring a separate struct (just above this field definition) that stores the data instead of using std::pair< ... >?

util/symexec.cpp
```diff
@@ -48,6 +48,9 @@ void sym_exec(State &s) {
 
     first = false;
   }
+  
+  if (s.foundReturn())
+    s.buildUB();
```
aqjune, line 53: Does with work with a basic block ending with unreachable as well?

kevindevos: It does work, if there's no return then the return_domain takes the default value of false. Or if there is a return and there's an unreachable in between, the UB should be added for the assume, which should make the expression false when the input to the assume is 0.

ir/state.h
kevindevos: Thanks!, i changed it and it should be a bit more readable now.

ir/state.h
kevindevos: Done!

kevindevos:
Made a couple of changes to the algorithm, namely an optimization to reduce redundancy in the formulas as much as possible using the concept of post-dominator logic. 
This requires creating a dominator tree, for which i will PR a fix shortly.

Very briefly the idea behind this optimization is to reduce the number of times certain expressions appear in ite's by checking whether or not a BB B post-dominates A, in which case B will always be visited whenever A is visited, thus the expressions added at B can be moved to the same level as those added in A. The function that makes this check is for now called canMoveExprsToDom.

To even further optimize this, there are certain cases where a path coming from some branch never reaches a return unless we include back-edges, this would fail the check for canMoveExprsToDom and prevent the optimization from being applied. For this the function propagateNoRetBB identifies these BB's that either always lead to an unreachable or a jump with only back-edges, allowing the check at canMoveExprsToDom to ignore these and let the optimizations take place in these scenarios.

All tests pass but one function in Transforms/IndVarSimplify/widen-loop-comp.ll, telling me source is more defined than target while master just gives a timeout. I'm not sure how to go about resolving this, especially when i don't think i've changed function_domain. 
I will run the tests once more because of this merge just to make sure, and change this to open status if i get the same results.

Not sure what happened with the rebase adding so many commits, it appears to have added duplicates.

kevindevos:
I've spent a lot of time figuring things out and experimenting on what i should do for the ub formulas of function_domain for noreturn.
For now if no noreturn call is present the function_domain is simply replaced by the return_domain.  There's also some code that isn't used anymore, for example buildTargetData, which i used to run the ub algorithm separately for a given noreturn call, from which the resulting ub formulas may not reduce solving time enough to compensate running the algorithm multiple times. I also don't yet see an easy solution to building better ub specifically for function_domain such that all noreturn domains and all valid returns are considered into the formula.

@aqjune @nunoplopes Also forgive me for asking since my knowledge on noreturn and what problems it can impose is limited, but would it be a good idea to not check function domain in the solver if no noreturn call was seen in a run? It seems to me that in this case the return_domain and function_domain are identical and seems the solver is doing duplicate work? 

So far on the fastest 4500 llvm tests everything seems to work well, however i will run the full llvm test suite and report back with the results.

nunoplopes:
Close this one as unfortunately it doesn't improve performance. Z3 is too smart.. 😎

PR #370, opened 2020-05-02
Implement memcmp/bcmp

aqjune:
This implements memcmp/bcmp by manually unrolling comparisons `memcmp_unroll_cnt` times.

This only considers integer bytes (otherwise poison), but it already seems to work quite well for LLVM unit tests.

Has one regression from unit tests: Transforms/ExpandMemCmp/X86/memcmp.ll due to this kind of transformation:

```
define i32 @cmp2(* nocapture readonly %x, * nocapture readonly %y) {                
%0:                                                                                 
  %call = memcmp * nocapture readonly %x, * nocapture readonly %y, i64 2            
  ret i32 %call                                                                     
}                                                                                   
=>                                                                                  
define i32 @cmp2(* nocapture readonly %x, * nocapture readonly %y) {                
%0:                                                                                 
  %1 = bitcast * nocapture readonly %x to *                                         
  %2 = bitcast * nocapture readonly %y to *                                         
  %3 = load i16, * %1, align 1                                                      
  %4 = load i16, * %2, align 1                                                      
  %5 = bswap i16 %3                                                                 
  %6 = bswap i16 %4                                                                 
  %7 = zext i16 %5 to i32                                                           
  %8 = zext i16 %6 to i32                                                           
  %9 = sub i32 %7, %8                                                               
  ret i32 %9                                                                        
}
```
If x[0] = 0, y[0] = 1, x[1] = y[1] = poison, %3 & %4 in tgt are full-poison whereas src only compares the first non-poison byte and returns non-zero.

This is what is expected & also a ugly part of the current status; bitwise poison will explain this, but it will be a huge pain for checking correctness of all other optimizations because we don't want to redefine all relevant operations bitwisely.

Actually, I think this result also makes sense somehow in perspective of a sanitizer: before transformation, it does not touch x[1]/y[1], but after that it touches it and does arithmetic operation on it.

Fixing these transformations in LLVM doesn't seem trivial; we can introduce notions like 'load with bytes frozen', or use vectorized load (load <2 x i8>), but any of these is likely to effect performance, so it will need some work.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2403,6 +2403,110 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
 }
 
 
+
+static pair<expr,expr>
+encode_memcmp(State &s, const Pointer &p1, const Pointer &p2, AndExpr &prefix,
+              const expr &max_num, bool is_bcmp, unsigned i) {
+  expr ei = expr::mkUInt(i, max_num.bits());
+  expr zero = expr::mkUInt(0, 32);
+  if (ei.uge(max_num).isTrue())
+    return { move(zero), true };
+
+  // TODO: Pointers?
+  auto [val1, ub1] = s.getMemory().load((p1 + i)(), IntType("i8", 8), 1);
+  auto [val2, ub2] = s.getMemory().load((p2 + i)(), IntType("i8", 8), 1);
+  AndExpr ub_and;
+  ub_and.add(val1.non_poison);
+  ub_and.add(val2.non_poison);
+  ub_and.add(move(ub1));
+  ub_and.add(move(ub2));
+
+  expr val_eq = val1.value == val2.value;
+
+  expr result_neq;
+  if (is_bcmp) {
+    result_neq = expr::mkFreshVar("bcmp_nonzero", zero);
+    s.addQuantVar(result_neq);
+    s.addPre(result_neq != zero);
+  } else {
+    expr pos = expr::mkFreshVar("memcmp_pos", expr::mkUInt(0, 31));
+    expr neg = expr::mkFreshVar("memcmp_neg", expr::mkUInt(0, 31));
+    s.addQuantVar(pos);
+    s.addQuantVar(neg);
+    s.addPre(pos != expr::mkUInt(0u, 31));
+    pos = expr::mkUInt(0, 1).concat(pos);
+    neg = expr::mkUInt(1, 1).concat(neg);
+    result_neq = expr::mkIf(val1.value.ugt(val2.value), move(pos), move(neg));
+  }
+
+  expr ub = ub_and();
+  if (val_eq.isFalse())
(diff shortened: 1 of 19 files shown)


tests/alive-tv/memory/memcmp-ub.src.ll
```diff
@@ -0,0 +1,58 @@
+target datalayout = "e-p:64:64:64"
+
+define i32 @ub() {
+  %p = alloca i32
+  %p8 = bitcast i32* %p to i8*
+  %res = call i32 @memcmp(i8* %p8, i8* %p8, i64 4) ; ub
```
nunoplopes, line 6: why is this UB? The input has 4 bytes.

aqjune: Because it is comparing poison bytes. If memcmp is converted into a loop, it has branch on poison, so this is UB.

ir/instr.cpp
```diff
@@ -2403,6 +2403,110 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
 }
 
 
+
+static pair<expr,expr>
+encode_memcmp(State &s, const Pointer &p1, const Pointer &p2, AndExpr &prefix,
```
nunoplopes, line 2408: isn't it possible to refactor this to share the code with strlen()? That was the original plan I guess. Have a generic way to implement these loopy intrinsics.

aqjune: That makes sense - we have prefix, and i, and maximum unroll count. I will work on it.

ir/instr.cpp
```diff
@@ -2403,48 +2434,124 @@ unique_ptr<Instr> Memcpy::dup(const string &suffix) const {
...
+    ub_and.add(val1.non_poison);
+    ub_and.add(val2.non_poison);
+    ub_and.add(move(ub1));
+    ub_and.add(move(ub2));
+
+    expr ei = expr::mkUInt(i, vn.bits());
+    auto val_eq = val1.value == val2.value;
+    expr continue_cond = val_eq && ei.ult(vn);
+    if (is_last)
+      continue_cond &= (p1 != p2);
```
aqjune, line 2467: This allows early exit, helping prove `memcmp(p, p, n) -> 0` optimization.

aqjune:
I made the commit, but what about making this as a separate PR? It is making the diff bigger.

nunoplopes:
@aqjune Please have a look at https://github.com/AliveToolkit/alive2/commit/ebdc5015712c84e671359aac545a9e6643bd68b8 for some follow up simplifications.
In particular I reduced from 2->1 quantified vars per iteration. This one was an easy reduction; to reduce to just 1 var overall would required the prefix formula and thus it becomes less clear if it's profitable.

aqjune:
Yes, I saw that change and I think it makes sense.

PR #371, opened 2020-05-08
Define memcmp(p,q,0) as zero, fix the last iteration condition

aqjune:
According to man page, memcmp(p, q, 0) is defined as zero: http://man7.org/linux/man-pages/man3/memcmp.3.html .
Another fix is that the next iteration condition of memcmp was incorrectly defined; it should compare i+1 < n. For example, if n == 1, there should be no second iteration (i = 0 should be the last one). This resolves two regressions (Transforms/InstCombine/bcmp-1.ll , Transforms/InstCombine/memcmp-1.ll ).

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2498,12 +2498,12 @@ StateValue Memcmp::toSMT(State &s) const {
     auto val_eq = val1.value == val2.value;
     return { expr::mkIf(val_eq, zero, result_neq),
              move(ub_and),
-             val_eq && expr::mkUInt(i, vn.bits()).ult(vn) };
+             val_eq && expr::mkUInt(i + 1, vn.bits()).ult(vn) };
   };
   auto [val, ub]
     = LoopLikeFunctionApproximator(ith_exec).encode(s, memcmp_unroll_cnt);
-  s.addUB(move(ub));
-  return { move(val), true };
+  s.addUB(vnum.ugt(0).implies(move(ub)));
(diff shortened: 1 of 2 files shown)


nunoplopes:
@aqjune 3 memcmp unit tests were failing with this patch.. (fixed here: https://github.com/AliveToolkit/alive2/commit/f7ff96781b9d02bb2399a3de36b69994c3528dc3)

aqjune:
Sorry, it was my mistake. :( `vnum == zero` had sort mismatch.
For the ule -> ult part, I think the original code was correct; if vnum was 1, the continue variable should be false because one iteration is enough. To assure myself, I'll run LLVM unit tests (as well as alive2 unit tests).

nunoplopes:
that was just a simplification; ule is better than ult. Internally Z3 only has ule/sle; everything is expressed on top of these.

aqjune:
I think the number of iterations changed after the fix: if ule was used, it should have been `i + 2 <= n`.

aqjune:
With https://github.com/aqjune/alive2/commit/068b5d038818c26903b9d47639bafdcf584bd19c , two tests (`Transforms/InstCombine/bcmp-1.ll` , `Transforms/InstCombine/memcmp-1.ll`) can be fixed. May I make a PR for this?

nunoplopes:
Ah, you're right. My patch was bogus. Please do a PR, yes, thanks

PR #372, opened 2020-05-09
Use set for non_local_blk_nonwritable to prune out duplicated bids

aqjune:
This resolves the stack overflow error on some large input bitcode pairs:

[overflow.zip](https://github.com/AliveToolkit/alive2/files/4602945/overflow.zip)

The reason was as follows:

(1) non_local_blk_nonwritable is a vector of bids
(2) Memory::mkIf merges two non_local_blk_nonwritable s without pruning out duplicated bids, causing exponential growth
(3) Pointer::isWritable does `non_local &= this_bid != bid` for each bid in non_local_blk_nonwritable
(4) AndExpr::add, which is called by Pointer::isDereferenceable, raises stack overflow when the expression creatd from Pointer::isWritable is given.

By defining non_local_blk_nonwritable as set instead, step (2) isn't happening anymore.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -1411,7 +1411,7 @@ Memory::alloc(const expr &size, unsigned align, BlockKind blockKind,
       state->addAxiom(p.getAddress().extract(align_bits - 1, 0) == 0);
 
     if (blockKind == CONSTGLOBAL)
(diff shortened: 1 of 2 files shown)


nunoplopes:
nice, thanks!

PR #374, opened 2020-05-09
Define MemInstr & let memory related classes inherit from this

aqjune:
This PR defines MemInstr and makes memory related instructions inherit this class.

### DIFF
--- a/ir/instr.h
+++ b/ir/instr.h
@@ -465,10 +465,9 @@ public:
   bool initDead() const { return initially_dead; }
 
   uint64_t getMaxAllocSize() const override;
-  uint64_t getMaxAccessSize() const override { return 0; }
-  uint64_t getMaxGEPOffset() const override { return 0; }
-  ByteAccessInfo getByteAccessInfo() const override
-  { return ByteAccessInfo::empty(); }
+  uint64_t getMaxAccessSize() const override;
+  uint64_t getMaxGEPOffset() const override;
+  ByteAccessInfo getByteAccessInfo() const override;
 
   std::vector<Value*> operands() const override;
   void rauw(const Value &what, Value &with) override;
@@ -495,12 +494,10 @@ public:
   Value& getSize() const { return *size; }
   bool isRealloc() const { return ptr != nullptr; }
 
-  uint64_t getMaxAllocSize() const override
-  { return getIntOr(*size, UINT64_MAX); }
-  uint64_t getMaxAccessSize() const override { return 0; }
-  uint64_t getMaxGEPOffset() const override { return 0; }
-  ByteAccessInfo getByteAccessInfo() const override
-  { return ByteAccessInfo::empty(); }
+  uint64_t getMaxAllocSize() const override;
+  uint64_t getMaxAccessSize() const override;
+  uint64_t getMaxGEPOffset() const override;
+  ByteAccessInfo getByteAccessInfo() const override;
 
   std::vector<Value*> operands() const override;
   void rauw(const Value &what, Value &with) override;
@@ -521,8 +518,8 @@ public:
   Value& getSize() const { return *size; }
 
   uint64_t getMaxAllocSize() const override;
-  uint64_t getMaxAccessSize() const override { return 0; }
-  uint64_t getMaxGEPOffset() const override { return 0; }
+  uint64_t getMaxAccessSize() const override;
+  uint64_t getMaxGEPOffset() const override;
   ByteAccessInfo getByteAccessInfo() const override;
 
   std::vector<Value*> operands() const override;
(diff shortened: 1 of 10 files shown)


aqjune:
Well, push force was affecting the history of this PR too... :/
I'm running LLVM unit tests.
BTW, the diff is pretty big. Do you want me to split this into (1) Add MemInstr (2) Let analyses use it? This PR will be what the final code look like.

aqjune:
LLVM unit test result: +3 errors
```
Transforms/InstCombine/bitcast-phi-uselistorder.ll
Transforms/InstCombine/gep-combine-loop-invariant.ll
Transforms/InstCombine/load.ll
```
but they are all conversions from timeouts.
 bitcast-phi-uselistorder.ll is https://bugs.llvm.org/show_bug.cgi?id=45152
load.ll is already registered as wrong in the dashboard too.
gep-combine-loop-invariant.ll is https://bugs.llvm.org/show_bug.cgi?id=44861

ir/instr.h
```diff
@@ -569,19 +682,26 @@ class Memset final : public Instr {
...
          unsigned align_dst, unsigned align_src, bool move)
-    : Instr(Type::voidTy, "memcpy"), dst(&dst), src(&src), bytes(&bytes),
+    : MemInstr(Type::voidTy, "memcpy"), dst(&dst), src(&src), bytes(&bytes),
             align_dst(align_dst), align_src(align_src), move(move) {}
 
   Value& getBytes() const { return *bytes; }
   unsigned getSrcAlign() const { return align_src; }
   unsigned getDstAlign() const { return align_dst; }
+
+  uint64_t getMaxAllocSize() const override { return 0; }
```
nunoplopes, line 699: please don't define virtual methods in the headers. this increases compile time a lot.

ir/type.cpp
```diff
@@ -1398,4 +1398,32 @@ bool isNonPtrVector(const Type &t) {
...
+
+    for (unsigned i = 0, e = agg->numElementsConst(); i != e;  ++i) {
+      if (hasSubByte(agg->getChild(i)))
+        return true;
+    }
+  }
+  return false;
+}
+
+uint64_t getCommonAccessSize(const IR::Type &ty) {
```
nunoplopes, line 1416: Common is not a good name, as the meaning is not immediate in this case. Perhaps `getAccessSize` is enough?

aqjune: I think `getAccessSize(ty)` can be interpreted as `sizeof(ty)`..
Or what about `getByteSize`?

aqjune:
Rebased

tools/transform.cpp
```diff
@@ -746,12 +511,63 @@ static void calculateAndInitConstants(Transform &t) {
...
+      for (auto &i : BB->instrs()) {
+        if (returns_local(i))
+          ++num_locals;
+        else
+          num_max_nonlocals_inst += returns_nonlocal(i);
+
+        if (auto *mi = dynamic_cast<const MemInstr *>(&i)) {
+          max_alloc_size  = max(max_alloc_size, mi->getMaxAllocSize());
+          max_access_size = max(max_access_size, mi->getMaxAccessSize());
+          max_gep         = add_saturate(max_gep, mi->getMaxGEPOffset());
```
nunoplopes, line 529: I think this needs to add alloc_size & access size as well.
Otherwise the code doesn't seem equivalent to what we had before. I'm surprise no test fails (it should!). When doing loads, allocs, etc, we need offsets to be big enough to accommodate the UB constraints.

ir/instr.h
```diff
@@ -606,13 +716,20 @@ class Memcmp final : public Instr {
...
-class Strlen final : public Instr {
+class Strlen final : public MemInstr {
   Value *ptr;
 public:
   Strlen(Type &type, std::string &&name, Value &ptr)
-    : Instr(type, std::move(name)), ptr(&ptr) {}
+    : MemInstr(type, std::move(name)), ptr(&ptr) {}
 
   Value *getPointer() const { return ptr; }
+
```
nunoplopes, line 726: you missed to outline the methods below :)

aqjune: Sorry, it's done.

tools/transform.cpp
aqjune: It's done at line 634 and below.
Maybe the name of max_gep was a bit confusing (there is max_geps as well), so I changed it

nunoplopes: the alloc size is not considered AFAICT

aqjune: I think max_alloc_size was reflected as well:
```
  // we need an extra bit because 1st bit of size is always 0
  bits_size_t = ilog2_ceil(max_alloc_size, true);
  bits_size_t = min(max(bits_for_offset, bits_size_t)+1, bits_program_pointer);
```
`max_alloc_size` was reflected to `bits_size_t` but not to `bits_for_offset` in the original code.

PR #376, opened 2020-05-12
Factor out attributes as an independent class

aqjune:
Before working on dereferenceable attributes, this factors out attributes as an independent class, because dereferenceable carries additional information (the num of bytes) which cannot be simply encoded as a bit.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1271,8 +1271,8 @@ unique_ptr<Instr> InsertValue::dup(const string &suffix) const {
 }
 
 
-void FnCall::addArg(Value &arg, unsigned flags) {
-  args.emplace_back(&arg, flags);
+void FnCall::addArg(Value &arg, const Attributes &attrs) {
+  args.emplace_back(&arg, attrs);
 }
 
 vector<Value*> FnCall::operands() const {
@@ -1295,13 +1295,11 @@ void FnCall::print(ostream &os) const {
   os << "call " << print_type(getType()) << fnName << '(';
 
   bool first = true;
-  for (auto &[arg, flags] : args) {
+  for (auto &[arg, attrs] : args) {
     if (!first)
       os << ", ";
 
-    if (flags & ArgByVal)
-      os << "byval ";
-    os << *arg;
+    os << attrs.str() << *arg;
     first = false;
   }
   os << ')';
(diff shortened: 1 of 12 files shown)


ir/memory.cpp
```diff
@@ -356,18 +356,18 @@ static unsigned bits_shortbid() {
...
-static expr attr_to_bitvec(unsigned attributes) {
+static expr attr_to_bitvec(const ParamAttrs &attrs) {
   if (!bits_for_ptrattrs)
     return expr();
 
   uint64_t bits = 0;
   auto idx = 0;
-  auto to_bit = [&](bool b, Input::Attribute a) -> uint64_t {
-    return b ? (((attributes & a) ? 1 : 0) << idx++) : 0;
+  auto to_bit = [&](bool b, ParamAttrs::Attribute a) -> uint64_t {
```
nunoplopes, line 365: needs reference here

ir/instr.cpp
```diff
@@ -1271,8 +1271,8 @@ unique_ptr<Instr> InsertValue::dup(const string &suffix) const {
 }
 
 
-void FnCall::addArg(Value &arg, unsigned flags) {
-  args.emplace_back(&arg, flags);
+void FnCall::addArg(Value &arg, const ParamAttrs &attrs) {
```
nunoplopes, line 1274: this should probably be passed as &&

llvm_util/llvm2alive.cpp
```diff
@@ -271,11 +271,11 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
     unique_ptr<Instr> ret_val;
     auto argI = fn->arg_begin(), argE = fn->arg_end();
     for (auto &arg : args) {
-      unsigned attr = FnCall::ArgNone;
+      ParamAttrs attr = ParamAttrs::None;
```
nunoplopes, line 274: Maybe introduce an empty constructor

ir/value.cpp
```diff
@@ -188,22 +173,24 @@ StateValue Input::toSMT(State &s) const {
...
-    val = getType().mkInput(s, smt_name.c_str(), attributes);
+    val = getType().mkInput(s, smt_name.c_str(), attrs);
   }
 
   if (!config::disable_undef_input) {
-    auto [undef, vars] = getType().mkUndefInput(s, attributes);
-    for (auto &v : vars) {
-      s.addUndefVar(move(v));
+    auto [undef, vars] = getType().mkUndefInput(s, attrs);
+    if (undef.isValid()) {
```
nunoplopes, line 188: why this change? this is unrelated.
in general, we don't optimize for !isValid() case since most formulas should be valid.

aqjune: Oh yes, this is an unrelated change. I'll revert this change.

PR #377, opened 2020-05-13
Support dereferenceable attribute

aqjune:
This PR adds support for dereferenceable attribute at function parameters.

Dereferenceable for function's return value isn't included because it requires factoring out function attributes as a class, which will make the diff big.

LLVM unit tests: two failures added;
```
Transforms/MergeICmps/X86/entry-block-shuffled.ll
Transforms/MergeICmps/X86/multiple-blocks-does-work.ll
```

The first one does `if (x) { if (y) { ... } }` -> `if (y) { if (x) { ... } }` which is illegal because it may introduce UB.
The second one is related with memcmp's semantics; Currently, memcmp raises UB if poison bytes are compared, but this requires memcmp to return poison...

### DIFF
--- a/ir/attrs.h
+++ b/ir/attrs.h
@@ -9,15 +9,18 @@ namespace IR {
 
 class ParamAttrs final {
   unsigned bits;
+  uint64_t derefBytes;
 
 public:
   enum Attribute { None = 0, NonNull = 1<<0, ByVal = 1<<1, NoCapture = 1<<2,
-                   ReadOnly = 1<<3, ReadNone = 1<<4 };
+                   ReadOnly = 1<<3, ReadNone = 1<<4, Dereferenceable = 1<<5 };
 
   ParamAttrs(unsigned bits = None) : bits(bits) {}
 
   bool has(Attribute a) const { return (bits & a) != 0; }
   void set(Attribute a) { bits |= (unsigned)a; }
+  uint64_t getDerefBytes() const { return derefBytes; }
+  void setDerefBytes(uint64_t bytes) { derefBytes = bytes; }
 
   friend std::ostream& operator<<(std::ostream &os, const ParamAttrs &attr);
 };
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -272,55 +272,52 @@ public:
       make_unique<FnCall>(*ty, value_name(i), move(fn_name), move(attrs),
                           !known);
     unique_ptr<Instr> ret_val;
-    auto argI = fn->arg_begin(), argE = fn->arg_end();
-    const auto &alist = i.getAttributes();
-    uint64_t argidx = 0;
 
-    for (auto &arg : args) {
+    for (uint64_t argidx = 0; argidx < i.arg_size(); ++argidx) {
+      auto *arg = args[argidx];
       ParamAttrs attr;
       // TODO: Once attributes at a call site are fully supported, we should
       // call handleAttributes().
 
-      if (uint64_t derefb = alist.getParamDereferenceableBytes(argidx)) {
+      if (i.paramHasAttr(argidx, llvm::Attribute::Dereferenceable)) {
         attr.set(ParamAttrs::Dereferenceable);
+
+        uint64_t derefb = 0;
+        // dereferenceable at caller
+        if (i.getAttributes()
+             .hasParamAttr(argidx, llvm::Attribute::Dereferenceable))
+          derefb = i.getParamAttr(argidx, llvm::Attribute::Dereferenceable)
+                    .getDereferenceableBytes();
+        if (argidx < i.getCalledFunction()->arg_size())
+          derefb = max(derefb,
+              i.getCalledFunction()->getParamDereferenceableBytes(argidx));
+        assert(derefb);
         attr.setDerefBytes(derefb);
       }
 
-      if (argI != argE) {
-        // Check whether arg itr finished early because it was var arg
-        if (argI->hasByValAttr())
-          attr.set(ParamAttrs::ByVal);
+      if (i.paramHasAttr(argidx, llvm::Attribute::ByVal))
+        attr.set(ParamAttrs::ByVal);
 
-        uint64_t derefb = argI->getType()->isPointerTy() ?
-                          argI->getDereferenceableBytes() : 0;
-        if (derefb) {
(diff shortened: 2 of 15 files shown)


llvm_util/llvm2alive.cpp
```diff
@@ -270,13 +270,30 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
+        attr.derefBytes = derefb;
+      }
+
       if (argI != argE) {
         // Check whether arg itr finished early because it was var arg
         if (argI->hasByValAttr())
           attr.set(ParamAttrs::ByVal);
-        else if (argI->hasReturnedAttr()) {
+
+        uint64_t derefb = argI->getType()->isPointerTy() ?
```
nunoplopes, line 289: can this not be a pointer??

llvm_util/llvm2alive.cpp
```diff
@@ -270,13 +270,30 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
                                     !known);
     unique_ptr<Instr> ret_val;
     auto argI = fn->arg_begin(), argE = fn->arg_end();
+    auto alist = i.getAttributes();
```
nunoplopes, line 273: &

llvm_util/llvm2alive.cpp
```diff
@@ -270,13 +270,30 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
+    uint64_t argidx = 0;
+
     for (auto &arg : args) {
       ParamAttrs attr;
```
nunoplopes, line 277: can we reuse handleAttribute() here?

ir/attrs.h
```diff
@@ -28,6 +29,8 @@ class ParamAttrs final {
       ret += "readonly ";
     if (has(ParamAttrs::ReadNone))
       ret += "readnone ";
+    if (has(ParamAttrs::Dereferenceable))
```
nunoplopes, line 32: this code is now conflicted, sorry

llvm_util/llvm2alive.cpp
aqjune: getDereferenceableBytes() fails if argI was not a pointer.

llvm_util/llvm2alive.cpp
aqjune: There are a few encodings that are unsupported at a callsite yet (nonnull/readnone/readonly). I think it might not be okay to show them in the printing?

aqjune: I left them as a TODO comment at the place, with a minor update to the handleAttributes (so the comment makes sense)

ir/attrs.h
aqjune: Conflicts resolved

aqjune:
The assertions revived, removing

aqjune:
Removed. BTW, I think it is good to have invariants as explicitly stated as assertions; while implementing this I was incorrectly calling addUB() at Memory::mkInput. When I have a bandwidth I'd like to come up with a patch that adds assertions which seems to be important to me.

llvm_util/llvm2alive.cpp
```diff
@@ -273,13 +273,32 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
     auto argI = fn->arg_begin(), argE = fn->arg_end();
+    const auto &alist = i.getAttributes();
+    uint64_t argidx = 0;
+
     for (auto &arg : args) {
       ParamAttrs attr;
+      // TODO: Once attributes at a call site are fully supported, we should
+      // call handleAttributes().
+
+      if (uint64_t derefb = alist.getParamDereferenceableBytes(argidx)) {
```
nunoplopes, line 284: why do you check for dereferenceable twice? once here and once below?

llvm_util/llvm2alive.cpp
```diff
@@ -818,10 +838,9 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
     return true;
   }
 
-  optional<ParamAttrs> handleAttributes(llvm::Argument &arg) {
+  optional<ParamAttrs> handleAttributes(const llvm::AttributeSet &aset) {
     ParamAttrs attrs;
-    for (auto &attr : arg.getParent()->getAttributes()
-                         .getParamAttributes(arg.getArgNo())) {
+    for (auto &attr : aset) {
```
nunoplopes, line 843: why this change?  Sorry, I'm a bit confused with the attributes. They can come from either the call instruction or from the declaration, but I think there's an API that gives the union of those?

llvm_util/llvm2alive.cpp
aqjune: It is because a dereferenceable tag can be at either the callsite or at a function signature.
At a call instruction:
```
BB:
  call void @f(i32* dereferenceable(4) %p)
```
At a function signature:
```
declare void @f(i32* dereferenceable(4) %ptr)
```

llvm_util/llvm2alive.cpp
aqjune: The change was made to make the comment that suggests a change to use handleAttribute() (at line 281) consistent with the implementation.
For the API that returns the union of those, I also think the API should be at somewhere because it would be pretty convenient, but I couldn't find it. :/

nunoplopes: see here: https://llvm.org/doxygen/Instructions_8cpp_source.html#l00321
`paramHasAttr(argno, attr`

aqjune: That was helpful for removing the big if statement at visitCallInst, thanks.
However I feel I'm still not at the best answer; the main obstacle was that dereferenceable bytes should have retrieved case by case (it is possible that there exists such functionality but again I couldn't find it again :/).
Anyway, I feel if we fully support readonly/readnone/etc attached at call sites, things should become simpler. But I think it is not in high priority..

aqjune: Or, to minimize diff, I can just remove this change (the change at the argument of hnadleAttributes). What do you think?

aqjune:
Rebase (resolve conflict at transform.cpp)

PR #378, opened 2020-05-13
Factor out FnAttrs class

aqjune:
This PR factors out function attributes as FnAttrs class.


aqjune:
To clarify the motivation: factoring out function attrs as a class seems to be helpful in the view of maintenance in the long run; there are other attributes that takes a parameter (align, dereferenceable_or_null) as well, and adding one field per attribute makes the code complex. For example, constructor of FnCall should take these as parameters otherwise.

PR #381, opened 2020-05-18
Encode call site refinement without using quantified variables

aqjune:
This patch removes introduction of quantified variables that were created for encoding the return values of functions when poisons were given as inputs. 

No regressions in Alive2 unit tests & LLVM unit tests
LLVM unit tests' timeout #: 836 -> 817

### DIFF
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -397,8 +397,11 @@ void State::mkAxioms(State &tgt) {
 
         expr refines(true);
         for (unsigned i = 0, e = ins.size(); i != e; ++i) {
-          refines &= ins[i].non_poison.implies(ins[i].value == ins2[i].value &&
-                                               ins2[i].non_poison);
+          string used = fn + "#arg" + to_string(i) + "used";
+          refines &=
+            expr::mkBoolVar(used.c_str()).implies(
+              ins[i].non_poison.implies(ins[i].value == ins2[i].value &&
+                                        ins2[i].non_poison));
         }
 
         if (refines.isFalse())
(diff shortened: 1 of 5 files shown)


nunoplopes:
Just had another thought about this patch and it may not be correct. This is because we need to support going from e.g.:
```
call f(poison)
call f(poison)
=>
call f(3)
call f(4)
```

The results of these calls may be non-poison. And thus we need to be able to refine each of the f(poison) calls independently to a different value.

I added 2 new tests here: https://github.com/AliveToolkit/alive2/commit/eb7e31e1d1a2318cd0c7dcafcd86fe7b33181931
Both current Alive2 & your patch fail to detect bugs in both test cases.

aqjune:
Probably tomorrow I'll be able to come up with my solution for this.
My observation is that some arguments of a function may not contribute to the result of a function call. This explains the conversion from poison argument to 3 and 4.

aqjune:
(Rebased to the master branch)
Commit 2966ed4 and e924fb8 is the new one.
Each argument may be either used argument or a dead argument.
If it is a dead argument, refinement of function inputs should not consider it.

To revisit the call-error5.opt:
```
%c1 = call i8 @f(i8 poison) nowrite noread                                         
%c2 = call i8 @f(i8 poison) nowrite noread                                         
%d = add i8 %c1, %c2                                                               
ret i8 %d                                                                          
  =>                                                                               
ret i8 1                                                                           
```

If f was `int f(int x) { return 1; }`, this is incorrect. With this patch, alive successfully raises an error.

```
call f(poison)
call f(poison)
=>
call f(3)
call f(4)
```
In this case, we are already encoding the return value of f() per call sites, so I think it is okay. call.src/tgt.ll has the example.

nunoplopes:
A function may read an argument conditionally. E.g. only if the value of a global value=1.
Is checking just the arguments enough? It's not obvious to me.

Your new test isn't sufficient to convince. We need negative tests, as the issues with the current code are around making the formula unsat too soon.
E.g. test that f(poison) + f(poison) => f(3) + f(4); assume(false)
(don't remember if this test is already there or not)

PR #382, opened 2020-05-19
Let memcmp return poison when comparing poisonous bytes

aqjune:
This fixes Transforms/MergeICmps/X86/multiple-blocks-does-work.ll .

### DIFF
--- a/tests/alive-tv/memory/memcmp-noub3.srctgt.ll
+++ b/tests/alive-tv/memory/memcmp-noub3.srctgt.ll
@@ -4,7 +4,7 @@ define i32 @src() {
   %p = alloca i32
   %p8 = bitcast i32* %p to i8*
   store i8 0, i8* %p8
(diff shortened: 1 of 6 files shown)


aqjune:
No regression found.

tests/alive-tv/memory/memcmp-noub3.srctgt.ll
```diff
@@ -0,0 +1,23 @@
+target datalayout = "e-p:64:64:64"
+
+define i32 @src() {
```
nunoplopes, line 3: isn't this test exactly equal to the previous one? Maybe you wanted to return poison or smth?

aqjune: Oh yeah, it was my mistake.
It was supposed to be the removed ub2 function.

Issue #384, opened 2020-05-21
Check function attributes in refinement


nunoplopes:
If we are verifying a function that was a `dereferenceable` attribute, we need to ensure that all returned pointers respect that constraint.
Similar for no return functions, etc, etc.

nunoplopes:
@aqjune do you have time to look into this, since you've been looking into attributes recently?

aqjune:
Ok, I'll look into it.

aqjune:
```
  enum Attribute { None = 0, NoRead = 1 << 0, NoWrite = 1 << 1,
                   ArgMemOnly = 1 << 2, NNaN = 1 << 3, NoReturn = 1 << 4,
                   Dereferenceable = 1 << 5, NonNull = 1 << 6,
                   NoFree = 1 << 7, NoUndef = 1 << 8 };
```

ArgMemOnly, NoRead, NoWrite seem nontrivial

nunoplopes:
see test in https://github.com/AliveToolkit/alive2/commit/ed7a09c697c8b40b4183c9aaffb0fe16a1a15eaf that should fail but doesn't ATM.

PR #388, opened 2020-05-24
Add support for dereferenceable attribute in function definition

aqjune:
This adds support for dereferenceable attribute in function definition (#384).


aqjune:
No additional regression in LLVM unit tests

Issue #289, opened 2020-02-10
Poison-ness of nonnull argument


aqjune:
```
define void @f(i8* nonnull %p) {                                                    
  ; Can %p be poison?
}                                                                                  
```
I think %p cannot be poison because it is UB for a caller to pass a poison value as a nonnull argument.
If it is well-defined for a caller to pass poison, the following transformation does not hold:
```
call @f(poison); // well-defined
->
call @f(null); // UB
```

regehr:
Hi Juneyoung, do you have a reference for "it is UB for a caller to pass a poison value as an argument"? I haven't seen this.

aqjune:
Hello. It meant it is UB to pass poison as a nonnull argument; sorry for the ambiguous expression.
The running example is here: https://godbolt.org/z/g8cgb3
The optimized program is directly passing null to g(nonnull), which is UB according to LangRef. Therefore the source program should raise UB too (which is passing poison to g()).

regehr:
Ah, thanks, I missed the nonnull attribute in your example!

If you have to prove non-poison to use the nonnull attribute, I expect that almost all uses of it will become unsound, since proving non-poison is generally hard. Is this your impression also?

aqjune:
Yes I think so; interprocedural analysis that adds nonnull flag will become unsound in general.
We cannot simply add freeze (e.g. %x -> freeze %x in the example above) as well, because %x should not be freezed into null.

PR #389, opened 2020-05-25
Add support for nonnull attribute

aqjune:
This adds support for nonnull attribute.

Resolves two failures:
```
Transforms/CorrelatedValuePropagation/basic.ll
Transforms/InstSimplify/compare.ll
```

This encodes  f(nonnull poison) as UB, because poison can be folded into null, making f(nonnull null), which is UB. 
Relevant issue: https://github.com/AliveToolkit/alive2/issues/289
There was a relevant thread in llvm-dev in the past, too.

This semantics may not explain a few existing optimizations. For example, we cannot say 'gep inbounds p, 1' is non-null.
But I think it is better to unsupport a few optimizations than breaking the general refinement from poison to a concrete value at this point...

### DIFF
--- a/ir/value.cpp
+++ b/ir/value.cpp
@@ -190,7 +190,9 @@ StateValue Input::toSMT(State &s) const {
     val = getType().mkInput(s, smt_name.c_str(), attrs);
   }
 
-  const bool has_deref = hasAttribute(ParamAttrs::Dereferenceable);
+  bool has_deref = hasAttribute(ParamAttrs::Dereferenceable);
+  bool has_nonnull = hasAttribute(ParamAttrs::NonNull);
+
   if (!config::disable_undef_input && !has_deref) {
     auto [undef, vars] = getType().mkUndefInput(s, attrs);
     for (auto &v : vars) {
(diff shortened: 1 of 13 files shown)


ir/value.cpp
```diff
@@ -199,10 +201,15 @@ StateValue Input::toSMT(State &s) const {
...
+  if (has_deref || has_nonnull) {
     Pointer p(s.getMemory(), val);
-    s.addAxiom(p.isDereferenceable(attrs.getDerefBytes(), bits_byte/8, false));
-    s.addAxiom(type == 0);
+    if (has_deref) {
+      s.addAxiom(type == 0);
+      s.addAxiom(p.isDereferenceable(attrs.getDerefBytes(), bits_byte/8, false));
+    }
+    if (has_nonnull && !has_deref) {
+      s.addAxiom(type.extract(1, 1) == 0);
```
nunoplopes, line 211: why allow nonnull arguments to be undef?
A function call with undef as argument is UB, so I think we can assume nonnull arguments aren't either poison or undef.

aqjune: Because an input can be partially undef with non-null value I think.
Memory::mkUndefInput has a nice way of encoding non-null undef value.

PR #391, opened 2020-05-28
Factor out Memory::PtrInput class, put assertions in State's methods

aqjune:
To resolve https://web.ist.utl.pt/nuno.lopes/alive2/index.php?hash=4beb2b117e2fdd2c&test=Transforms%2FSink%2Fcall.ll in the future, ptr_inputs should have readonly flag too (currently it only has byval, and ptr_input is represented as a pair of StateValue and bool).
This PR factors out it as a Memory::PtrInput so it will have more fields like byval, readonly, etc.

Also, I put a few assertions into State's methods which should fail if current_bb was not set.


aqjune:
Currently I have no plan for adding the readonly flag in the near future, but we don't know what's going to happen in the future, so I just made the change in my spare time.

Issue #385, opened 2020-05-21
Alive2 accepts a case which should fail


aqjune:
```
$ cat nondet-call.ll
define i1 @src() {
	%guard = freeze i1 undef
	br i1 %guard, label %COMPARE, label %EXIT

COMPARE:
	call void @g()
	ret i1 1

EXIT:
	ret i1 0
}

define i1 @tgt() {
	%guard = freeze i1 undef
	br i1 %guard, label %COMPARE, label %EXIT

COMPARE:
	ret i1 1

EXIT:
	ret i1 0
}

declare void @g()
```

This should fail because in source if 1 is returned g() should have been called, but in target g() is never called when 1 is returned.

nunoplopes:
tgt refines source. Just replace undef with false.
Please reopen if I'm mistaken.

PR #392, opened 2020-05-29
Memory and value refinement should be checked together if there's nondet. exec.

aqjune:
This resolves issue #385 , which happened because there was no entanglement between refinement of return value and refinement of final memory.
As the refinement formula becomes more complex, this causes slowdown in some cases; I added has_nondet_op which checks whether there is an operation that returns a nondeterministic value. If there is no nondet op, there is only a single execution given an input, so entanglement between the return value and final memory isn't needed.

Running LLVM unit tests...

### DIFF
--- a/tools/transform.cpp
+++ b/tools/transform.cpp
@@ -328,8 +328,9 @@ check_refinement(Errors &errs, Transform &t, State &src_state, State &tgt_state,
 
   auto src_mem = src_state.returnMemory();
   auto tgt_mem = tgt_state.returnMemory();
-  auto [memory_cnstr, ptr_refinement0] = src_mem.refined(tgt_mem, false);
+  auto [memory_cnstr0, ptr_refinement0] = src_mem.refined(tgt_mem, false);
   auto &ptr_refinement = ptr_refinement0;
+  auto memory_cnstr = value_cnstr && memory_cnstr0;
 
   if (check_expr(axioms_expr && (pre_src && pre_tgt)).isUnsat()) {
     errs.add("Precondition is always false", false);
--- a/tests/alive-tv/refinement/nondet-fail.srctgt.ll
+++ b/tests/alive-tv/refinement/nondet-fail.srctgt.ll
@@ -1,0 +1,26 @@
+define i1 @src() {
+	%guard = freeze i1 undef
+	br i1 %guard, label %COMPARE, label %EXIT
+
+COMPARE:
+	call void @g()
+	ret i1 1
+
+EXIT:
+	ret i1 0
+}
+
+define i1 @tgt() {
+	%guard = freeze i1 undef
+	br i1 %guard, label %COMPARE, label %EXIT
+
+COMPARE:
+	ret i1 1
+
+EXIT:
+	ret i1 0
+}
+
+declare void @g()
+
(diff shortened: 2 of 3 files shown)


aqjune:
No regression, timeout: 871 -> 870

nunoplopes:
What's the impact on running time if this is always enabled?

tools/transform.cpp
```diff
@@ -568,13 +591,22 @@ static void calculateAndInitConstants(Transform &t) {
                    isCast(ConversionOp::Ptr2Int, i)) {
           max_alloc_size = max_access_size = cur_max_gep = UINT64_MAX;
           has_int2ptr |= isCast(ConversionOp::Int2Ptr, i) != nullptr;
-          has_ptr2int |= isCast(ConversionOp::Ptr2Int, i) != nullptr;
+          if (auto *p2i = isCast(ConversionOp::Ptr2Int, i)) {
+            has_ptr2int = true;
+            // A nonlocal block has a fixed address.
+            has_nondet_op |= !returns_nonlocal(p2i->getValue(), true);
```
nunoplopes, line 597: non-local blocks don't have fixed addresses. They use an UF vs a variable for local blocks, but it's non-deterministic as well, no?

aqjune: Yes, in the twin semantics it is nondeterministic as well.
Here it is opt out because it does not generate quantified variables in that case. Updated the comment

aqjune:
Will leave the performance result when the test is done

aqjune:
For gzip, -smt-to=60 seconds: the number of function pairs:
master (aac8616):
```
#Correct:1720
#Incorrect:2
#Timeout:660
#EncodingTimeout:7
#PrecondFalse:472
#InvalidExpr:3062
#Unsupported:963
#OOM:411
```

this patch:

```
#Correct:1725
#Incorrect:2
#Timeout:652
#EncodingTimeout:3
#PrecondFalse:472
#InvalidExpr:3062
#Unsupported:963
#OOM:418
```

(note that total number of function pairs increased a bit after not using llvm-extract at all)
It raises OOM a bit, but overall it seems fine.

nunoplopes:
My question is what's the impact if the return refinement formula is always added to the memory constraint?
I'm asking because 1) that would rid of the code to find out whether it should be included, and 2) that code isn't enabling this patch in all required cases -- you left out some cases that you think aren't important, but we simply don't know.

aqjune:
Sorry, missed the 'always'. I will leave the result after the experiment.

aqjune:
Timeout of LLVM unit tests: 1123 -> 1157, gzip is still running

aqjune:
gzip:
```
#Correct:1706
#Incorrect:2
#Timeout:665
#EncodingTimeout:5
#PrecondFalse:472
#InvalidExpr:3062
#Unsupported:963
#OOM:422
```
Maybe the cost of double-checking value & memory isn't significant?
Anyway, it is good to make a patch smaller, so I'll leave the refinement part only

aqjune:
Total running time is actually reduced (maybe due to the early OOM?)
Sum of gzip's running time of all alive-tv processes:  38.177 -> 37.745

PR #394, opened 2020-05-30
Fix and extend dominators algorithm in function.h to support multi-root CFGs

kevindevos:
The algorithm originally implemented by Cooper, Keith D.; Harvey, Timothy J.; and Kennedy, Ken (2001) in the paper http://www.cs.rice.edu/~keith/EMBED/dom.pdf had a flaw in which they assumed that two different basic blocks always have a common parent which is not the case when dead code exists.
I only made a couple of changes to the algorithm added in the pull request https://github.com/AliveToolkit/alive2/pull/304 such that dominator trees for bb's that have paths to them from program entry are correct, however it is not possible to guarantee the same for the bbs that are not reachable from program entry. This should not affect the ub algorithm in https://github.com/AliveToolkit/alive2/pull/369.

The main changes are ignoring back-edges, making sure that every block without predecessors is its own dominator, and prioritise returning the predecessor on the tree with root on program entry when intersect is called on two subtrees of BB's with no common root.  
I tested this together with the changes in https://github.com/AliveToolkit/alive2/pull/369 and everything is working correctly.

### DIFF
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -306,7 +306,7 @@ void DomTree::buildDominators() {
     (void)instr;
     // skip back-edges
     visited_src.insert(&src);
-    if (visited_src.find(&tgt) == visited_src.end())
+    if (!visited_src.count(&tgt))
       doms.at(&tgt).preds.push_back(&doms.at(&src));
   }
 
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -293,6 +293,8 @@ void CFG::printDot(ostream &os) const {
 // Relies on Alive's top_sort run during llvm2alive conversion in order to
 // traverse the cfg in reverse postorder to build dominators.
 void DomTree::buildDominators() {
+  unordered_set<const BasicBlock*> visited_src;
+
   // initialization
   unsigned i = f.getBBs().size();
   for (auto &b : f.getBBs()) {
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ if (MSVC)
   set(CMAKE_EXE_LINKER_FLAGS         "/LTCG:INCREMENTAL ${CMAKE_EXE_LINKER_FLAGS}")
 else()
   set(CMAKE_CXX_FLAGS                "-Wall -Werror -march=native -fPIC ${CMAKE_CXX_FLAGS}")
-  set(CMAKE_CXX_FLAGS_DEBUG          "-g -O0 ${CMAKE_CXX_FLAGS_DEBUG}")
+  set(CMAKE_CXX_FLAGS_DEBUG          "-g -Og ${CMAKE_CXX_FLAGS_DEBUG}")
```
aqjune: I was using -O0 for debugging too. :/ LLDB did not properly show the contents of variables when compiled with -Og... I'll leave this as an issue

aqjune:
A random comment: I think it will be great if there is a test for domtree. For example, we can add an argument that dumps the dom tree into stdout & use CHECK: directive to see whether it works as expected.  @nunoplopes What do you think?

nunoplopes:
For future reference, here's an important test (Transforms/JumpThreading/unreachable-loops.ll):
```
define void @unreachable_multi_bbs_loop() {
bb:
  %tmp = call i32 @a()
  %tmp1 = icmp eq i32 %tmp, 1
  br i1 %tmp1, label %bb5, label %bb8

; unreachable two bbs loop.
bb3:
  br label %bb2

bb2:
  %tmp4 = icmp ne i32 %tmp, 1
  switch i1 %tmp4, label %bb3 [
    i1 0, label %bb5
    i1 1, label %bb8
  ]

bb5:
  %tmp6 = phi i1 [ %tmp1, %bb ], [ false, %bb2 ]
  br i1 %tmp6, label %bb8, label %bb7

bb7:
  br label %bb8

bb8:
  ret void
}
declare i32 @a()
```

PR #398, opened 2020-06-02
Add an option that encodes standard I/O functions as unknown fns

aqjune:
This adds `-io-nobuiltin` commandline option that encodes standard I/O functions as unknown.

InstCombine is the only optimization that transforms standard I/O libraries like printf/scanf/... into something else, so this flag can be turned on when the optimization does not have instcombine.

### DIFF
--- a/llvm_util/known_fns.cpp
+++ b/llvm_util/known_fns.cpp
@@ -5,6 +5,7 @@
 #include "llvm_util/utils.h"
 #include "ir/function.h"
 #include "ir/instr.h"
+#include "util/config.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/Analysis/MemoryBuiltins.h"
 #include "llvm/Analysis/TargetLibraryInfo.h"
@@ -15,6 +16,7 @@ using namespace std;
 
 #define RETURN_KNOWN(op)    return { op, true }
 #define RETURN_FAIL_KNOWN() return { nullptr, true }
+#define RETURN_FAIL_UNKNOWN() return { nullptr, false }
 
 namespace llvm_util {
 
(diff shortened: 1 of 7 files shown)


llvm_util/known_fns.cpp
```diff
@@ -40,7 +42,33 @@ known_call(llvm::CallInst &i, const llvm::TargetLibraryInfo &TLI,
   auto decl = i.getCalledFunction();
   llvm::LibFunc libfn;
   if (!decl || !TLI.getLibFunc(*decl, libfn) || !TLI.has(libfn))
-    return { nullptr, false };
+    RETURN_FAIL_UNKNOWN();
+
+  if (util::config::io_nobuiltin) {
+    switch (libfn) {
```
nunoplopes, line 48: libfn might used unitialized here

scripts/opt-alive.sh
```diff
@@ -10,15 +10,23 @@ set -e
 # attributor, functionattrs: inter procedural pass that deduces and/or propagates attributes
 # metarenamer: anonymizes function names
 PASSES="argpromotion deadargelim globalopt hotcoldsplit inline ipconstprop ipsccp mergefunc partial-inliner tbaa loop-extract extract-blocks safe-stack place-safepoints attributor functionattrs metarenamer lowertypetests extract-blocks openmpopt prune-eh -Os -Oz -O1 -O2 -O3"
+PASSES_SIMPLIFYLIB="-instcombine"
 
 TV="-tv"
-for p in $PASSES; do
-  for arg in $@; do
+IO_NOBUILTIN="-tv-io-nobuiltin"
```
nunoplopes, line 16: can we flip the default of this flag, such that if no flag is given there are no false positives. If no flag is given, io functions are marked as "unknown know function".
Only if the flag is given we mark them as regular functions. Then it's the user responsibility (or our script's) to ensure that the flag is only passed when the optimization doesn't take advantage of printf et al semantics.

nunoplopes: nevermind, this is what you do already. all good. sorry for the noise.

PR #401, opened 2020-06-03
Add a simplification rule to DisjointExpr

aqjune:
This reduces the time of a few oggenc's function pairs' preprocessing, from more than one hour to less than one minute.

### DIFF
--- a/smt/exprs.cpp
+++ b/smt/exprs.cpp
@@ -110,8 +110,14 @@ template<> DisjointExpr<expr>::DisjointExpr(const expr &e, bool unpack_ite,
         if (auto rhs_val = rhs.lookup(lhs_domain)) {
           add(lhs_v.concat(*rhs_val), c && lhs_domain);
         } else {
+          expr from, to(false);
+          if (!lhs_domain.isNot(from)) {
+            from = lhs_domain;
+            to = true;
+          }
+
           for (auto &[rhs_v, rhs_domain] : rhs) {
-            add(lhs_v.concat(rhs_v.subst(lhs_domain, true).simplify()),
+            add(lhs_v.concat(rhs_v.subst(from, to).simplify()),
                 c && lhs_domain && rhs_domain);
           }
         }


aqjune:
Attach pairs: 
[oggenc-slow.zip](https://github.com/AliveToolkit/alive2/files/4731136/oggenc-slow.zip)

nunoplopes:
This patch does not preserve semantics of the original code. See:
`lhs_v.concat(rhs_v)`
vs
`lhs_v.concat(rhs_v.subst(lhs_domain, true).simplify())`

I guess the speedup comes from here. Why is this a good change?

aqjune:
The speedup at preprocessing comes from removing `subst`, but `subst` certainly does some simplifications. I couldn't say my suggested change was strictly better, sorry for the confusion.

Instead, I left a simple optimization that utilizes lhs_domain: if lhs_domain = (not e), e can be replaced with false. This helps simplifying expressions because lhs_domain can be negation of something. This reduced example shows how it works: 
```
define i8 @f(* %p) {
%0:
  %q = gep * %p, 1 x i64 1
  %v = load i8, * %q, align 1
  ret i8 %v
}
```
Note that this does not resolve the preprocessing time issue of the attached pairs above.

BTW, I think this preprocessing issue is related with how GEP is encoded: it (1) splits bid/offset/attrs (2) updates the offset (3) concatenates these three. When the GEP-ed pointer is dereferenced, the expression is given to DisjointExpr, and if bid was DisjointExpr it tries to do the substitution.

nunoplopes:
where am I supposed to look with your example? I run `-se-verbose -smt-verbose` before & after the patch and the output is exactly the same.

aqjune:
Seems like simplification at the later pipeline folded it finally - I checked it by directly printing `rhs_v.subst(from, to)` just after my changes at DisjointExpr.

nunoplopes:
I get the same exprs before and after:
```
EXPR: (concat #b0
        (bvadd #x1
               (ite (= ((_ extract 0 0) ty_%p) #b0)
                    ((_ extract 3 0) %p)
                    undef!2)))
EXPR: (concat ((_ extract 4 4) %p) (bvadd #x1 ((_ extract 3 0) %p)))
EXPR: (concat #b0
        (bvadd #x1
               (ite (= ((_ extract 0 0) ty_%p) #b0)
                    ((_ extract 3 0) %p)
                    undef!3)))
EXPR: (concat ((_ extract 4 4) %p) (bvadd #x1 ((_ extract 3 0) %p)))
```

when running `alive-tv` on the following input:
```
define i8 @f(i8* %p) {
  %q = getelementptr i8, i8* %p, i32 1
  %v = load i8, i8* %q, align 1
  ret i8 %v
}
```

nunoplopes:
never mind, I see the result now; mixed the binaries before.

PR #404, opened 2020-06-05
Add undef vars that are registered to the stored value only

aqjune:
IIUC Memory::store's `undef_vars` argument takes undef variables that are used by the stored value only.

This PR makes it explicit.

Reduces preprocessing time of a few function pairs including [bzip2-slow.zip](https://github.com/AliveToolkit/alive2/files/4734074/bzip2-slow.zip)


aqjune:
Had a second thought about this, and undef vars that are used for pointer as well as control-flow should be considered too. Will update this

aqjune:
The undef parts seem pretty complex to me & easy to be wrong; I'd like to revisit this when I have enough bandwidth.

PR #405, opened 2020-06-05
Vector reductions: unary integral reductions

LebedevRI:
https://llvm.org/docs/LangRef.html#experimental-vector-reduction-intrinsics
> Horizontal reductions of vectors can be expressed using the
> following intrinsics. Each one takes a vector operand as an input
> and applies its respective operation across all elements of the vector,
> returning a single scalar result of the same element type.

Now that since https://reviews.llvm.org/D80867 /
https://github.com/llvm/llvm-project/commit/e50059f6b6b32c134e0f4f81281a774cb460524e
the (experimental) reductions are considered to be somewhat non-outcasts
and are produced by the middle-end, i feel like this is the perfect time
for alive2 to learn about them :)

The reductions can be classified into four groups:
* integral math/logic (this pr) - `add`/`mul` / `and`/`or`/`xor`
* integral aggregationh - `smax`/`smin`/`umax`/`umin`
* fp math - `fadd`/`fmul`
* fp aggregationh - `fmax`/`fmin`

As it can be seen, this patch adds support for the first group only.
Rest to follow.

I have considered not creating new `UnaryReductionOp`,
but using `UnaryOp`, but it ends up being really ugly because
these operations change the type of operand. In particular,
we'd need to spell them something like `reduce_add <2 x i8> %x to i8`,
and the type printing won't roundtrip.

Thoughts?

CC @rotateright, @topperc


aqjune:
Could you add a few tests that are written in LLVM IR, so we have a check that alive-tv can successfully parse bitcodes with the intrinsics?
It will also allow us to run LLVM unit tests as well. :)

LebedevRI:
That reminds me, some day i should bother to try to add code coverage build here..

LebedevRI:
> Could you add a few tests that are written in LLVM IR, so we have a check that alive-tv can successfully parse bitcodes with the intrinsics?

Tests added, thanks.

> It will also allow us to run LLVM unit tests as well. :)

Note that the llvm2alive support already was in place.

aqjune:
Oh yes, it was already there.
Regarding separation of UnaryReductionOp, I would prefer merging it into UnaryOp and not printing the destination type if it was a reduction op, but not strongly support it; I'll follow Nuno's opinion for this

LebedevRI:
> not printing the destination type if it was a reduction op

I'm not exactly sure what you mean.

Currently, the alive2 syntax for them is:
`%y = reduce_add <2 x i8> %x`, where `%y` becomes `i8`.

Problem 1: `UnaryOp::print()` would then print it as `%y = reduce_add i8 %x`,
which is obviously not what it was.
Explicitly spelling out the return type seems contrived - we don't do that, ever maybe?
Problem 2: if we change syntax to `%y = reduce_add i8 %x`, where `%y` should be  `<2 x i8>`,
well, that breaks parser because it doesn't know with what type to parse %x.

> I would prefer merging it into UnaryOp

So that is what i initially had, but basically in every function i needed a special if-block to handle these. It seems less contrived to separate them from the start. But i guess i'll follow @nunoplopes review :)

aqjune:
Maybe I was thinking about ConversionOp when talking about printing, which already had different types for its operand and return value. Anyway, now I somehow agree that UnaryOp and UnaryReductionOp can be separated because UnaryOp is returning the same type.

nunoplopes:
thank you!

LebedevRI:
Thanks! I'll followup with integral aggregations next.

PR #406, opened 2020-06-05
Let blockValRefined's expr::load stop looking into const initializer mem

aqjune:
I remember that this issue was mentioned before: this is my suggested solution.

expr::load can recursively call a load on a subexpression if it is store and known that the key does not match.
But, if there are many stores nested inside, the calls can go deep, causing long preprocessing time.
This can especially happen when there are many constant initializers such as string literals.

This PR adds an argument to expr::load so it stops simplification if the expression is structurally equivalent to it. This can be used at Memory:: blockValRefined, as this PR does.
When Memory::refined is called with skip_constants = true, looking through constant initializers is not necessary.

[bzip2-slow2.zip](https://github.com/AliveToolkit/alive2/files/4737974/bzip2-slow2.zip)

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -797,9 +797,10 @@ expr Pointer::fninputRefined(const Pointer &other, bool is_byval_arg) const {
 expr Pointer::blockValRefined(const Pointer &other) const {
   if (m.non_local_block_val.eq(other.m.non_local_block_val))
     return true;
-
-  Byte val(m, m.non_local_block_val.load(shortPtr()));
-  Byte val2(other.m, other.m.non_local_block_val.load(other.shortPtr()));
+  Byte val(m, m.non_local_block_val.load(
+      shortPtr(), &m.initial_non_local_block_val));
+  Byte val2(other.m, other.m.non_local_block_val.load(
+      other.shortPtr(), &other.m.initial_non_local_block_val));
 
   // refinement if offset had non-ptr value
   expr np1 = val.nonptrNonpoison();
@@ -1072,6 +1073,7 @@ Memory::Memory(State &state) : state(&state) {
     return;
 
   non_local_block_val = mk_block_val_array();
+  non_local_block_val_var = non_local_block_val;
   non_local_block_liveness = mk_liveness_array();
 
   // Non-local blocks cannot initially contain pointers to local blocks
@@ -1260,8 +1262,7 @@ Memory::mkCallState(const vector<PtrInput> *ptr_inputs, bool nofree) const {
 
   // TODO: handle havoc of local blocks
 
-  auto blk_val = mk_block_val_array();
-  st.block_val_var = expr::mkFreshVar("blk_val", blk_val);
+  st.block_val_var = expr::mkFreshVar("blk_val", non_local_block_val_var);
 
   {
     Pointer p(*this, "#idx", false);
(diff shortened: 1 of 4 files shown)


nunoplopes:
what's the speedup of this patch?

aqjune:
With -skip-smt, the result looks like following:
gzip:
```
Total Elapsed Time: 21.146 hrs // Sum of individual processes                              
Top 5 longest inputs:                                       
        328.607 9813.Early CSE w_ MemorySSA-unzip.txt        
        320.535 9814.Jump Threading-unzip.txt                
        281.781 9815.Value Propagation-unzip.txt             
        272.994 12250.Loop Vectorization-inflate_dynamic.txt 
        272.287 12265.Unroll loops-inflate_dynamic.txt       
=>
Total Elapsed Time: 18.637 hrs                              
Top 5 longest inputs:                                      
        263.732 9813.Early CSE w_ MemorySSA-unzip.txt       
        261.122 9814.Jump Threading-unzip.txt               
        246.258 526.SROA-inflate_codes.txt                  
        241.196 245.Early CSE-inflate_codes.txt             
        227.365 9836.Rotate Loops-unzip.txt                 
```
Interestingly, bzip2 had more speedup:
bzip2:
```
Total Elapsed Time: 24.510 hrs                        
Top 5 longest inputs:                                
        1213.930 9728.Loop Vectorization-compress.txt 
        1210.740 9773.Unroll loops-compress.txt       
        1210.690 9860.Loop Vectorization-compress.txt 
        1184.630 9951.Loop Vectorization-compress.txt 
        1170.020 10041.Loop Vectorization-compress.txt
 =>
Total Elapsed Time: 14.489 hrs                       
Top 5 longest inputs:                               
        517.992 10535.Loop Vectorization-compress.txt
        513.193 10256.Loop Vectorization-compress.txt
        509.034 9860.Loop Vectorization-compress.txt 
        508.698 9951.Loop Vectorization-compress.txt 
        506.987 10041.Loop Vectorization-compress.txt
```

nunoplopes:
that speedup in bzip2 is really surprising.. Either it has tons of constants or the AST traversal is unnecessarily slow. And these are just the loads for refinement; there are also the loads of the program..
Are we sure all the initializers there are needed? How many bytes are initialized in these benchmarks?
I'm not fully against this patch, but I think we need to understand the root cause of the issue, as solving it may bring more benefits than just this patch alone.

aqjune:
(Rebased to resolve conflicts)
For the attached pair, it loads 657 bytes, which isn't very large.
I also think the traversal is slow.
mkAxiom for function call encoding is the place where it frequently takes long time - The program has 43 function pairs that finally calls addPre()... A few of those are standard library fns like fprintf/exit/, I wonder whether their encoding can be simplified rather than just assuming as an unknown function..

aqjune:
I see a few multiplication factors that may result in calling more expr::load :
Memory::refined has a loop over bids, which calls blockRefined multiple times. Pointer::blockRefined again calls Pointer::blockValRefined for the number of the block size. if it decided to unroll byte offsets. However disabling these would make SMT queries more complicated, so I'm not sure whether just disabling these unrolls is the answer...

aqjune:
The reason was that initial_non_local_block_val did not match anymore after blk_val is replaced with the newer one.
CallState should have tracked what the new initial_non_local_block_val is.
Updated it (& rebased) to resolve the issue.

aqjune:
LLVM unit tests: no regression
bzip2 preprocessing time: no speedup. Maybe removing the initializers were too powerful :)
gzip preprocessing time: 25.159 -> 15.284 hrs

aqjune:
Stale branch, close this

PR #409, opened 2020-06-06
set build RPATH so we can find z3 in non-standard locations

regehr:
so forever I have not been able to get alive2 to find libz3.so (at runtime, not cmake time) unless that library is in a standard location like /usr/lib. recently I had to fix similar issues in souper, so figured I'd fix them here as well. this tells CMake to set the RPATH of build products so that libraries can be found at runtime.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,9 @@ endif()
 
 message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
 
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON)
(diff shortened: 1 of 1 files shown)


regehr:
to see the effect of this patch, use a command like the one below (this is linux-specific). without this patch the path to the z3 library is not part of RUNPATH

```
$ objdump -x alive-tv | grep -i path
  RUNPATH              $ORIGIN/../lib:/home/regehr/llvm-install/lib:/home/regehr/souper-regehr/third_party/z3-install/lib
$ 
```

PR #411, opened 2020-06-06
Vector reductions: unary integral aggregation reductions

LebedevRI:
Continuation of https://github.com/AliveToolkit/alive2/pull/405

https://llvm.org/docs/LangRef.html#experimental-vector-reduction-intrinsics
> Horizontal reductions of vectors can be expressed using the
> following intrinsics. Each one takes a vector operand as an input
> and applies its respective operation across all elements of the vector,
> returning a single scalar result of the same element type.

Here, i'm adding alive2 support for all the
vector integral aggregation reductions,
namely `smax`/`smin`/`umax`/`umin`.

CC @rotateright, @topperc


LebedevRI:
Thank you!
Only fp reductions are left; i'm not great with fp, but i'll try to take a look..

Issue #412, opened 2020-06-07
dereferenceable attribute does not guarantee that the pointer is never freed


aqjune:
According to LangRef's nofree attribute and https://reviews.llvm.org/D49165 , dereferenceable attribute cannot not guarantee that the pointer is unfreed inside the function body.
This implies that tests/alive-tv/attrs/dereferenceable-callsite.srctgt.ll should fail .
Found this issue while encoding lifetime of escaped local blocks.
Seems to be related with lifetime update after function call. Will have a look.

nunoplopes:
That's not my reading. It says that the pointer is not freed when the function starts, and then it can be free'd by some function call. This corresponds to what we implement AFAICT. Do you agree?

aqjune:
What I understood from LangRef was something like this.

```
// https://github.com/AliveToolkit/alive2/blob/master/tests/alive-tv/attrs/dereferenceable-callsite.srctgt.ll
define i32 @src(i1 %c, i32* %p) {
ENTRY:
  call void @f(i32* dereferenceable(4) %p)
  br i1 %c, label %A, label %EXIT
A:
  %v1 = load i32, i32* %p
  br label %EXIT
EXIT:
  %val = phi i32 [%v1, %A], [0, %ENTRY]
  ret i32 %val
}

define i32 @tgt(i1 %c, i32* %p) {
  call void @f(i32* dereferenceable(4) %p)
  %v1 = load i32, i32* %p
  %val = select i1 %c, i32 %v1, i32 0
  ret i32 %val
}
```

f() can be possibly something like
```
void f(p) { free(p); }
```
So speculating load unconditionally is still disallowed.
If f() has `nofree`, f(p) will never free p, so this optimization is now allowed.

nunoplopes:
Ah, for function calls, right. I agree; our implementation is not correct then.
Can you fix this & maybe add support for the notfree attribute? Thanks!

aqjune:
Given this example, 
[callsite-split.srctgt.ll.zip](https://github.com/AliveToolkit/alive2/files/4742842/callsite-split.srctgt.ll.zip)

Alive2 master fails with minor modification:
If Pointer::getAllocType has no optimizations (in other words, if it has only the last getValue call), value mismatch happens.
I'll have a look about this tomorrow.

PR #413, opened 2020-06-07
Add nofree attribute

aqjune:
This addresses #412.

Running this with LLVM unit test raises two failures, but this seems to be a pre-existing bug in Alive2. I'll write down the details at #412.


nunoplopes:
Sorry, we got a conflict :S
Can you also add code to FnCall pls? it's UB for a nofree function to call a non-nofree function.

aqjune:
I encoded it as 'addUB(livenesses before the call == livenesses after the call)' at State::addFnCall.
Does this seem to make sense to you? A non-nofree function call may not deallocate blocks, in that case it is safe to call it.

PR #415, opened 2020-06-08
PtrType::refines should use returnMemory()

aqjune:
PtrType::refines should use returnMemory().
Found this bug while investigating https://github.com/AliveToolkit/alive2/issues/412#issuecomment-640274035

Running LLVM unit tests to see its impact


aqjune:
\# of failures: no regression

Timeout: 1167 -> 1166
No change other than this

PR #417, opened 2020-06-10
add memory model tests; these are parameterized by block size, not bitwidth

regehr:
not sure these are great, since mostly the new memory tests either fail at 11 (falling off the unroll limit of 10, I guess) or go all the way to 256. I'd be happy for feedback from Juneyoung or Nuno

### DIFF
--- a/scripts/perf-testing/negative/memcpy1.ll
+++ b/scripts/perf-testing/negative/memcpy1.ll
@@ -1,0 +1,17 @@
+declare void @llvm.memcpy.p0i8.p0i8.i32(i8*, i8*, i32, i1)
+declare i32 @memcmp(i8* nocapture, i8* nocapture, i64)
+
+define i32 @src(i8*) {
+  %p1 = alloca [X x i8]
+  %p2 = bitcast [X x i8]* %p1 to i8*
+  call void @llvm.memcpy.p0i8.p0i8.i32(i8* %p2, i8* %0, i32 X, i1 0)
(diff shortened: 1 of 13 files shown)


aqjune:
I think the performance of memcmp can be assessed independently from memcpy and load/stores: it may emit a large & complex expression, so it can hide the cost of other memory operations like memcpy or load.
Also, the unroll count of memcmp is set to 10, so after 10 the results may be less interesting.

I'd like to suggest having tests that consist of memcpy+store+load+(gep,..). It may give more meaningful metrics.
For this (modified) example: [memcpy1.ll.zip](https://github.com/AliveToolkit/alive2/files/4763166/memcpy1.ll.zip)
test.pl printed 34 on my laptop, and I think this metric is meaningful because it implies that verifying a bitcode pair that has struct copy of larger than 34 bytes from arg to local may take a long time. (BTW, I wonder how the number 34 came out; someday I really need to study how Z3 internal works. :) )

aqjune:
Also, the encoding of memory operations differs on whether pointer addresses are observed or not. To enforce it, simply adding ptrtoint on any pointer will work.

nunoplopes:
Right what Juneyoung mention is important: we have a lot of optimizations for the common case, but if something doesn't hold, the encoding becomes much heavier. Escaping pointers breaks some optimizations. Another is having memory accesses of both ints & pointers (otherwise we optimize the memory to contain just the right type).
Throwing function calls into the mix is also explosive :)

regehr:
this is super helpful, thanks guys!

PR #400, opened 2020-06-03
Encode escaped local block's values after fn call

aqjune:
This is a simple start to encode escaped local blocks.

Running unit tests + gzip/bzip2...

PR #420, opened 2020-06-11
Escape pointers that are given to fn args

aqjune:
This is factored out from #400 to simplify diff & reduce conflicts that may happen in another working PR.

This simply escapes local blocks that are given to fn calls, if its block id is straightforward only.

This also makes llvm2alive add nocapture if the fn arg has the attribute, too.

Always thank for your review!


aqjune:

aqjune:
bzip2:
```
Total Elapsed Time: 27.546 hrs
#Correct:934                  
#Incorrect:4                  
#Timeout:612                  
#EncodingTimeout:1            
#PrecondFalse:468             
#InvalidExpr:848              
#Unsupported:1027             
#OOM:309                      
=>
Total Elapsed Time: 27.739 hrs
#Correct:933
#Incorrect:4
#Timeout:615
#EncodingTimeout:0
#PrecondFalse:468
#InvalidExpr:848
#Unsupported:1027
#OOM:308
```

gzip:
```
Total Elapsed Time: 40.743 hrs
#Correct:2078                 
#Incorrect:2                  
#Timeout:720                  
#EncodingTimeout:2            
#PrecondFalse:473             
#InvalidExpr:2604             
#Unsupported:963              
#OOM:455                      
=>
Total Elapsed Time: 40.898 hrs
#Correct:2074
#Incorrect:2
#Timeout:727
#EncodingTimeout:1
#PrecondFalse:473
#InvalidExpr:2604
#Unsupported:963
#OOM:453
```
No regression in # incorrect, minor increase in timeout.

ir/state.cpp
```diff
@@ -241,6 +241,20 @@ State::addFnCall(const string &name, vector<StateValue> &&inputs,
...
+    if (!v.nocapture && v.val.non_poison.isTrue()) {
+      Pointer p(memory, v.val.value);
+      if (!p.isLocal().isTrue())
+        continue;
+      // TODO: this should support more complex cases, such as (c ? bid1: bid2).
+      // Current implementation of extract_possible_local_bids cannot be used
+      // because it causes false positive for an input in gzip.
+      uint64_t short_bid;
+      if (p.getShortBid().isUInt(short_bid))
+        memory.escapeLocal(short_bid);
```
nunoplopes, line 254: I really don't like this logic in this file. It should be moved to memory.cpp, e.g. Memory::escapePtr(StateValue&).
If extract_possible_local_bid is not good enough, but we should have a look into why as well. It only needs to be as good as LLVM's constant folding. (this can go into another PR)

ir/state.cpp
```diff
@@ -241,6 +241,20 @@ State::addFnCall(const string &name, vector<StateValue> &&inputs,
     return vector<StateValue>(out_types.size());
   }
 
+  for (auto &v : ptr_inputs) {
+    if (!v.nocapture && v.val.non_poison.isTrue()) {
```
nunoplopes, line 245: can this be changed to `!isFalse()`?  Would be better for false negatives.

ir/state.cpp
aqjune: The failure comes from a code like this:
```
p = alloca
use p ; but never escape p

q = load * glb ; as p is never escaped, q cannot be p
f(q) ; q cannot be p, but current extract_possible_local_bid cannot catch it
```

To do this, we need to track what 'worst case' is & let extract_possible_local_bid get the worst case rather than simply marking all bits as failing.

I'll work on this in a separate PR.

aqjune: For this PR, I'll reflect your comment.

aqjune:
Regarding the follow-up patch, I was a bit busy today, but I'll be able to make it by noon there

PR #421, opened 2020-06-12
non_local_block_liveness should be preserved if nofree exists

aqjune:
This caused invalidexpr errors after #398 and nofree are landed because many io functions are labeled as nofree.


nunoplopes:
ah damn, my fault. introduced this in your patch-- gah!

PR #422, opened 2020-06-13
Let fninputRefined see bytes of local blocks

aqjune:
This PR makes fninputRefined see bytes of local blocks.
This reproduces two bugzilla reports. Note that `pr41949-2.srctgt.ll` is the original version of `pr41949.srctgt.ll` that used function call to observe the byte contents.

This starts with seeing integer bytes of the escaped blocks.
I believe that this will provide a good baseline metrics; any encoding that is more complex than this should raise more timeouts.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -771,35 +771,36 @@ expr Pointer::fninputRefined(const Pointer &other, bool is_byval_arg) const {
   expr size2 = other.blockSize();
   expr off2 = other.getOffsetSizet();
 
-  expr local
-    = expr::mkIf(isHeapAllocated(),
-                 other.isHeapAllocated() && off == off2 && size2.uge(size),
+  expr local =
+      (other.isLocal() || other.isByval() || is_byval_arg) &&
+      isHeapAllocated().implies(other.isHeapAllocated() && off == off2);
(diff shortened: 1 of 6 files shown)


ir/memory.cpp
```diff
@@ -771,35 +771,36 @@ expr Pointer::fninputRefined(const Pointer &other, bool is_byval_arg) const {
...
+  expr local =
+      (other.isLocal() || other.isByval() || is_byval_arg) &&
+      isHeapAllocated().implies(other.isHeapAllocated() && off == off2);
+  bool iswrite = !is_byval_arg;
 
-  // TODO: this induces an infinite loop
-  // block_refined(other);
+  expr ofs = expr::mkFreshVar("fninput_ofs", expr::mkUInt(0, bits_for_offset));
+  Pointer this_ofs = *this + ofs;
+  Pointer other_ofs = other + ofs;
```
aqjune, line 781: This had to be `+ ofs` because `*this` and `other` may have different offsets.

aqjune:
LLVM unit tests: no regression, interestingly 5 timeouts decreased

aqjune:
Resolve conflict

aqjune:
#428 contains this patch

PR #425, opened 2020-06-13
Another fix to dominators to ensure progress in some cases

kevindevos:
One of the functions in the oggenc benchmark resulted in an infinite loop in the intersect function with:

``` 
(gdb) p f1->order
$2 = 21
(gdb) p f2->order
$3 = 19
(gdb) p f1->dominator->order
$4 = 22
(gdb) p f2->dominator->order
$5 = 19
(gdb) p f1->dominator->dominator->order
$6 = 22
```
Because `f2->order < f1->order` and `f2->dominator == f2` nothing will change in each iteration. It is another side effect of multiple root CFG's where the order here loses meaning. This change should ensure that progress is made in these cases. Tested with my ub implementation and everything looks correct.


nunoplopes:
not needed anymore as we now preprocess functions to remove unreachable BBs.

PR #426, opened 2020-06-15
Do not escape local block ids if ptr is loaded from initial mem

aqjune:


aqjune:

ir/memory.cpp
```diff
@@ -1706,8 +1706,16 @@ expr Memory::checkNocapture() const {
 }
 
 void Memory::escapeLocalPtr(const expr &ptr) {
+  auto bids = extract_possible_local_bids(*this, ptr);
   uint64_t bid;
-  for (const auto &bid_expr : extract_possible_local_bids(*this, ptr)) {
+  unsigned hi, lo;
+  expr sel, blk, idx;
+  if (bids.size() == 1 && bids[0].isExtract(sel, hi, lo) &&
```
nunoplopes, line 1713: is there any reason to restrict this to size() == 1?  This rule seems to be applicable with any set of bids, no?

aqjune: I was considering the escape2.srctgt.ll case only, but it makes sense. Updated

PR #427, opened 2020-06-15
Phi - Build smaller path formulas that distinguish each value for each phi instruction.

kevindevos:
wip


kevindevos:
wip - requires merging + simplifying

kevindevos:
With newer versions of Alive and LLVM this PR no longer brings significant performance improvements sadly

PR #428, opened 2020-06-17
Implement mapping between local blocks escaped via function calls

aqjune:
This implements mapping between local blocks escaped via function args.

Mapping is carried as an LocalBlkMap class object, and only target memory maintains it.
- LocalBlkMap uses a bitvector to check whether there is a mapping from target local bid to the source local bid.
- LocalBlkMap uses a lambda function to get the mapped source bid from target bid.
However, mapping should appear as `forall tgt, exists src, mapping, ...` in the refinement, and we cannot quantify lambda. To resolve the issue, each fresh variable that represent the src local bid stored to the lambda function is quantified, and the lambda is expanded by expr::load whenever used.

Using this mapping, the input/output memory of src/tgt function calls are related (as discussed before).
Memory::refined now checks mapped local blocks of src/tgt, if check_local is set. Pointer::blockRefined, Pointer::blockValRefined takes the check_local as arguments as well.
Memory::mkCallState is the place where the mapping is updated.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2044,6 +2044,7 @@ StateValue Return::toSMT(State &s) const {
   auto &retval = s[*val];
   s.addUB(s.getMemory().checkNocapture());
   addUBForNoCaptureRet(s, retval, val->getType());
+  s.getMemory().markAllocasAsDead();
 
   auto &attrs = s.getFn().getFnAttrs();
   bool isDeref = attrs.has(FnAttrs::Dereferenceable);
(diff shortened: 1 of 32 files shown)


aqjune:
LLVM unit tests: no regression. Timeout: 1151 (no change)

aqjune:
gzip, bzip2: [result.txt](https://github.com/AliveToolkit/alive2/files/4807154/result.txt)

aqjune:
[malloc-poison.srctgt.ll.zip](https://github.com/AliveToolkit/alive2/files/4812047/malloc-poison.srctgt.ll.zip) Investigating the reason

aqjune:
Stale branch, will open a PR with newer version

PR #431, opened 2020-06-19
zero_bits_offset should return zero if addresses observed

aqjune:
I think zero_bits_offset should be fixed because bits_byte can be larger than 8 if all loads/stores have >1 alignments whereas the pointers may have offsets that have non-multiple of align.


aqjune:
Running unit tests

aqjune:
No faillure added, timeout: 1151 -> 1155

PR #432, opened 2020-06-19
Compare msb + nonpoison bits at once when converting bytes to value

aqjune:
When loading a non-pointer value from bytes, we should check whether there is no poison byte & no pointer byte.
If bits_int_poison is 1, this can be done efficiently by extracting and checking relevant bits from Byte at once 
This can also be applied when bits_int_poison > 1, but requires a bigger change.

[oggenc.srctgt.zip](https://github.com/AliveToolkit/alive2/files/4804601/oggenc.srctgt.zip)
Running -se-verbose from this pair will show the diff.

Moved Type.combine_poison to Memory.cpp's static function because it was only used there.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -179,6 +179,10 @@ expr Byte::nonptrValue() const {
   return p.extract(start + bits_byte - 1, start);
 }
 
+expr Byte::isNonptrNonpoison() const {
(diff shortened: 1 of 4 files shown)


nunoplopes:
what's the speedup?

aqjune:
(running tests)

aqjune:
No noticeable improvement in either # of correctness or vcgen time for gzip/bzip2/oggenc, which was unexpected :/ I'll close this

PR #434, opened 2020-06-21
Use pattern-match to simplify at DisjointExpr for faster vcgen

aqjune:
When DisjointExpr explodes an expression, it uses expr::subst() to simplify a subexpression using domain. But, as we're exploding SMT formula for pointers only, a simple pattern-match for the expression and making simplified expressions works too.

Has a nice speedup of vcgen time at oggenc, no regression in # correct and other benchmarks' vcgen time.
Result of bzip2/gzip/oggenc: [result.txt](https://github.com/AliveToolkit/alive2/files/4809321/result.txt)

### DIFF
--- a/smt/exprs.cpp
+++ b/smt/exprs.cpp
@@ -82,6 +82,26 @@ ostream &operator<<(ostream &os, const OrExpr &e) {
   return os << e();
 }
 
+static expr subst_for_ptr(const expr &e, const expr &from, const expr &to) {
+  // If a pointer p is `gep p0, i`:
+  //  p = (extract-bid ptr0) ++ (bvadd (extract-offset ptr0) i)
+  //
+  // p0 again may be loaded from the byte b:
+  //   p = (extract-bid (ite b.isPtr b.ptrvalue null))
+  //       ++ (bvadd (extract-offset (ite b.isPtr b.ptrvalue null)) i)
+  expr a, b, cond, then, els;
+  unsigned hi, lo;
+
+  if (e.isAdd(a, b))
+    return subst_for_ptr(a, from, to) + subst_for_ptr(b, from, to);
+  else if (e.isIf(cond, then, els) && cond.eq(from))
+    // Don't look further
+    return expr::mkIf(to, then, els);
+  else if (e.isExtract(a, hi, lo))
+    return subst_for_ptr(a, from, to).extract(hi, lo);
+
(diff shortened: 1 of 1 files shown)


aqjune:
+2 incorrect pairs in oggenc are known failures (so it is okay)

smt/exprs.cpp
```diff
@@ -117,7 +137,7 @@ template<> DisjointExpr<expr>::DisjointExpr(const expr &e, bool unpack_ite,
           }
 
           for (auto &[rhs_v, rhs_domain] : rhs) {
-            add(lhs_v.concat(rhs_v.subst(from, to).simplify()),
+            add(lhs_v.concat(subst_for_ptr(rhs_v, from, to).simplify()),
```
nunoplopes, line 140: is this semantically equivalent as the previous code?
can you do an `assert(rhs_v.subst(from, to).eq(subst_for_ptr(rhs_v, from, to)))` and run over the tests to see if it fails?

aqjune: No problem. I couldn't run it yet (had a few things to do), but will be able to run on single file benchmarks after then.

aqjune:
I tried the assert eq check, but it failed because subst() can reorder operands:
```
----- after subst_for_ptr:
(bvadd (ite (= #b0 ((_ extract 0 0) ty_%call1)) %call1 undef!23)
       #x0000000000000001
       (ite (= #b0 ((_ extract 0 0) ty_%call)) %call undef!22))
----- after subst:
(bvadd #x0000000000000001
       (ite (= #b0 ((_ extract 0 0) ty_%call)) %call undef!22)
       (ite (= #b0 ((_ extract 0 0) ty_%call1)) %call1 undef!23))
```

Instead, I made it look into operands of ite as well, because it converted a few timeouts to corrects in oggenc.

aqjune:
As the encoding of blocks has changed, reevaluation is needed (will do when I have time later)

aqjune:
[results-disjexpr.zip](https://github.com/AliveToolkit/alive2/files/5020681/results-disjexpr.zip)
Baseline: 7706ee4 (yesterday)
IMO this patch is quite effective for shortening vcgen of bzip2 and oggenc. # of correct/incorrect/timeout's change seems to be within errors.
BTW, there are a few new incorrect pairs; I'll reduce them.

PR #436, opened 2020-06-22
Optimize nocapture encoding if only constant ptrs are stored

aqjune:
This is a simple patch that makes Memory::checkNocapture() return simply true if only constant pointers (null, global vars, undef) are stored into memory.

### DIFF
--- a/ir/globals.h
+++ b/ir/globals.h
@@ -78,7 +78,9 @@ extern bool has_null_block;
 /// Whether the programs do memory accesses that load/store int/ptrs
 extern bool does_int_mem_access;
 extern bool does_ptr_mem_access;
+
 extern bool does_ptr_store;
+extern bool does_ptr_store_constantsonly; // stores non-constant ptrs only?
 
 extern unsigned heap_block_alignment;
 
--- a/ir/globals.h
+++ b/ir/globals.h
@@ -80,7 +80,7 @@ extern bool does_int_mem_access;
 extern bool does_ptr_mem_access;
 
 extern bool does_ptr_store;
-extern bool does_ptr_store_constantsonly; // stores non-constant ptrs only?
+extern bool does_ptr_store_nonconstant; // may store non-constant ptrs?
 
 extern unsigned heap_block_alignment;
 
--- a/ir/globals.cpp
+++ b/ir/globals.cpp
@@ -42,7 +42,7 @@ bool has_null_block;
 bool does_int_mem_access;
 bool does_ptr_mem_access;
 bool does_ptr_store;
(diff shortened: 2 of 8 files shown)


ir/globals.h
```diff
@@ -72,7 +72,14 @@ extern bool has_null_block;
 /// Whether the programs do memory accesses that load/store int/ptrs
 extern bool does_int_mem_access;
 extern bool does_ptr_mem_access;
-extern bool does_ptr_store;
+
+struct PtrStoreInfo {
+  bool has_store;
+  bool constant_ptrs_only;
+  operator bool() const;
```
nunoplopes, line 79: you can inline this definition in the header

ir/globals.h
```diff
@@ -72,7 +72,14 @@ extern bool has_null_block;
 /// Whether the programs do memory accesses that load/store int/ptrs
 extern bool does_int_mem_access;
 extern bool does_ptr_mem_access;
-extern bool does_ptr_store;
+
+struct PtrStoreInfo {
```
nunoplopes, line 76: why this complication (vs adding another bool)?
Is there other info you expect to collect from stored ptrs?

aqjune: Yes. I'm going to collect whether there is any local pointer that is stored, so a struct was used here.

aqjune: I'm not strongly in favor of having structs here, so fixed it to be const bools instead

aqjune:
Ready to get review again

nunoplopes:
is this any useful in practice? nocapture is not common at all.
Unless there's some interesting case, i would suggest we drop this patch and focus efforts on the new memory branch instead (after November).

aqjune:
I think nocapture isn't rare - In case of sqlite3, among 1700 functions, about 500 functions have pointer arguments that are labeled as nocapture.
Storing a null pointer to somewhere seems to happen quite frequently, and attributes are attached at the early part of optimization pipeline, so I think this is helpful.

aqjune:
LLVM unit test has improvement (1145 -> 1139 timeouts); maybe it is related to that nocapture was added to Alive2 because it appeared top from the list of unsupported features.

nunoplopes:
I suggest we do pencils down now, and focus on the paper submission. We go back to memory afterwards.
My impression is that this patch (and a few other optimizations that we have currently) are not needed with the new memory branch. I can show you why later.

aqjune:
Aha.. I see your point, capture analysis can be more precisely done in the new memory version.

nunoplopes:
Right 😀

PR #437, opened 2020-06-23
fix bugs regarding *alloc functions

aqjune:
Memory::alloc has this condition:
```
  state->addPre(nonnull.implies(precond && nooverflow));
```
But, `precond` can be statically false (e.g. the size is poison), and then Z3 can find a proper `nonnull` that makes src always have false precondition. This leads to reporting `Source is more defined than target`.

To be specific, Malloc::toSMT gives `np_size` to `non_poison`, and `mkBoolVar("malloc_never_fails")` to `nonnull`. If the size is poison, this precondition becomes `malloc_never_fails.implies(false)`. Z3 can assign `true` to `malloc_never_fails` and make refinement always false (which is `exists tgt, pre(tgt) /\ forall src, pre(src) -> refines(tgt, src)`).

Removing this condition works for bzip2,gzip,oggenc (~sqlite3 is running~ sqlite3 is also fine), but requires updating alloca to check signed overflow ~rather than~as well as unsigned overflow due to Transforms/InstCombine/2011-06-13-nsw-alloca.ll in LLVM unit tests.


aqjune:
Removing this also required realloc to consider poison-ness of the new size as well; If the new size is poison, Memory::alloc was simply giving up allocation, causing subsequent memcpy/free to give up dereferenceability check. So, `realloc(ptr, poison) -> realloc(ptr, non-poison)` failed.

aqjune:
Passes LLVM unit tests.
Three `*alloc-poison.srctgt.ll` are reduced from the single file benchmarks; they fail without this patch.
`alloca-array.srctgt.ll` fails without the new addUB at Alloc::toSMT.

nunoplopes:
I think we may have a fundamental flaw in how the precondition is used to assemble the refinement constraints. We may need to switch quantifiers for src's Pre.
Right now we do (primed vars = tgt):
`\exists I, N', O . \forall N . Pre'(I, N') /\ P'(I, N, O) /\ (\not Pre(I, N) \/ \not P(I, N, O))`

this may make more sense instead:
`\exists I, N', N'', O . \forall N . Pre'(I, N') /\ P'(I, N, O) /\ (\not Pre(I, N'') \/ \not P(I, N, O))`

Meaning: it's ok for src/tgt to mismatch if there's some non-determinism that makes the pre fail. Rather than the pre must fail for all non-det as above.
I need to think again about this. I'm unsure what's the right proof obligation.

aqjune:
In this case, would encoding disjointness of local blocks matter?
`Pre` is the disjointness of local blocks in src, and `N''` is the addresses of src's local blocks, then Z3 can find some addresses of local blocks that are overlapping so `Pre` would be always false.

ir/instr.cpp
```diff
@@ -2311,17 +2311,18 @@ void Alloc::print(std::ostream &os) const {
...
   if (mul) {
     auto &[mul_e, mul_np] = s[*mul];
     s.addUB(mul_np);
-    sz = sz.zextOrTrunc(bits_size_t);
     auto m = mul_e.zextOrTrunc(bits_size_t);
     s.addUB(sz.mul_no_uoverflow(m));
     sz = sz * m;
   }
 
+  s.addUB(sz.extract(bits_size_t - 1, bits_size_t - 1) == 0);
```
nunoplopes, line 2325: this is not correct. There's no word in the standard about overflow being UB. And now that I think about this, I don't like the existing code that a few lines above..

nunoplopes:
> In this case, would encoding disjointness of local blocks matter?
> `Pre` is the disjointness of local blocks in src, and `N''` is the addresses of src's local blocks, then Z3 can find some addresses of local blocks that are overlapping so `Pre` would be always false.

I need to think about this tmr fresh.

regehr:
Nuno, I have no idea if it's related, but @manasij7479 has an example where a LHS containing an assume turns into too many quantifiers in the eventual query. Manasij do you want to share that here?

nunoplopes:
> Nuno, I have no idea if it's related, but @manasij7479 has an example where a LHS containing an assume turns into too many quantifiers in the eventual query. Manasij do you want to share that here?

That might be unrelated. FWIW I'm aware of an issue between assume & function calls, which might be related to what you are describing. But I would love to see more examples.

nunoplopes:
Ok, my 2nd take on refinement. We need to prove 2 things **when both Pre(I) and Pre'(I) hold**:
1) `dom contained in dom'`
2) `output'(I) contained in output(I)`

Translating to math we get (already negated for SMT):
1) `∃ I, N, N'' . Pre(I, N) ∧ dom(I, N) ∧ Pre'(I, N'') ∧ (∀ N' . ¬ Pre'(I, N') ∨ ¬ dom'(I, N'))`
reasoning: if Pre' is always false for a given I, we should skip that I. Otherwise, we must ensure that dom' holds.
Currently we miss the existentially quantified dom', which is why we see those spurious reports for inputs for which Pre never holds.

2) `∃ I, N', N'', O . Pre'(I, N') ∧ P'(I, N', O) ∧ Pre(I, N'') ∧ (∀ N . ¬ Pre(I, N) ∨ ¬ P(I, N, O))`
same reasoning as before. our current encoding misses the same bit.

Comments?

nunoplopes:
Thanks to covid we now have science being done live without safety nets like a good reality show 😁
Ok, I've implemented the refinement conditions above, but of course something was missing. Function calls are not account for above. They are not internal non-determinism, and right now I don't know where that should go. Need to spend more time on this tmr.

nunoplopes:
The fix was much less problematic that anticipated. It was about dealing with the particular corner case your test cases show (Pre(I, N) = false for fixed I and arbitrary N). This requires duplicating Pre when N is non-empty, but perf differences are negligible.

https://github.com/AliveToolkit/alive2/commit/d93f33ff34bc2a96b83cdcfbcc05aaee4c5061cb

PR #439, opened 2020-06-24
Add refinedLocal/refinedByVal, make byval at definition have no effect

aqjune:
This is the first splitted pull request from #428.
Originally I planned to contain the byval bug only, became to think that having relevant codes included would be better.

Conceptually, what this patch does is to make Pointer::refined and Pointer::fninputRefined equivalent except byval processing. This is done by factoring out two functions and calling them:

- Pointer::refinedByVal(other) encodes a refinement for byval-passed pointers. This purely does dereferenceability check only, and doesn't check alloc type / etc. In the upcoming patches, this should compare bytes as well.
- Pointer::refinedLocal(other) encodes a refinement for local pointers. This check alloc types, offsets, etc.

Note that this PR makes refinement of local blocks at Pointer::fninputRefined more strict: even if two blocks are alloca, their sizes should be equivalent. LLVM does not seem to change the size of an alloca passed to fn arg, unless it is passed as byval. I tested it with LLVM unit tests and had no regression. Same with bzip2/gzip.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2044,6 +2044,7 @@ StateValue Return::toSMT(State &s) const {
   auto &retval = s[*val];
   s.addUB(s.getMemory().checkNocapture());
   addUBForNoCaptureRet(s, retval, val->getType());
+  s.getMemory().markAllocasAsDead();
 
   auto &attrs = s.getFn().getFnAttrs();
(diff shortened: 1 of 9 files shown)


aqjune:
Stale branch, will make newer one later

PR #442, opened 2020-07-04
Fix a bug of dereferenceable attribute (it should not guarantee the pointer is aligned)

aqjune:
IMO this happened because FnCall wasn't defined as a subclass of MemInstr.
I changed FnCall to inherit MemInstr.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -21,6 +21,8 @@ using namespace std;
     val = &with
 #define DEFINE_AS_RETZERO(cls, method) \
   uint64_t cls::method() const { return 0; }
+#define DEFINE_AS_RETFALSE(cls, method) \
+  bool cls::method() const { return false; }
 #define DEFINE_AS_EMPTYACCESS(cls) \
   MemInstr::ByteAccessInfo cls::getByteAccessInfo() const \
   { return {}; }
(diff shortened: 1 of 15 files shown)


aqjune:
Found this bug while running tests after disabling byte widening opt

aqjune:
No regression in unit tests

ir/instr.cpp
```diff
@@ -1367,6 +1369,38 @@ unique_ptr<Instr> InsertValue::dup(const string &suffix) const {
...
+      if (arg.second.has(ParamAttrs::Dereferenceable)) {
+        has_deref = true;
+        break;
+      }
+  }
+  if (!has_deref)
+    // No dereferenceable attribute
+    return {};
+  // dereferenceable(n) does not guarantee that the pointer is n-byte aligned
+  return ByteAccessInfo::intOnly(1);
```
nunoplopes, line 1401: this is a pessimizing things a bit:
 1) it forces int bytes to exist. We should have a "any" return to avoid this. I think there are other fns that could use the same (e.g. memcmp)
 2) it brings all bits_byte down to 8. We can't have all functions with fn calls to have 8-bit bytes. That's a big perf impact. We need a way around this.

aqjune: 1. For any byte this should be an easy change.

2. We can check whether all path to dereferenceable requires some alignment of the pointer, but this requires flow-sensitive analysis.
I guess when LLVM infers `dereferenceable` from preceding memory accesses, it will also attach `align` tags. It will naturally solve the problem. When `align` is supported, this should be fixed (which should be after paper submission). What do you think?

nunoplopes: hopefully it shouldn't need any data-flow analysis, just a bit of pattern matching maybe? It doesn't have to be perfect; just be  good for the 99% cases, and give up elsewhere I guess. But let's think about this after the submission, yes :)

aqjune:
Running LLVM unit tests. 
I'm filtering out function pairs without memory ops, and this seems to make test pretty fast..!

aqjune:
(but this time forgot to turn on the filter)
No regression in the Unit Test result.

aqjune:
I found from ReleaseNote the similar statement from what this patch is doing (https://github.com/llvm/llvm-project/blob/release/11.x/llvm/docs/ReleaseNotes.rst#changes-to-the-llvm-ir ):

```
dereferenceable attributes and metadata on pointers no longer imply anything about the alignment of the pointer in question. Previously, some optimizations would make assumptions based on the type of the pointer. This behavior was undocumented. To preserve optimizations, frontends may need to be updated to generate appropriate align attributes and metadata.
```

It mentions that the frontend needs to be updated to give an appropriate align, and the trunk seems to already generate it (https://godbolt.org/z/q9cxP4 ); maybe we can rely on it?
I remember that with this patch the performance change wasn't observable in the benchmarks as well (it was around the early July).

PR #447, opened 2020-07-06
Stricter refinement check for nonlocals

aqjune:
Will leave experimental results here after they're done

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -778,10 +778,10 @@ expr Pointer::refined(const Pointer &other) const {
   // TODO: this induces an infinite loop
   //local &= block_refined(other);
 
-  return isBlockAlive().implies(
-           other.isBlockAlive() &&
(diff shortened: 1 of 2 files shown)


ir/memory.cpp
```diff
@@ -809,9 +809,9 @@ expr Pointer::fninputRefined(const Pointer &other, set<expr> &undef,
   // TODO: this induces an infinite loop
   // block_refined(other);
 
-  return isBlockAlive().implies(
-           other.isBlockAlive() &&
-             expr::mkIf(isLocal(), local, *this == other));
+  return expr::mkIf(isLocal(),
+      isBlockAlive().implies(other.isBlockAlive() && local),
+      *this == other && isBlockAlive() == other.isBlockAlive());
```
nunoplopes, line 814: can we use `isBlockAlive() == other.isBlockAlive()` be used for both cases?

PR #453, opened 2020-08-06
Support noundef attribute

aqjune:
Since it is an important attribute, I made a patch that adds it.

noundef for aggregate is left as unsupported, because noundef does not constraint padding of aggregate values to be non-undef.

It reuses the undef-checking function that is used by br/switch to encode UB.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1458,14 +1458,13 @@ strip_undef(State &s, const Value &val, const expr &e) {
 
   expr c, a, b, lhs, rhs, ty;
   unsigned h, l;
-  uint64_t n;
 
   // (ite (= ((_ extract 0 0) ty_%var) #b0) %var undef!0)
   if (e.isIf(c, a, b) && s.isUndef(b) && c.isEq(lhs, rhs)) {
-    if (lhs.isUInt(n))
+    if (lhs.isZero())
       swap(lhs, rhs);
 
-    if (rhs.isUInt(n) && n == 0 &&
+    if (rhs.isZero() &&
         lhs.isExtract(ty, h, l) && h == 0 && l == 0 && isTyVar(ty, a))
       return { move(a), c };
   }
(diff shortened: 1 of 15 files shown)


ir/instr.cpp
```diff
@@ -1451,27 +1451,62 @@ void FnCall::print(ostream &os) const {
     os << "\t; WARNING: unknown known function";
 }
 
-static void unpack_inputs(State&s, Type &ty, const ParamAttrs &argflag,
+static pair<expr,expr> // <extracted value, not_undef>
+strip_undef(State &s, const Value &val, const expr &e) {
```
aqjune, line 1455: Renamed jump_undef_condition , since it isn't only used for jump now on.

tests/alive-tv/attrs/noundef.srctgt.ll
```diff
@@ -0,0 +1,16 @@
+define i32 @src(i1 noundef %cond) {
```
LebedevRI, line 1: Passing-by question, might be barking at the wrong tree here:
is `{i32, i32} noundef %v` valid? I think alive2 doesn't accept it now.

aqjune: It is valid, but simply unsupported yet. It requires implementation of element-wise undef of inputs.

nunoplopes: not really. since noundef makes everything non-undef and non-poison. It can be implemented right away I think.

aqjune: IMO padding elements of aggregates is the issue - It can be undef even if noundef is attached.

nunoplopes: you can still implement it by making all padding undef or not undef, and the remaining values non-undef. Not perfect, but better than nothing.

aqjune: If that is okay, I'll put it in my todo list - I guess the patch wouldn't be very complex to implement.

LebedevRI: FWIW i asked out of practical interest, i wanted that for https://reviews.llvm.org/D85787 test coverage.

PR #456, opened 2020-08-10
Minor refactoring to fn call add

aqjune:
This is a simple patch that does minor refactoring to State::addFnCall and Memory::mkFnRet.


ir/memory.cpp
```diff
@@ -1595,10 +1595,6 @@ pair<expr, expr> Memory::mkUndefInput(const ParamAttrs &attrs) const {
 
 pair<expr,expr>
 Memory::mkFnRet(const char *name, const vector<PtrInput> &ptr_inputs) {
-  for (auto &in : ptr_inputs) {
-    if (!in.byval && !in.nocapture && !in.val.non_poison.isFalse())
-      escapeLocalPtr(in.val.value);
-  }
```
aqjune, line 1601: This is done by caller of Memory::mkFnRet, which is State::addFnCall.

PR #459, opened 2020-08-12
Add support for noundef aggregates

aqjune:
This is a patch that adds support for aggregate inputs with noundef attribute.
It seems an empty aggregate input isn't supported well (mkInput/mkUndefInput returns invalidexpr);
`noundefaggr-empty.srctgt.ll` is a relevant test; it is marked as xfail.
Since a fix for the empty aggregate bug may conflict with this PR, I'll make a follow-up patch after this is reviewed.

### DIFF
--- a/ir/type.cpp
+++ b/ir/type.cpp
@@ -943,10 +943,18 @@ AggregateType::mkUndefInput(State &s, const ParamAttrs &attrs) const {
   vector<expr> vars;
 
   for (unsigned i = 0; i < elements; ++i) {
-    auto [v, vs] = children[i]->mkUndefInput(s, attrs);
-    v = children[i]->toBV(move(v));
-    val = i == 0 ? move(v) : val.concat(v);
-    vars.insert(vars.end(), vs.begin(), vs.end());
+    auto *aty = dynamic_cast<AggregateType *>(children[i]);
+    if (!attrs.has(ParamAttrs::NoUndef) || isPadding(i) || aty){
+      // If it has a nested aggregate, the child deals with noundef
+      auto [v, vs] = children[i]->mkUndefInput(s, attrs);
+      v = children[i]->toBV(move(v));
+      val = i == 0 ? move(v) : val.concat(v);
+      vars.insert(vars.end(), vs.begin(), vs.end());
+    } else {
+      auto v = children[i]->getDummyValue(true).value;
+      v = children[i]->toBV(move(v));
+      val = i == 0 ? move(v) : val.concat(v);
+    }
   }
   return { move(val), move(vars) };
 }
--- a/ir/type.cpp
+++ b/ir/type.cpp
@@ -805,12 +805,18 @@ unsigned AggregateType::np_bits() const {
 }
 
 StateValue AggregateType::getDummyValue(bool non_poison) const {
+  return getDummyValue(non_poison, non_poison);
+}
+
+StateValue AggregateType::getDummyValue(bool non_poison,
+                                        bool non_poison_padding) const {
   if (elements == 0)
     return { expr::mkUInt(0, 1), expr::mkUInt(non_poison, 1) };
 
   vector<StateValue> vals;
   for (unsigned i = 0; i < elements; ++i) {
-    vals.emplace_back(children[i]->getDummyValue(non_poison));
+    bool np = isPadding(i) ? non_poison_padding : non_poison;
+    vals.emplace_back(children[i]->getDummyValue(np));
   }
   return aggregateVals(vals);
 }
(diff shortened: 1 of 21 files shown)


tests/alive-tv/attrs/noundefaggr-disableundefinput.srctgt.ll
```diff
@@ -0,0 +1,33 @@
+; TEST-ARGS: -disable-undef-input
+; FIXME: This transformation should raise 'Source is more defined than target' error
```
nunoplopes, line 2: uhm, so the issue is that padding should always be poison irrespective of the disable-xx-input, right?  We can do that, as we know which fields are padding and which are not, right?

ir/value.cpp
```diff
@@ -175,6 +175,19 @@ void Input::print(ostream &os) const {
   UNREACHABLE();
 }
 
+static bool has_padding(const AggregateType *aggr_ty) {
```
nunoplopes, line 178: code would become simpler if this function received Type* as input instead

ir/value.cpp
```diff
@@ -194,8 +207,11 @@ StateValue Input::toSMT(State &s) const {
     val = getType().mkInput(s, smt_name.c_str(), attrs);
   }
 
+  auto *aggr_ty = dynamic_cast<const AggregateType *>(&getType());
+  // Aggregate's paddings can be undef even if it is noundef
```
nunoplopes, line 211: can we make padding poison?

ir/type.cpp
```diff
@@ -943,10 +943,18 @@ AggregateType::mkUndefInput(State &s, const ParamAttrs &attrs) const {
   vector<expr> vars;
 
   for (unsigned i = 0; i < elements; ++i) {
-    auto [v, vs] = children[i]->mkUndefInput(s, attrs);
-    v = children[i]->toBV(move(v));
-    val = i == 0 ? move(v) : val.concat(v);
-    vars.insert(vars.end(), vs.begin(), vs.end());
+    auto *aty = dynamic_cast<AggregateType *>(children[i]);
```
nunoplopes, line 946: same comment as in the other file: padding should be poison I think.

tools/transform.cpp
```diff
@@ -94,7 +94,7 @@ static void print_varval(ostream &os, State &st, const Model &m,
   for (unsigned i = 0, e = agg->numElementsConst(); i < e; ++i) {
     if (i != 0)
       os << ", ";
-    print_varval(os, st, m, var, agg->getChild(i), agg->extract(val, i));
+    print_varval(os, st, m, nullptr, agg->getChild(i), agg->extract(val, i));
```
aqjune, line 97: Fixed here because `var` cannot be used when printing agg's elements.
It makes print_single_varval's `assert(n == 0)` fail.

aqjune:
A bit of code was added because noundef for aggregates at return values & function calls (not input only) should have been added.

ir/value.cpp
```diff
@@ -220,25 +206,22 @@ StateValue Input::toSMT(State &s) const {
...
+  if (has_deref) {
     Pointer p(s.getMemory(), val);
-    if (has_deref) {
-      s.addAxiom(type == 0);
-      s.addAxiom(p.isDereferenceable(attrs.derefBytes, bits_byte/8, false));
-    }
-    if (has_nonnull && !has_deref) {
-      s.addAxiom(type.extract(1, 1) == 0);
-    }
+    s.addAxiom(p.isDereferenceable(attrs.derefBytes, bits_byte/8, false));
```
aqjune, line 211: This neatens the poison encoding of nonnull attribute; `has_nonnull` is added to `never_poison`.

aqjune: Also, since `has_deref` already existed at `never_undef`, `s.addAxiom(type == 0);` is not necessary.

tests/alive-tv/attrs/noundefaggr-disableall.srctgt.ll
```diff
@@ -0,0 +1,29 @@
+; TEST-ARGS: -disable-undef-input -disable-poison-input
```
nunoplopes, line 1: I don't think this should prevent padding to be poison. Padding should always be poison regardless.

tools/transform.cpp
nunoplopes: I think the right fix is to skip padding elements

tests/alive-tv/attrs/noundefaggr-disableall.srctgt.ll
aqjune: I thought about this, and I agree - padding should not convey meaningful data. They cannot be retrieved using `extractvalue`, so something is weird otherwise.
This implies that reading aggregate from the memory should always fill the padding with poison - should we update `Memory::load` to do this? Maybe suggestion to LangRef, too.

nunoplopes: sounds good, thanks!

tools/transform.cpp
aqjune: I tried this, and it still happened when -disable-undef-input was added.
I updated print_single_varval to deal with the case when n was 2:
```
    if (n == 1) {
      os << "undef";
      return;
    } else if (n == 2) { // this branch is added
      os << "poison";
      return;
    }
```

aqjune:
TODOs here:
- Better empty aggregate support
- Memory::load update w.r.t. padding
- LangRef update suggestion about padding

PR #462, opened 2020-08-16
Fix invalid expr error when an empty struct is used

aqjune:
This patch encodes an empty aggregate as a statevalue `{expr::mkUInt(0, 1), expr::mkUInt(1, 1)}`.
I used uint for non-poison to make `AggregateType::np_bits` return a valid value.

LLVM unit tests: no failure regression, no invalidexpr test added

### DIFF
--- a/ir/type.cpp
+++ b/ir/type.cpp
@@ -728,8 +728,8 @@ StateValue AggregateType::aggregateVals(const vector<StateValue> &vals,
                                         bool needsPadding) const {
   assert(vals.size() + (needsPadding ? numPaddingsConst() : 0) == elements);
   // structs can be empty
-  if (vals.empty() && !needsPadding)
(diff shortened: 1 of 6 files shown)


ir/type.cpp
```diff
@@ -789,6 +786,10 @@ AggregateType::extract(const StateValue &val, unsigned index, bool fromInt)
 }
 
 unsigned AggregateType::bits() const {
+  if (elements == 0)
+    // It is set as 1 because zero-width bitvector is invalid.
+    return 1;
```
nunoplopes, line 791: uhm, doesn't this have an impact on stores? Storing an empty struct is an no-op, right?  But not if the size is 1 bit, I guess.

aqjune: I think it is okay - it's because an aggregate is elementwisely stored.
I added a relevant test for this.

PR #464, opened 2020-08-16
Automatically fill padding with poison when loading an aggregate value

aqjune:
No regression from LLVM unit tests.

I'll run single file benchmarks too & see whether there is any failure.
If there is no failure, I think I can send the suggestion to llvm-dev.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -1947,13 +1947,19 @@ StateValue Memory::load(const Pointer &ptr, const Type &type, set<expr> &undef,
     vector<StateValue> member_vals;
     unsigned byteofs = 0;
     for (unsigned i = 0, e = aty->numElementsConst(); i < e; ++i) {
+      // Padding is filled with poison.
+      if (aty->isPadding(i)) {
+        byteofs += getStoreByteSize(aty->getChild(i));
+        continue;
+      }
+
       auto ptr_i = ptr + byteofs;
       auto align_i = gcd(align, byteofs % align);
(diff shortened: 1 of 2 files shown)


aqjune:
Has no regression from single file benchmark (finds a few bugs which are known to be unsound transformations). I think this patch is fine.

When talking about this with LLVM people, I'd like to say that the loaded padding should be undef first. The reason is that saying padding should be poison seems to be an orthogonal issue from this.

ir/memory.cpp
```diff
@@ -1947,13 +1947,19 @@ StateValue Memory::load(const Pointer &ptr, const Type &type, set<expr> &undef,
...
+      }
+
       auto ptr_i = ptr + byteofs;
       auto align_i = gcd(align, byteofs % align);
       member_vals.emplace_back(load(ptr_i, aty->getChild(i), undef, align_i));
       byteofs += getStoreByteSize(aty->getChild(i));
     }
     assert(byteofs == bytecount);
-    return aty->aggregateVals(member_vals);
+    return aty->aggregateVals(member_vals, true);
```
nunoplopes, line 1962: is there any use of aggregateVals(foo, false) left?

aqjune: AggregateValue::toSMT and AggregateType::fromInt use it; In case of other places it seems giving false is not necessary.

aqjune:
The LangRef update landed: https://reviews.llvm.org/rG98e5776897ce

PR #469, opened 2020-08-25
Make malloc(poison) UB

aqjune:
As discussed, this is a patch that makes malloc(poison) UB.


ir/instr.cpp
```diff
@@ -2448,6 +2450,7 @@ StateValue Calloc::toSMT(State &s) const {
   auto &[nm, np_num] = s.getAndAddUndefs(*num);
   auto &[sz, np_sz] = s.getAndAddUndefs(*size);
   auto np = np_num && np_sz;
+  s.addUB(np);
```
nunoplopes, line 2453: should the return below be changed to return { foo, true} ? i.e., malloc won't return poison anymore, right?

aqjune: Right, I updated it & added a test

Issue #468, opened 2020-08-25
State should track definitely not poison vars


nunoplopes:
As an optimization, we should propagate non-poisonness of things like:
```
%a = udiv %b, %c  ; implies %c is not-poison
use(%c)  ; should have np = true
```

This property depends on the control-flow, so needs to be tracked in state per BB, and merged between predecessors accordingly.
What's a good API to state inform a var can't be poison? (and perhaps call addUB() by iself?)
Also, can we also propagate non-undefness? (this needs carrying a different whole StateValue)

aqjune:
What about simply updating s[%c].non_poison to true when the udiv is executed (BinOp::toSMT is called)?

nunoplopes:
No, you can't do that since it's only non-poison is the current branch. Might not be in others. This is a control-flow specific property,

aqjune:
Sorry, you're right.
Can we expand DomainPreds to contain the non_poison information?
Or, predecessor_data may have something like triple<DomainPreds, AnalysisInfo, DisjointExpr<Memory>>,  instead of pair<DomainPreds, DisjointExpr<Memory>>.
State::operator[] can use the AnalysisInfo to return an optimized value.

nunoplopes:
Yep, I was thinking in something like that.
plus we need a new e.g. State::getNonPoison() that does [] + adds np to UB + stores info that it can't be poison.
Feel free to give it a try if you want (& benchmark so we see if it helps at all). I'm a bit short on time ATM.

aqjune:
Okay, I'll give it a try and see its performance change.

PR #472, opened 2020-08-27
Add poison analysis

aqjune:
This implements #468.

This removes one timeout from Alive2 unit tests.
I'll run & leave the LLVM unit test / single file benchmark results here.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -632,7 +632,7 @@ StateValue BinOp::toSMT(State &s) const {
   } else {
     scalar_op = [&](auto a, auto ap, auto b, auto bp) -> StateValue {
       auto [v, np] = fn(a, ap, b, bp);
-      return { move(v), ap && (BinOp::isDivOrRem(op) ? true : bp) && np };
+      return { move(v), ap && bp && np };
     };
   }
 
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -249,7 +249,7 @@ void BinOp::print(ostream &os) const {
 
 static void div_ub(State &s, const expr &a, const expr &b, const expr &ap,
                    const expr &bp, bool sign) {
-  s.addUB(bp);
+  // addUB(bp) is not needed because it is registered by getAndAddPoisonUB.
   s.addUB(b != 0);
   if (sign)
     s.addUB((ap && a != expr::IntSMin(b.bits())) || b != expr::mkInt(-1, b));
@@ -632,12 +632,12 @@ StateValue BinOp::toSMT(State &s) const {
   } else {
     scalar_op = [&](auto a, auto ap, auto b, auto bp) -> StateValue {
       auto [v, np] = fn(a, ap, b, bp);
-      return { move(v), ap && bp && np };
+      return { move(v), ap && (BinOp::isDivOrRem(op) ? true : bp) && np };
     };
   }
 
   auto &a = s[*lhs];
-  auto &b = s[*rhs];
+  auto &b = isDivOrRem(op) ? s.getAndAddPoisonUB(*rhs) : s[*rhs];
 
   if (lhs->getType().isVectorType()) {
     auto retty = getType().getAsAggregateType();
@@ -744,6 +744,18 @@ expr BinOp::getTypeConstraints(const Function &f) const {
 unique_ptr<Instr> BinOp::dup(const string &suffix) const {
   return make_unique<BinOp>(getType(), getName()+suffix, *lhs, *rhs, op, flags,
                             fmath);
+}
+
+bool BinOp::isDivOrRem(const Op op) {
+  switch (op) {
+  case Op::SDiv:
+  case Op::SRem:
+  case Op::UDiv:
+  case Op::URem:
+    return true;
+  default:
+    return false;
+  }
 }
 
 
@@ -2530,10 +2542,9 @@ void Free::print(std::ostream &os) const {
 }
 
 StateValue Free::toSMT(State &s) const {
-  auto &[p, np] = s[*ptr];
-  s.addUB(np);
+  auto p = s.getAndAddPoisonUB(*ptr).value;
   // If not heaponly, don't encode constraints
-  s.getMemory().free(p, !heaponly);
+  s.getMemory().free(move(p), !heaponly);
 
   if (s.getFn().getFnAttrs().has(FnAttrs::NoFree) && heaponly)
     s.addUB(expr(false));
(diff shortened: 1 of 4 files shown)


ir/instr.cpp
```diff
@@ -2530,10 +2542,9 @@ void Free::print(std::ostream &os) const {
 }
 
 StateValue Free::toSMT(State &s) const {
-  auto &[p, np] = s[*ptr];
-  s.addUB(np);
+  auto p = s.getAndAddPoisonUB(*ptr).value;
```
nunoplopes, line 2545: &p and no move below

ir/instr.cpp
```diff
@@ -2706,9 +2717,8 @@ void Load::print(std::ostream &os) const {
 }
 
 StateValue Load::toSMT(State &s) const {
-  auto &[p, np] = s[*ptr];
-  s.addUB(np);
-  auto [sv, ub] = s.getMemory().load(p, getType(), align);
+  auto p = s.getAndAddPoisonUB(*ptr).value;
```
nunoplopes, line 2720: &p

ir/state.cpp
```diff
@@ -95,14 +126,27 @@ const StateValue& State::getAndAddUndefs(const Value &val) {
...
 
+const StateValue& State::getAndAddPoisonUB(const Value &val) {
+  auto &v = (*this)[val];
+  if (!v.non_poison.isTrue()) {
+    analysis.non_poison_vals.insert(&val);
+    if (v.non_poison.isBool())
+      addUB(v.non_poison);
+    else
+      // UB if any element is poison
+      addUB(v.non_poison != expr::mkUInt(-1, v.non_poison.bits()));
```
nunoplopes, line 137: uhm, i would make this == 0. We want all elements to be non-poison.

ir/state.cpp
```diff
@@ -60,10 +72,28 @@ const StateValue& State::operator[](const Value &val) {
   auto &[sval, uvars] = val_uvars;
   (void)var;
 
+  auto simplify = [&](StateValue& sv, bool use_new_slot) -> StateValue& {
+    if (analysis.non_poison_vals.count(&val)) {
+      // Store a simplified state value into tmp_values
+      if (use_new_slot) {
+        assert(i_tmp_values < tmp_values.size());
+        tmp_values[i_tmp_values++] = sv;
+      } else {
```
nunoplopes, line 81: this else branch isn't needed. the value is already there

ir/state.cpp
```diff
@@ -60,10 +72,28 @@ const StateValue& State::operator[](const Value &val) {
...
+        assert(i_tmp_values > 0);
+        tmp_values[i_tmp_values - 1] = sv;
+      }
+      StateValue &sv_new = tmp_values[i_tmp_values - 1];
+      expr np = sv_new.non_poison;
+      sv_new.non_poison = np.isBool() ? true : expr::mkUInt(-1, np.bits());
```
nunoplopes, line 87: should be 0, which is a non-poison value. Also, no copy on the line above.

ir/state.cpp
aqjune: Sorry, my mistake :/ Naively thought that non_poison being true means it should be 1 when it is a bitvector

aqjune:
The incorrect encoding in non_poison caused a few failures in LLVM unit tests.
Rerunning, should be done in ~45 mins

ir/state.cpp
```diff
@@ -95,14 +123,27 @@ const StateValue& State::getAndAddUndefs(const Value &val) {
...
+  auto &v = (*this)[val];
+  if (v.isValid()) {
+    // If val is an aggregate, all elements should be non-poison
+    expr is_np = v.non_poison.isBool() ? v.non_poison : v.non_poison == 0;
+    if (!is_np.isTrue()) {
+      analysis.non_poison_vals.insert(&val);
+      addUB(move(is_np));
+    }
+  }
+  return v;
```
nunoplopes, line 136: this is not resetting np to true if np wasn't true before.

aqjune: Callers of getAndAddPoisonUB are using the `value` field only (except BinOp, which assumes non_poison is true if if was div or rem so now).
Or, the state value can be temporarily stored to `tmp_values[i_tmp_values++]`, its non_poison is updated to true, and the reference may be returned, but it would cause use of an extra slot.

aqjune:
For LLVM unit tests: no difference in performance, no regression in failure.
I'll leave single file benchmark result after it's done here.

ir/state.cpp
nunoplopes: it's alright to use an extra slot

aqjune: Well, okay then. I will update this part

aqjune:
Well, the result of single file benchmark wasn't impressive. :/ Compared with master.
[result.txt](https://github.com/AliveToolkit/alive2/files/5137559/result.txt)
But the difference isn't big, and I believe this PR can be a good start for performing undef analysis as well (which will be much helpful for performance)

ir/state.cpp
```diff
@@ -95,14 +123,22 @@ const StateValue& State::getAndAddUndefs(const Value &val) {
   return v;
 }
 
+const StateValue& State::getAndAddPoisonUB(const Value &val) {
+  analysis.non_poison_vals.insert(&val);
+  auto &np = get<1>(values[values_map.at(&val)]).first.non_poison;
```
nunoplopes, line 128: this is not correct because np may refer undef variables.

aqjune: Sorry, fixed it. It was tricky, hope it is correct this time.

nunoplopes: looks good to me, thanks!
BTW, can you run tests again after these changes to check if the results are the same?

aqjune: Sure. I'll leave the LLVM unit test results first, then the single file benchmarks.

aqjune:
LLVM unit tests: no regression
When single file benchmarks are done, I'll leave comments here

aqjune:
Results: 
[result.txt](https://github.com/AliveToolkit/alive2/files/5139556/result.txt)
+1 incorrect in oggenc is valid (canonicalizes load pointer to load i64)
(sqlite3 is still running)

aqjune:
For sqlite3, the performance seems better: [result-sqlite3.txt](https://github.com/AliveToolkit/alive2/files/5143833/result-sqlite3.txt)

nunoplopes:
The results are a bit mixed. Some new bugs, some bugs missed. Time is meh.  Z3 is too smart 😅
Should we merge this? What are your thoughts for "undef analysis" you mentioned above?

aqjune:
Maybe I can try two things in a few days -
(1) See how it goes with the recent clang - now printf/scanf/... has noundef at its arguments. Probably this may affect the performance.
(2) Yes, the undef analysis - I think it's worth giving it a try. Would using the strip_undef be helpful? This won't remove all undefs though.

nunoplopes:
strip_undef is only used for branch & switch. Those propagate non-undefness through the CFG, but probably those registers are not used elsewhere, so upside for optimization is probably low. e.g. the code we usually see is icmp + br and the value of icmp isn't reused.
For those builtins maybe they can help. It's worth trying.

aqjune:
I did not have a bandwidth for this (it needs to rebase the bitcode generating patch & run scripts); I'll start this after the talk preparation is done.

regehr:
@aqjune let me know if I can help look at slides or give feedback on a practice talk!

aqjune:
Will start working on this

PR #483, opened 2020-09-15
Support for fminimum, fmaximum

zhengyangl:
Support for for fminimum, fmaximum, langref: https://llvm.org/docs/LangRef.html#llvm-minimum-intrinsic, https://llvm.org/docs/LangRef.html#llvm-maximum-intrinsic

I am starting a TV run on llvm test suite now :)


ir/instr.cpp
```diff
@@ -594,6 +598,24 @@ StateValue BinOp::toSMT(State &s) const {
...
     break;
 
+  case FMinimum:
+  case FMaximum:
+    fn = [&](auto a, auto ap, auto b, auto bp) -> StateValue {
+      auto v = [&](expr &a, expr &b) {
+        expr zpos = expr::mkNumber("0", a), zneg = expr::mkNumber("-0", a);
+        expr cmp = (op == FMinimum) ? a.fole(b) : a.foge(b);
+        expr neg_cond = (op == FMinimum) ? a.isFPNeg() || b.isFPNeg()
+                                         : a.isFPNeg() && b.isFPNeg();
```
nunoplopes, line 608: is this bit needed? Did you check the spec to see it smt already does the right thing?  (it's very likely as LLVM and SMT aren't inventing different semantics)

zhengyangl: Yeah it is needed.  The SMT less than returns false if both operand is either +0 or -0 but LLVM's maximum and minimum are sensitive to the sign of zeros. 

See https://github.com/Z3Prover/z3/blob/72d140334fff2d285956b8419b552d727619156a/src/ast/fpa/fpa2bv_converter.cpp#L2237

zhengyangl:
I checked the tv result, seems that this patch doesn't expose any bugs. The patch can verify the correctness of ./InstCombine/maximum.ll and ./InstCombine/minimum.ll except a few timeouts.

PR #487, opened 2020-09-20
Add undef analysis

aqjune:
This is a patch that follows #472 .

This patch adds a boolean argument to `State::getAndAddPoisonUB` that states whether passing an undef is UB as well and should strip it.

Consists of two commits - with the first commit only, the result of single file benchmark didn't improve. 

With the second comment, on my machine, `alive-tv/memory/storeptr.src.ll` raises no more timeout (became 2.7 sec). Single file benchmarks and unit tests are running.
But the second commit requires an assumption that passing a partially undefined pointer to load/store raises UB. I think this can be justified because memory sanitizer also detects such case (when an 'expensive detection' flag is turned on)

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2769,7 +2769,10 @@ void Load::print(std::ostream &os) const {
 }
 
 StateValue Load::toSMT(State &s) const {
-  auto &p = s.getAndAddPoisonUB(*ptr).value;
+  // NOTE: We neeed clarification whether loading a partially undefined pointer
+  // is UB. Memory sanitizer has an option (detecting a partially undef pointer)
+  // for this, so worth discussing it.
+  auto &p = s.getAndAddPoisonUB(*ptr, true).value;
   auto [sv, ub] = s.getMemory().load(p, getType(), align);
   s.addUB(move(ub));
   return sv;
@@ -2811,7 +2814,10 @@ void Store::print(std::ostream &os) const {
 }
 
 StateValue Store::toSMT(State &s) const {
-  auto &p = s.getAndAddPoisonUB(*ptr).value;
+  // NOTE: We neeed clarification whether store to a partially undefined pointer
+  // is UB. Memory sanitizer has an option (detecting a partially undef pointer)
+  // for this, so worth discussing it.
+  auto &p = s.getAndAddPoisonUB(*ptr, true).value;
   auto &v = s[*val];
   s.getMemory().store(p, v, val->getType(), align, s.getUndefVars());
   return {};
@@ -2915,7 +2921,7 @@ void Memcpy::print(ostream &os) const {
 StateValue Memcpy::toSMT(State &s) const {
   auto &[vdst, np_dst] = s[*dst];
   auto &[vsrc, np_src] = s[*src];
-  auto &vbytes = s.getAndAddPoisonUB(*bytes).value;
+  auto &vbytes = s.getAndAddPoisonUB(*bytes, true).value;
   s.addUB((vbytes != 0).implies(np_dst && np_src));
 
   if (vbytes.bits() > bits_size_t)
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1587,79 +1587,6 @@ void FnCall::print(ostream &os) const {
     os << "\t; WARNING: unknown known function";
 }
 
-static expr eq_except_padding(const Type &ty, const expr &e1, const expr &e2) {
-  const auto *aty = ty.getAsAggregateType();
-  if (!aty)
-    return e1 == e2;
-
-  StateValue sv1{expr(e1), expr()};
-  StateValue sv2{expr(e2), expr()};
-  expr result = true;
-
-  for (unsigned i = 0; i < aty->numElementsConst(); ++i) {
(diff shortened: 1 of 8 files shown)


ir/instr.cpp
```diff
@@ -2309,7 +2225,7 @@ StateValue Return::toSMT(State &s) const {
 
   auto &attrs = s.getFn().getFnAttrs();
   if (attrs.poisonImpliesUB())
-    retval = s.getAndAddPoisonUB(*val);
+    retval = s.getAndAddPoisonUB(*val, attrs.undefImpliesUB());
```
nunoplopes, line 2228: cant we add an assert to attributes to check that undefImpliesUB() implies poisonImpliesUB()?  Otherwise this code makes me nervous.

ir/instr.cpp
```diff
@@ -2860,7 +2769,10 @@ void Load::print(std::ostream &os) const {
 }
 
 StateValue Load::toSMT(State &s) const {
-  auto &p = s.getAndAddPoisonUB(*ptr).value;
+  // NOTE: We neeed clarification whether loading a partially undefined pointer
+  // is UB. Memory sanitizer has an option (detecting a partially undef pointer)
+  // for this, so worth discussing it.
+  auto &p = s.getAndAddPoisonUB(*ptr, true).value;
```
nunoplopes, line 2775: I would say it's weird to store to an undef pointer, even if just 1 bit of the offset is undef.
Maybe submit a patch to LangRef to clarify this so we are all on the same page.

ir/state.cpp
```diff
@@ -143,10 +238,24 @@ static expr not_poison_except_padding(const Type &ty, const expr &np) {
...
+State::getAndAddPoisonUB(const Value &val, bool undef_ub_too) {
+  auto &sv = (*this)[val];
+
+  if (!analysis.non_poison_vals.insert(&val).second && !undef_ub_too)
+    return sv;
+
+  expr v = sv.value;
+
+  if (undef_ub_too) {
+    auto inserted = analysis.non_undef_vals.find(&val);
```
nunoplopes, line 251: inserted is not a good choice of var name here. can it be simply `I` for iterator?

nunoplopes: also, you can use `try_emplace(&val)`, for example, to avoid the second map lookup below.

ir/state.cpp
```diff
@@ -143,10 +238,24 @@ static expr not_poison_except_padding(const Type &ty, const expr &np) {
...
+    if (inserted != analysis.non_undef_vals.end()) {
+      v = inserted->second;
+    } else {
+      v = strip_undef_and_add_ub(*this, val, v);
+      analysis.non_undef_vals[&val] = v;
+    }
+  }
 
   // mark all operands of val as non-poison if they propagate poison
```
nunoplopes, line 260: we should skip this bit (up to addUb below) if we had already marked the value as non-poison before.

ir/state.cpp
```diff
@@ -184,7 +293,9 @@ const OrExpr* State::jumpCondFrom(const BasicBlock &bb) const {
   return I == pres.end() ? nullptr : &I->second.domain.path;
 }
 
-bool State::isUndef(const expr &e) const {
+bool State::isUndef(const expr &e, const Value *used_by) const {
+  if (used_by)
+    return at(*used_by).second.count(e) != 0;
```
nunoplopes, line 298: does this give any advantage? I would be surprised as undef vars don't really shrink as execution goes.
If there's no advantage, just remove this stuff as it's unneeded complexity.

ir/instr.cpp
aqjune: Added the assertion at undefImpliesUB

ir/instr.cpp
aqjune: Okay. I have a noundef patch that I'm writing (in addition to the malloc), I'll submit a patch to LangRef after that

ir/state.cpp
aqjune: Updated the var name to I.
Regarding try_emplace: `strip_undef_and_add_ub` call below calls State::operator[] as well, and it looks through `analysis.non_undef_vals`. try_emplace causes an interaction between two.
To avoid unnecessary changes at State::operator[], simply find was used instead.

ir/state.cpp
aqjune: Added a guard

ir/state.cpp
aqjune: It was related with correctness: this change makes `tests/alive-tv/undef/br-undef3.srctgt.ll` success.

aqjune:
Single file benchmark: [undefanalysis_res.txt](https://github.com/AliveToolkit/alive2/files/5251559/undefanalysis_res.txt), slight increase in # of success & decrease in # of `smt tactic failed to show goal to be sat/unsat (incomplete quantifiers)` errors

ir/state.cpp
nunoplopes: ah, yes, freeze shrink undef. good.

ir/state.cpp
nunoplopes: sounds good.

PR #490, opened 2020-09-22
Add refinement between undef and non-undef value

### LINKED ISSUE #3: Root checking doesn't handle possible multiple users
Alive2 accepts the following, but shouldn't since there might be multiple users:
```
%x = freeze i4 undef
  =>
%x = undef
```

aqjune:
This patch adds a refinement that if the target's return value is undefined, so should the source's return value be.

I left the description as a comment at `encode_undef_refinement()`.

This patch resolves #3 .

### DIFF
--- a/tools/transform.cpp
+++ b/tools/transform.cpp
@@ -287,13 +287,13 @@ static expr preprocess(Transform &t, const set<expr> &qvars0,
 
 static expr
 encode_undef_refinement_per_elem(
-    const Type &ty, const expr &a, const expr &a2, const expr &a_np,
+    const Type &ty, const StateValue &sva, const expr &a2,
     const expr &b, const expr &b2) {
   const auto *aty = ty.getAsAggregateType();
   if (!aty)
-    return (a_np && a == a2).implies(b == b2);
+    return (sva.non_poison && sva.value == a2).implies(b == b2);
 
-  StateValue sva{expr(a), expr(a_np)}, sva2{expr(a2), expr()};
+  StateValue sva2{expr(a2), expr()};
   StateValue svb{expr(b), expr()}, svb2{expr(b2), expr()};
   expr result = true;
 
@@ -302,9 +302,8 @@ encode_undef_refinement_per_elem(
       continue;
 
     result &= encode_undef_refinement_per_elem(aty->getChild(i),
-        aty->extract(sva, i).value, aty->extract(sva2, i).value,
-        aty->extract(sva, i).non_poison, aty->extract(svb, i).value,
-        aty->extract(svb2, i).value);
+        aty->extract(sva, i), aty->extract(sva2, i).value,
+        aty->extract(svb, i).value, aty->extract(svb2, i).value);
   }
   return result;
 }
@@ -348,7 +347,7 @@ encode_undef_refinement(const Type &type, const State::ValTy &ap,
   expr b2 = b.subst(repls_tgt);
 
   return
-    encode_undef_refinement_per_elem(type, a, a2, ap.first.non_poison, b, b2);
+    encode_undef_refinement_per_elem(type, ap.first, a2, b, b2);
 }
 
 static void
--- a/tools/transform.cpp
+++ b/tools/transform.cpp
@@ -285,6 +285,73 @@ static expr preprocess(Transform &t, const set<expr> &qvars0,
 }
 
 
+static expr
+encode_undef_refinement_per_elem(
+    const Type &ty, const expr &a, const expr &a2, const expr &a_np,
+    const expr &b, const expr &b2) {
+  const auto *aty = ty.getAsAggregateType();
+  if (!aty)
+    return (a_np && a == a2).implies(b == b2);
+
+  StateValue sva{expr(a), expr(a_np)}, sva2{expr(a2), expr()};
+  StateValue svb{expr(b), expr()}, svb2{expr(b2), expr()};
+  expr result = true;
+
+  for (unsigned i = 0; i < aty->numElementsConst(); ++i) {
+    if (aty->isPadding(i))
+      continue;
+
+    result &= encode_undef_refinement_per_elem(aty->getChild(i),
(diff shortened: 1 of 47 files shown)


tools/transform.cpp
```diff
@@ -285,6 +285,73 @@ static expr preprocess(Transform &t, const set<expr> &qvars0,
...
+}
+
+static expr
+encode_undef_refinement(const Type &type, const State::ValTy &ap,
+                        const State::ValTy &bp, set<expr> &qvars) {
+  // Undef refinement: (src-nonpoison /\ src-nonundef) -> tgt-nonundef
+  //
+  // Full refinement formula:
+  //   forall N_tgt, exists N_src,
+  //      (src_nonpoison /\
```
nunoplopes, line 319: the CI build is failing because of this line

tools/transform.cpp
```diff
@@ -285,6 +285,73 @@ static expr preprocess(Transform &t, const set<expr> &qvars0,
...
+        aty->extract(sva, i).value, aty->extract(sva2, i).value,
+        aty->extract(sva, i).non_poison, aty->extract(svb, i).value,
+        aty->extract(svb2, i).value);
+  }
+  return result;
+}
+
+static expr
+encode_undef_refinement(const Type &type, const State::ValTy &ap,
+                        const State::ValTy &bp, set<expr> &qvars) {
```
nunoplopes, line 314: I think we want to pass qvars by value here so the new qvars don't leak to the remaining queries

tools/transform.cpp
```diff
@@ -285,6 +285,73 @@ static expr preprocess(Transform &t, const set<expr> &qvars0,
 }
 
 
+static expr
+encode_undef_refinement_per_elem(
+    const Type &ty, const expr &a, const expr &a2, const expr &a_np,
```
nunoplopes, line 290: it seems easier if you pass a/a_np as a stateValue. It avoid 2x evaluations of extract below

nunoplopes:
What's the impact of this patch in terms of performance?
Any regression on LLVM unit tests?

nunoplopes:
I was also thinking this doesn't take advantage of the state's knowledge of what's undef and not undef. Can we leverage that somehow? (I've some thoughts, but wanted you to think first so we can compare notes)

tools/transform.cpp
aqjune: This is intended because source's new undef vars should be added to qvars.

tools/transform.cpp
aqjune: Fixed

tools/transform.cpp
nunoplopes: ok, but then you are relying on Z3's preprocessor to remove all those useless quantified vars for the other queries. Let's see the benchmarks..

nunoplopes:
> I think the encoding of `ret` instruction successfully does it - it calls `s[*val]`, which applies the result of the analysis.

I meant, it would be nice not to duplicate the two return value expressions.
Now that you mention State::operator[], it really seems like it can be optimized to not rewrite the undef values nor create new undef variables.

aqjune:
> I meant, it would be nice not to duplicate the two return value expressions.
> Now that you mention State::operator[], it really seems like it can be optimized to not rewrite the undef values nor create new undef variables.

Made it as a separate PR here: #491

aqjune:
ph7 is taking a long time (only about half of the pairs are verified after 11 hrs), so I attach the remaining benchmarks first:
[results.txt](https://github.com/AliveToolkit/alive2/files/5266544/results.txt)
BTW, for a few benchmarks, master's result became dramatically better than https://github.com/AliveToolkit/alive2/pull/487#issuecomment-695774711 , and I think this is due to 48984a , which is nice.

aqjune:
LLVM unit tests: no more failures found, has +10 timeouts

tools/transform.cpp
```diff
@@ -285,6 +285,71 @@ static expr preprocess(Transform &t, const set<expr> &qvars0,
...
+  //      (src_nonpoison /\ retval_src(N_src) == retval_src(N_src')) ->
+  //      retval_tgt(N_tgt) == retval_tgt(N_tgt')
+
+  if (dynamic_cast<const VoidType *>(&type))
+    return expr(true);
+
+  vector<pair<expr, expr>> repls_src, repls_tgt;
+  for (auto &u : ap.second) {
+    expr newvar = expr::mkFreshVar("undef", u);
+    repls_src.emplace_back(u, expr::mkFreshVar("undef", u));
```
nunoplopes, line 336: I think there's a typo here. it should use newvar, right?

aqjune: You're right, fixed

aqjune:
Well, let me run bzip2/gzip/oggenc again. :/

aqjune:
There is performance degradation, but it isn't a disaster:
[results.txt](https://github.com/AliveToolkit/alive2/files/5269376/results.txt)

aqjune:
sqlite3 is running, I'll leave it as a comment here.
BTW, I found why ph7 is so slow: the algorithm that finds the list of 'different' functions is too conservative, causing most of the functions in ph7 to be targetted. Needs to be revised; this will make running other benchmarks faster too

aqjune:
sqlite3: [results-sqlite3.txt](https://github.com/AliveToolkit/alive2/files/5272792/results-sqlite3.txt)

regehr:
nice! I have a very big machine that's mostly idle at the moment, should I start running some of these?

aqjune:
For the single file benchmarks - I think using the machines here is enough.
But, we may need an experiment that shows Alive2 does not miss a bug. Many reviewers pointed out that in the past.
If we can generate many pairs that are not equivalent and test them, this should be better. However, I have no idea how to do this...
We can simply report that we reproduced bugs from Bugzilla and found no false negatives, but this might be slightly weak, because the number of reproducible pairs from Bugzilla isn't many.

aqjune:
This should be applied to inputs of function calls as well.
I'll prepare a patch for this.

PR #491, opened 2020-09-23
Don't add undef vars if analysis says it's never undef

aqjune:


nunoplopes:
This was not the optimization I had in mind in #490.
But, ok, can you explain why is this patch correct? Not rewriting undef is not correct in general, so I'm surprised of this patch is.

aqjune:
Eh, realized that I was misunderstanding the behavior of operator[]. :( I thought it was returning an old expression and then substituting the variables.
I think I now understand the intended optimization - when there are two ret instructions on different branches, substitution may not be needed, is this right?

nunoplopes:
I think we can think of a few optimizations:

1) we know a value can't be undef. We should take this opportunity to clear the undef vars. This is because we know that whatever value we set an under var to, we will always get the same expression value (otherwise we triggered UB before). This needs more thought to prove the cases when it's correct to do so.

2) We know a return value can't be undef (because it was marked before). So the refinement check can leverage this information to avoid checking whether it's undef or no as we already know it. I guess this requires building yet another expression, say is_return_undef, that gets constructed during symbolic execution that states for which return instructions the value is undef.

3) combine undefs of multiple return stmts as you suggested (this optimization is pending in the new memory branch for return memory as well). The motivation is that we only look at a specific return branch at a time, so we can share undef variables between different branches and save some variables. Sharing will happen if there are common sub-expressions in return values of different stmts. This optimization requires doing the undef rewrite later, not at return stmt evaluation I guess.

I think any of 3 optimizations above is worth experimenting with if the results for undef refinement aren't good. Number 1) is good regardless of undef refinement.

aqjune:
All of the optimizations sound great to me, I'm willing to work on these!

Before working on these, I think it is needed to shorten the VCGen time first, because it is likely that the optimizations will involve several iterations of running single file benchmarks. Including #434 (which should be updated) I have a few things that I'd like to experiment first.

aqjune:
This PR is updated to implement the first optimization. Test is running

nunoplopes:
Can you articulate why you think this is correct? It's not obvious to me why so. We need an argument that mentions all the cases when we assume some value is not undef and how quantifiers look like for each of the refinement queries.

aqjune:
My reasoning is that, If a value depends on undef variables, UB refinement already should have detected the case. So, once the UB refinement has passed, the value should be constant regardless of undef variables. Therefore, we can omit what we had to do with undef variables inside the expression.

For example:
```
; x = ite(is_x_undef, undef1, x)
f(i1 cond, i32 x) { 
  br cond, BB_TRUE, BB_FALSE
BB_TRUE:
  call g(noundef x)
  ret x
BB_FALSE:
  ret x
}
```
Without this patch, the return value is: `ite(cond, ite(is_x_undef, undef1, x), ite(is_x_undef, undef2, x))`, and {undef1, undef2} are quantified. (`strip_undef` can optimize further in this specific case, but let's ignore it)

However, with this patch, the return value becomes `ite(cond, ite(is_x_undef, undef1, x), ite(is_x_undef, undef1, x))`, and I think it's fine because the `undef1` in `cond == true` case isn't used.

When `cond == true`,  `is_x_undef` has to be false due to the UB refinement check.

aqjune:
Performance result:
[compare-master.f413e9d-optimizeundefrewrite.zip](https://github.com/AliveToolkit/alive2/files/5289405/compare-master.f413e9d-optimizeundefrewrite.zip)

aqjune:
Now we have ph7 running too: [ph7.txt](https://github.com/AliveToolkit/alive2/files/5308290/ph7.txt)
Its correct: timeout ratio is about 1:2, which is slightly higher than other benchmarks.

nunoplopes:
This test fails with this patch:
```
define i8 @src() {
  %a = and i8 undef, 1
  %c = icmp eq i8 %a, 0
  br i1 %c, label %then, label %else

then:
  ret i8 0

else:
  ret i8 1
}

define i8 @tgt() {
  ret i8 2
}
```

I'll add it to the test suite to avoid other regressions.

nunoplopes:
sorry, my bad. I had a patch locally that interfered with yours. This PR passes that test.
Let me review this again.

nunoplopes:
I added an assert to check whether all undef vars ended up in quant vars, and it failed for a couple unit tests where an input is just used as a function argument.
To be on the safe side I suggest we always add the undef vars to quant vars, e.g.:
```diff
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -215,6 +217,12 @@ const StateValue& State::operator[](const Value &val) {
     return *sv;
   };

+  if (is_non_undef) {
+    // We don't need to add uvar to undef_vars
+    quantified_vars.insert(uvars.begin(), uvars.end());
+    return simplify(sval, true);
+  }
+
   if (uvars.empty() || !used || disable_undef_rewrite) {
     used = true;
     undef_vars.insert(uvars.begin(), uvars.end());
```

What do you think?

nunoplopes:
the issue is that if a source undef var doesn't end up with a forall quantifier it' pretty weird. It's likely wrong, and I don't want to think about all corner cases. Adding those vars to quant vars makes it safely correct.

aqjune:
Yes, it seems weird... I added the safety net. Maybe working on fncall refinement to deal with undef inputs might reveal the bug (which is on my to-do list), I hope.

PR #496, opened 2020-10-02
Add -smt-random-seed

aqjune:
This PR adds a `-smt-random-seed` parameter that sets the value of `smt.random_seed` of Z3.

### DIFF
--- a/tools/alive-tv.cpp
+++ b/tools/alive-tv.cpp
@@ -63,6 +63,11 @@ static llvm::cl::opt<unsigned> opt_smt_to(
     "smt-to", llvm::cl::desc("Timeout for SMT queries (default=1000)"),
     llvm::cl::init(1000), llvm::cl::value_desc("ms"), llvm::cl::cat(opt_alive));
 
+static llvm::cl::opt<unsigned> opt_smt_random_seed(
+    "smt-random-seed",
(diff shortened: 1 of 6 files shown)


tools/alive.cpp
```diff
@@ -59,6 +59,8 @@ int main(int argc, char **argv) {
       show_smt_stats = true;
     else if (arg.compare(0, 8, "-smt-to:") == 0 && arg.size() > 8)
       smt::set_query_timeout(arg.substr(8).data());
+    else if (arg.compare(0, 17, "-smt-random-seed:") == 0 && arg.size() > 17)
```
nunoplopes, line 62: please add this to the help above pls

tools/alive-tv.cpp
```diff
@@ -63,6 +63,11 @@ static llvm::cl::opt<unsigned> opt_smt_to(
     "smt-to", llvm::cl::desc("Timeout for SMT queries (default=1000)"),
     llvm::cl::init(1000), llvm::cl::value_desc("ms"), llvm::cl::cat(opt_alive));
 
+static llvm::cl::opt<unsigned> opt_smt_random_seed(
```
nunoplopes, line 66: can you add the same to tv/tv.cpp pls?

Issue #492, opened 2020-09-26
Implement Z3 perf regression tracking


nunoplopes:
Z3 4.8.9 regressed perf compared to 4.8.8 in a few benchmarks (e.g., just see CI bots, both linux & Mac). It would be nice to have an automated way of diffing new timeouts in unit tests & LLVM test suite and then produce Z3 logs using `Z3_open_log` and friends. Similarly, we could have an automated way of running John's perl script to detect score degradations with new Z3 releases.

@NikolajBjorner says he'll fix any reported regression in 24 hours 😁

aqjune:
Would it be great if I build a small webpage that tracks this?

nunoplopes:
yes, I think that would be very useful, but it's not urgent 😀
It has a bit that is not obvious is that it needs to account for SMT random seeds. Z3 doesn't consider perf bugs if changing the seed makes it solve the problem. So we would need to sample a few seeds for each test and version.

NikolajBjorner:
If the average run-time over _several_ random seeds is higher, then it is still a perf regression.

aqjune:
Okay, maybe I can pick a few random seeds (well, randomly) and show the individual results as well as the average result.

NikolajBjorner:
As Nuno wrote use a few samples. The seeds don't need to be random, just numbers 1-10 will do the trick.

```
for i in range(10):
     os.system("z3.exe smt.random_seed=%d %s -st" % (i, file))
```

aqjune:
I made this page: http://147.46.242.54:15243
It tracks the performance of Alive2 w.r.t Z3 when running LLVM unit tests. The detailed page shows in which tests timeout ratios are high (http://147.46.242.54:15243/show/8985651-af3789a-6cc52e0).
Currently, it has only one slot filled but I'm going to run a batch of tests & fill the remaining slots. Also, there isn't `diff two results` page yet; I'll make the page, eh, maybe after the preparation of Q&A for my talk :)

aqjune:
Okay, diff is now [working](http://147.46.242.54:15243/compare/8985651-af3789a-6cc52e0/18a9159-af3789a-6cc52e0)! Dashboard is [here](http://147.46.242.54:15243).
Log dump can be downloaded from the statistics page for the specific configuration (e.g., [here](http://147.46.242.54:15243/show/8985651-af3789a-6cc52e0)).
I'll run experiment with the fresh Z3 1~2 times per week.

regehr:
[...]

PR #497, opened 2020-10-03
Let tv pass the custom z3 log file path to start_logging

aqjune:


aqjune:
This patch is needed to generate z3 logs in advance, which is helpful especially when many tvs are run in parallel.
I'm going to add log dump downloads to the homepage described at https://github.com/AliveToolkit/alive2/issues/492#issuecomment-703329013 too.

smt/smt.h
```diff
@@ -27,6 +27,6 @@ void set_memory_limit(uint64_t limit);
 bool hit_memory_limit();
 bool hit_half_memory_limit();
 
-void start_logging();
+void start_logging(std::string path = "z3_log.txt");
```
nunoplopes, line 30: can we change this to const char* pls to avoid copies?

PR #501, opened 2020-10-07
Implementing Tarjan's Loop Analysis Algorithm

zhengyangl:
WIP

### DIFF
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -481,7 +481,7 @@ void LoopAnalysis::analysis() {
   unsigned bb_count = f.getBBs().size();
 
   auto isAncestor = [this](unsigned w, unsigned v) -> bool {
-    return w <= v && v <= last.at(w);
+    return w <= v && v <= last[w];
   };
 
   vector<set<unsigned>> nonBackPreds(bb_count), backPreds(bb_count);
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -4,6 +4,8 @@
 #include "ir/function.h"
 #include "ir/instr.h"
 #include "util/errors.h"
+#include "util/unionfind.h"
+#include <fstream>
 
 using namespace smt;
 using namespace util;
@@ -289,6 +291,14 @@ void Function::print(ostream &os, bool print_header) const {
     os << "}\n";
 }
 
+void Function::unroll(unsigned k) {
+  if (k == 0)
+    return;
+  LoopAnalysis la(*this);
+  ofstream out("a.gv");
+  la.printDot(out);
+}
+
 ostream& operator<<(ostream &os, const Function &f) {
   f.print(os);
   return os;
@@ -425,6 +435,122 @@ void DomTree::printDot(std::ostream &os) const {
     }
   }
 
+  os << "}\n";
+}
+
+
+void LoopAnalysis::getDepthFirstSpanningTree() {
+  unsigned bb_count = f.getBBs().size();
+  node.resize(bb_count, nullptr);
+  last.resize(bb_count, -1u);
+
+  map<const BasicBlock*, vector<const BasicBlock*>> edges;
+  for (auto [src, dst, instr] : cfg) {
+    edges[&src].push_back(&dst);
+  }
+
+  unsigned current = 0;
+  std::vector<std::pair<const BasicBlock*,bool>> s;
+  s.emplace_back(&f.getFirstBB(), false);
+  while(!s.empty()) {
+    auto &[bb, flag] = s.back();
+    s.pop_back();
+
+    if (flag) {
+      last[number[bb]] = current - 1;
+    } else {
+      node[current] = bb;
+      number[bb] = current++;
+
+      s.emplace_back(bb, true);
+      for (auto tgt : edges[bb])
+        if (!number.count(tgt))
+          s.emplace_back(tgt, false);
+    }
+  }
+}
+
+// Implemention of Tarjan-Havlak algorithm.
+//
+// Irreducible loops are partially supported.
+//
+// Tarjan, R. (1974). Testing Flow Graph Reducibility.
+// Havlak, P. (1997). Nesting of reducible and irreducible loops.
+void LoopAnalysis::analysis() {
+  getDepthFirstSpanningTree();
+  unsigned bb_count = f.getBBs().size();
+
+  auto isAncestor = [this](unsigned w, unsigned v) -> bool {
+    return w <= v && v <= last.at(w);
+  };
+
+  vector<set<unsigned>> nonBackPreds(bb_count), backPreds(bb_count);
+  header.resize(bb_count, 0);
+  type.resize(bb_count, NodeType::nonheader);
+
+  for (auto [src, dst, instr] : cfg) {
+    unsigned v = number.at(&src), w = number.at(&dst);
+    if (isAncestor(w, v))
+      backPreds[w].insert(v);
+    else
+      nonBackPreds[w].insert(v);
+  }
+
+  UnionFind uf(bb_count);
+
+  for (unsigned w = bb_count - 1; w != -1u; --w) {
+    set<unsigned> P;
+    for (unsigned v : backPreds[w])
+      if (v != w)
+        P.insert(uf.find(v));
+      else
+        type[w] = NodeType::self;
+
+    if (!P.empty())
+      type[w] = NodeType::reducible;
+
+    set<unsigned> workList(P);
+    while (!workList.empty()) {
+      unsigned x = *workList.begin();
+      workList.erase(x);
+
+      for (unsigned y : nonBackPreds[x]) {
+        unsigned yy = uf.find(y);
+        if (!isAncestor(w, yy)) {
+          type[w] = NodeType::irreducible;
+          nonBackPreds[w].insert(yy);
+        } else if (yy != w && !P.count(yy)) {
+          P.insert(yy);
+          workList.insert(yy);
+        }
+      }
+    }
+
+    for (unsigned x : P) {
+      header[x] = w;
+      uf.merge(x, w);
+    }
+  }
+}
+
+void LoopAnalysis::printDot(ostream &os) const {
+  os << "digraph {\n";
+  for (unsigned i = 0, e = f.getBBs().size(); i != e; ++i) {
+    if (type[i] == NodeType::nonheader)
+      continue;
+    os << '"' << bb_dot_name(node[i]->getName())
+       << "\" [shape=circle]\n";
+  }
+  for (unsigned i = 0, e = f.getBBs().size(); i != e; ++i) {
+    // do not draw self loop for root
+    if (header[i] == i)
+      continue;
+    if (type[i] == NodeType::nonheader ||
+        type[header[i]] == NodeType::nonheader)
+      continue;
+    os << '"' << bb_dot_name(node[header[i]]->getName()) << "\" -> \""
+       << bb_dot_name(node[i]->getName()) << "\";\n";
+  }
   os << "}\n";
 }
 
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -450,7 +450,7 @@ void LoopAnalysis::getDepthFirstSpanningTree() {
   }
 
   unsigned current = 0;
-  std::vector<std::pair<const BasicBlock*,bool>> s;
+  vector<pair<const BasicBlock*,bool>> s;
   s.emplace_back(&f.getFirstBB(), false);
   while(!s.empty()) {
     auto &[bb, flag] = s.back();
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -255,6 +255,14 @@ bool Function::removeUnusedStuff(const multimap<Value*, Value*> &users,
   return changed;
 }
 
+void Function::unroll(unsigned k) {
+  if (k == 0)
+    return;
+  LoopAnalysis la(*this);
+  ofstream out("a.gv");
+  la.printDot(out);
+}
+
 void Function::print(ostream &os, bool print_header) const {
   {
     const auto &gvars = getGlobalVars();
@@ -289,14 +297,6 @@ void Function::print(ostream &os, bool print_header) const {
 
   if (print_header)
     os << "}\n";
-}
-
-void Function::unroll(unsigned k) {
-  if (k == 0)
-    return;
-  LoopAnalysis la(*this);
-  ofstream out("a.gv");
-  la.printDot(out);
 }
 
 ostream& operator<<(ostream &os, const Function &f) {
--- a/ir/function.h
+++ b/ir/function.h
@@ -156,6 +156,8 @@ public:
   bool removeUnusedStuff(const std::multimap<Value*, Value*> &users,
                          const std::vector<std::string_view> &src_glbs);
 
+  void unroll(unsigned k);
+  
   void print(std::ostream &os, bool print_header = true) const;
   friend std::ostream &operator<<(std::ostream &os, const Function &f);
 };
@@ -210,6 +212,24 @@ class DomTree final {
     DomTree(Function &f, CFG &cfg) : f(f), cfg(cfg) { buildDominators(); }
     const BasicBlock* getIDominator(const BasicBlock &bb) const;
     void printDot(std::ostream &os) const;
+};
+
+class LoopAnalysis final {
+  Function &f;
+  CFG cfg;
+
+  std::map<const BasicBlock*, unsigned> number;
+  std::vector<const BasicBlock*> node;
+  std::vector<unsigned> last;
+  void getDepthFirstSpanningTree();
+
+  std::vector<unsigned> header;
+  enum NodeType { nonheader, self, reducible, irreducible };
+  std::vector<NodeType> type;
+  void analysis();
+public:
+  LoopAnalysis(Function &f) : f(f), cfg(f) { analysis(); }
+  void printDot(std::ostream &os) const;
 };
 
 }
--- a/tools/alive-tv.cpp
+++ b/tools/alive-tv.cpp
@@ -94,6 +94,16 @@ static llvm::cl::opt<std::string> opt_tgt_fn(
     "tgt-fn", llvm::cl::desc("Name of tgt function (without @)"),
     llvm::cl::cat(opt_alive), llvm::cl::init("tgt"));
 
+static llvm::cl::opt<unsigned> opt_src_unrolling_factor(
+    "src-unroll",
+    llvm::cl::desc("Unrolling factor for src function (default=0)"),
+    llvm::cl::cat(opt_alive), llvm::cl::init(0));
+
+static llvm::cl::opt<unsigned> opt_tgt_unrolling_factor(
+    "tgt-unroll",
+    llvm::cl::desc("Unrolling factor for tgt function (default=0)"),
+    llvm::cl::cat(opt_alive), llvm::cl::init(0));
+
 static llvm::cl::opt<bool> opt_tactic_verbose(
     "tactic-verbose", llvm::cl::desc("SMT Tactic verbose mode"),
     llvm::cl::cat(opt_alive), llvm::cl::init(false));
@@ -283,6 +293,7 @@ static void compareFunctions(llvm::Function &F1, llvm::Function &F2,
     ++errorCount;
     return;
   }
+  Func1->unroll(opt_src_unrolling_factor);
 
   auto Func2 = llvm2alive(F2, llvm::TargetLibraryInfoWrapperPass(targetTriple)
                                     .getTLI(F2), Func1->getGlobalVarNames());
@@ -292,6 +303,7 @@ static void compareFunctions(llvm::Function &F1, llvm::Function &F2,
     ++errorCount;
     return;
   }
+  Func2->unroll(opt_tgt_unrolling_factor);
 
   smt_init->reset();
   Transform t;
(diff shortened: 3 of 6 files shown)


zhengyangl:
@nunoplopes Hi Nuno, this is the code for the Tarjan's loop analysis algorithm, for now given a nested loop it could print the loop nested tree.

For example, https://alive2.llvm.org/ce/z/TimNv1 , here is the generated tree:

a

_n. = non-header; r. = reducible; s. = self_

Another one, https://alive2.llvm.org/ce/z/JQUCtq , the generated tree:

a

kevindevos:
I should already have loop unroll implemented very soon, i'm currently going through some tests that it flagged as wrong. Maybe it's best to wait and see what @nunoplopes says?

tools/alive-tv.cpp
```diff
@@ -284,6 +284,8 @@ static void compareFunctions(llvm::Function &F1, llvm::Function &F2,
     return;
   }
 
+  Func1->unroll();
```
nunoplopes, line 287: I guess this is a debug leftover?

ir/function.h
```diff
@@ -157,6 +157,7 @@ class Function final {
                          const std::vector<std::string_view> &src_glbs);
 
   void print(std::ostream &os, bool print_header = true) const;
+  void unroll();
```
nunoplopes, line 160: this shouldn't be placed in the middle of the 2 printing methods. also, it should receive k as argument.

ir/function.h
```diff
@@ -212,4 +213,43 @@ class DomTree final {
...
+  enum NodeType { nonheader, self, reducible };
+  std::vector<NodeType> type;
+  void printNodeType(std::ostream &os, const NodeType &t) const;
+  void analysis();
+public:
+  LoopAnalysis(Function &f) : f(f), cfg(f), dfst(f, cfg) { analysis(); }
+  void printDot(std::ostream &os) const;
+};
+
+class DisjointSet final {
```
nunoplopes, line 247: this is a generic datastructure. should be moved to utils.
Also, is this a union-find? if so, please change the name to make that explicit.

util/unionfind.h
```diff
@@ -0,0 +1,19 @@
...
+
+namespace util {
+
+class UnionFind final {
+  std::vector<unsigned> id;
+  unsigned root(unsigned i);
+public:
+  UnionFind(unsigned n);
+  unsigned find(unsigned i);
+  void unite(unsigned p, unsigned q);
```
nunoplopes, line 16: this is usually spelled union()

tools/alive-tv.cpp
```diff
@@ -94,6 +94,16 @@ static llvm::cl::opt<std::string> opt_tgt_fn(
     "tgt-fn", llvm::cl::desc("Name of tgt function (without @)"),
     llvm::cl::cat(opt_alive), llvm::cl::init("tgt"));
 
+static llvm::cl::opt<unsigned> opt_src_unrolling_factor(
+    "src-unrolling-factor",
```
nunoplopes, line 98: maybe `-src-unroll` is sufficient

util/unionfind.h
zhengyangl: Oh I was using unite to avoid the conflict with the keyword union. What should we call it, is uf_union better?

nunoplopes: ah, right 😵
then let's leave it like that 👍

ir/function.cpp
```diff
@@ -428,5 +438,115 @@ void DomTree::printDot(std::ostream &os) const {
...
+void LoopAnalysis::printNodeType(ostream &os, const NodeType &t) const {
+  switch (t) {
+  case nonheader: os<<"n."; break;
+  case self:      os<<"s."; break;
+  case reducible: os<<"r."; break;
+  }
+}
+
+// Tarjan, R. (1974).
+// Testing Flow Graph Reducibility.
```
zhengyangl, line 472: Reason for citing this rather than Havlak's: Both paper are describing the same algorithm except that Havlak's has extra steps for handling irreducible loops. As we won't support irreducible loops, it would be better to cite the original paper, imo.

We could change it back to Havlak's if someday we implemented the full algorithm.

regehr: you should cite both papers and say the same thign in the comment that you said here!

ir/function.cpp
```diff
@@ -428,5 +438,115 @@ void DomTree::printDot(std::ostream &os) const {
...
+    set<unsigned> workList(P);
+
+    while (!workList.empty()) {
+      unsigned x = *workList.begin();
+      workList.erase(x);
+
+      for (unsigned y : nonBackPreds[x]) {
+        unsigned yy = uf.find(y);
+        if (!isAncestor(w, yy)) {
+          UNREACHABLE();
```
zhengyangl, line 519: Nuno: what's the best way for terminating with an error message in Alive2, is assert(0 && "irreducible loop detected"); a good choice here?

nunoplopes: the code shouldn't crash with irreducible loops. It can produce a sub-optimal result, but shouldn't crash.

regehr: also, never use assert for error handling (although I believe we leave asserts enabled even in release builds). if you were going to exit here, you could do something like 
```   
 if (has_failure) {
      cerr << "Alive2: Transform doesn't verify; aborting!" << endl;
      exit(1);
    }
```

zhengyangl: @regehr correct!
@nunoplopes Need to think if it is safe to remove this line, or in another word, if continue going without break could give wrong result.

nunoplopes: maybe write the 3 missing lines from Havlak and call it a day?

nunoplopes:
btw please have a look at the Mac build bot. It's failing with warnings in the code.

zhengyangl:
For the test case @nunoplopes  mentioned in email:

```
void loop42(int a, int b, int c, int d, int e) {
  for (int aa = 0; aa < a; aa++) {
    for (int bb = 0; bb < b; bb++) {
      f();
    }
    for (int bb = 0; bb < c; bb++) {
      f();
    }
    for (int bb = 0; bb < d; bb++) {
      f();
    }
  }
  for (int bb = 0; bb < e; bb++) {
    f();
  }
}
```

The generated dot for -O0 (Havlak's style, will update the dot printer to show more information):
a

It looks correct to me. I Will try more cases.

zhengyangl:
Source
```
void loop2d(int a, int b, int c, int d, int e)  {
  for (int aa = 0; aa < a; aa++) {
    for (int bb = 0; bb < b; bb++) {
      for (int cc = 0; cc < c; cc++) {
        for (int dd = 0; dd < d; dd++) {
          if (f())
            break;
          for (int ee = 0; ee < e; ee++) {
            f();
          }
        }
      }
    }
  }
}
```
dot:
a

zhengyangl:
@nunoplopes, could you please tell me how the loop tree should be drawed one more time? The current printDot draws the tree exactly same as Havlak's paper's one (except for the Start/End block, which should always at the same position for all trees).

regehr:
I believe that graphviz has good support for boxes like this

zhengyangl:
@regehr So it would be a CFG with loop information annotated as boxes, such as the diagram below, am I understanding correctly? 
image

nunoplopes:
> @nunoplopes, could you please tell me how the loop tree should be drawed one more time? The current printDot draws the tree exactly same as Havlak's paper's one (except for the Start/End block, which should always at the same position for all trees).

only loop headers should appear. e.g.:
```
for (..) {   // header 1
 for () {    // header 2
   for() {}  // header 3
  }
 for() {}    // header 4
}
```

The graph only has 4 vertexes, connected as follows:
h1 ->h2
h2 ->h3
h1 ->h4

nunoplopes:
> @regehr So it would be a CFG with loop information annotated as boxes, such as the diagram below, am I understanding correctly?

Yep, for the CFG one you can use boxes like that to demark the loops. Personally I don't like that picture from Havlak's paper. I prefer either the loop nesting tree or an annotated CFG, where you can color the back edges, put boxes around loops, color loop headers/exits, etc. (all the info you need for unrolling)
Anyway, at least getting the loop nest info is important for unrolling.

zhengyangl:
> > @nunoplopes, could you please tell me how the loop tree should be drawed one more time? The current printDot draws the tree exactly same as Havlak's paper's one (except for the Start/End block, which should always at the same position for all trees).
> 
> only loop headers should appear. e.g.:
> 
> ```
> for (..) {   // header 1
>  for () {    // header 2
>    for() {}  // header 3
>   }
>  for() {}    // header 4
> }
> ```
> 
> The graph only has 4 vertexes, connected as follows:
> h1 ->h2
> h2 ->h3
> h1 ->h4

I've updated the printDot to only print loop headers. All the dot diagrams are updated as well :)

regehr:
please can you post a graph for one of the hard examples in the file I sent? :)

zhengyangl:
> Source
> 
> ```
> void loop2d(int a, int b, int c, int d, int e)  {
>   for (int aa = 0; aa < a; aa++) {
>     for (int bb = 0; bb < b; bb++) {
>       for (int cc = 0; cc < c; cc++) {
>         for (int dd = 0; dd < d; dd++) {
>           if (f())
>             break;
>           for (int ee = 0; ee < e; ee++) {
>             f();
>           }
>         }
>       }
>     }
>   }
> }
> ```
> 
> dot:
> image

This one for loop2d

zhengyangl:
source:
```
void loop3e(int a, int b, int c, int d, int e)  {
  for (int aa = 0; aa < a; aa++) {
    for (int bb = 0; bb < b; bb++) {
      for (int cc = 0; cc < c; cc++) {
        for (int dd = 0; dd < d; dd++) {
          for (int ee = 0; ee < e; ee++) {
            f();
            if (cond())
              break;
          }
          if (cond())
            break;
        }
        if (cond())
          break;
      }
      if (cond())
        break;
    }
    if (cond())
      break;
  }
}
```
image

zhengyangl:
@regehr loop4d contains irreducible loops, the nesting tree result is not showing expected 8->9->22->11->13. The reason could be either the sub-optimality of the _partial_ Havlak's algorithm or the ambiguity in identifying the loops. Lemme dig more.

```
void loop4b(int a, int b, int c, int d, int e)  {
  int aa, bb;
  if (cond()) {
    aa = bb = 0;
    goto l1;
  }
  for (aa = 0; aa < a; aa++) {
    for (bb = 0; bb < b; bb++) {
    l1:
      for (int cc = 0; cc < c; cc++) {
        for (int dd = 0; dd < d; dd++) {
          for (int ee = 0; ee < e; ee++) {
            f();
          }
        }
      }
    }
  }
}
```

a

regehr:
I want to see the boxes please!

util/unionfind.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+UnionFind::UnionFind(unsigned n) {
+  for (unsigned i = 0; i < n; ++i) {
+    id.push_back(i);
+  }
+}
+
+unsigned UnionFind::root(unsigned i) {
+  while (i != id[i]) i = id[i];
+  return i;
```
aqjune, line 19: Can we update id[i] to the new root?
This is a known technique that helps further flatten the tree ([link](https://en.wikipedia.org/wiki/Disjoint-set_data_structure#Finding_set_representatives)).
```
unsigned UnionFind::root(unsigned i0) {
   unsigned i = i0;
   while (i != id[i]) i = id[i];
   id[i0] = i;
   return i;
}
```

aqjune: To be precise, all nodes on the path from i0 to the root can be updated to have the found root.

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
   os << "}\n";
 }
 
+
+void LoopAnalysis::getDepthFirstSpanningTree() {
+  std::map<const BasicBlock*, std::vector<const BasicBlock*>> edges;
```
nunoplopes, line 443: we don't use std:: in c++ files

ir/function.h
```diff
@@ -212,4 +214,23 @@ class DomTree final {
     void printDot(std::ostream &os) const;
 };
 
+class LoopAnalysis final {
+  Function &f;
+  CFG cfg;
+
+  std::map<const BasicBlock*, unsigned> number;
+  std::vector<const BasicBlock*> node;
+  std::map<unsigned, unsigned> last;
```
nunoplopes, line 223: this can be changed to a vector

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+  unsigned current = 0;
+  for (auto [src, dst, instr] : cfg) {
+    edges[&src].push_back(&dst);
+  }
+
+  std::function<void(const BasicBlock *)> dfs = [&](const BasicBlock *bb) {
```
nunoplopes, line 449: please make the DFS iterative.

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+    node.push_back(bb);
+    number[bb] = current++;
+    for (auto tgt : edges [bb]) {
+      if (number.find(tgt) == number.end()) {
```
nunoplopes, line 453: !number.count(tgt)

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+  std::map<const BasicBlock*, std::vector<const BasicBlock*>> edges;
+  unsigned current = 0;
+  for (auto [src, dst, instr] : cfg) {
+    edges[&src].push_back(&dst);
+  }
+
+  std::function<void(const BasicBlock *)> dfs = [&](const BasicBlock *bb) {
+    node.push_back(bb);
+    number[bb] = current++;
+    for (auto tgt : edges [bb]) {
```
nunoplopes, line 452: no space before []

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+}
+
+void LoopAnalysis::printDot(ostream &os) const {
+  os << "digraph {\n";
+  for (unsigned i = 0; i < f.getBBs().size(); ++i) {
+    if (type[i] == NodeType::nonheader)
+      continue;
+    os << '"' << bb_dot_name(node[i]->getName())
+       << "\" [xlabel=\"";
+    printNodeType(os, type[i]);
```
nunoplopes, line 547: I think this node type isn't need anymore. it looks ugly anyway and they are all of equal type.

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+    for (unsigned x : P) {
+      header[x] = w;
+      uf.unite(x, w);
+    }
+  }
+}
+
+void LoopAnalysis::printDot(ostream &os) const {
+  os << "digraph {\n";
+  for (unsigned i = 0; i < f.getBBs().size(); ++i) {
```
nunoplopes, line 542: i = i, e = f.getBBs().size(); i != e

util/unionfind.cpp
```diff
@@ -0,0 +1,32 @@
...
+namespace util {
+
+UnionFind::UnionFind(unsigned n) {
+  for (unsigned i = 0; i < n; ++i) {
+    id.push_back(i);
+  }
+}
+
+unsigned UnionFind::root(unsigned i) {
+  while (i != id[i]) i = id[i];
```
nunoplopes, line 18: use newlines please

ir/function.h
```diff
@@ -212,4 +214,23 @@ class DomTree final {
     void printDot(std::ostream &os) const;
 };
 
+class LoopAnalysis final {
+  Function &f;
+  CFG cfg;
+
+  std::map<const BasicBlock*, unsigned> number;
```
nunoplopes, line 221: do we need this stuff to be in the header? Most of this stuff seems to be private to the implementation that is run ocne and the clients of this class never lookup this information?

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+  unsigned bb_count = f.getBBs().size();
+
+  auto isAncestor = [this](unsigned w, unsigned v) -> bool {
+    return w <= v && v <= last.at(w);
+  };
+
+  vector<set<unsigned>> nonBackPreds, backPreds;
+
+  for (unsigned w = 0; w < bb_count; ++w) {
+    nonBackPreds.push_back({});
```
nunoplopes, line 489: just initialize these vectors with the right size beforehand

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+    backPreds.push_back({});
+    header.push_back(0);
+    type.push_back(NodeType::nonheader);
+    for (auto [src, dst, instr] : cfg) {
```
nunoplopes, line 493: after all those vector initializations are moved out, this loop can be denested. Right now you have a quadratic loop here that can be avoided (I know it matches the paper, but since we don't have an efficient lookup from dst to src we should avoid being quadratic in the number of BBs).

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+        if (isAncestor(w, v))
+          backPreds[w].insert(v);
+        else
+          nonBackPreds[w].insert(v);
+      }
+    }
+  }
+  UnionFind uf(bb_count);
+
+  for (unsigned w = bb_count - 1; w != static_cast<unsigned>(-1); --w) {
```
nunoplopes, line 506: you can use -1u

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+    set<unsigned> P;
+    for (unsigned v : backPreds[w])
+      if (v != w)
+        P.insert(uf.find(v));
+      else
+        type[w] = NodeType::self;
+
+    if (!P.empty()) type[w] = NodeType::reducible;
```
nunoplopes, line 514: newline

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+    while (!workList.empty()) {
+      unsigned x = *workList.begin();
+      workList.erase(x);
+
+      for (unsigned y : nonBackPreds[x]) {
+        unsigned yy = uf.find(y);
+        if (!isAncestor(w, yy)) {
+          type[w] = NodeType::irreducible;
+          nonBackPreds[w].insert(yy);
+        } else if (yy != w && P.find(yy) == P.end()) {
```
nunoplopes, line 526: !P.count(yy)

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+  os << "digraph {\n";
+  for (unsigned i = 0; i < f.getBBs().size(); ++i) {
+    if (type[i] == NodeType::nonheader)
+      continue;
+    os << '"' << bb_dot_name(node[i]->getName())
+       << "\" [xlabel=\"";
+    printNodeType(os, type[i]);
+    os << "\", shape=circle]\n";
+  }
+  for (unsigned i = 0; i < f.getBBs().size(); ++i) {
```
nunoplopes, line 550: same here

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+        } else if (yy != w && P.find(yy) == P.end()) {
+          P.insert(yy);
+          workList.insert(yy);
+        }
+      }
+    }
+
+    for (unsigned x : P) {
+      header[x] = w;
+      uf.unite(x, w);
```
nunoplopes, line 535: gah, unite sounds weird to me.. Maybe merge instead?

util/unionfind.h
```diff
@@ -0,0 +1,19 @@
...
+// Copyright (c) 2018-present The Alive2 Authors.
+// Distributed under the MIT license that can be found in the LICENSE file.
+
+#include <vector>
+
+namespace util {
+
+class UnionFind final {
+  std::vector<unsigned> id;
+  unsigned root(unsigned i);
```
nunoplopes, line 12: we can rid of this method; it's duplicated for no reason

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+  case nonheader:   os<<"n."; break;
+  case self:        os<<"s."; break;
+  case reducible:   os<<"r."; break;
+  case irreducible: os<<"i."; break;
+  }
+}
+
+// Implemention of Tarjan-Havlak algorithm.
+//
+// Irreducible loops are partially supported yet.
```
nunoplopes, line 474: are only partially supported

ir/function.cpp
```diff
@@ -428,5 +438,127 @@ void DomTree::printDot(std::ostream &os) const {
...
+        dfs(tgt);
+      }
+    }
+    last[number[bb]] = current - 1;
+  };
+
+  dfs(&f.getFirstBB());
+}
+
+void LoopAnalysis::printNodeType(ostream &os, const NodeType &t) const {
```
nunoplopes, line 463: I think this can go

nunoplopes (changes requested):

ir/function.h
zhengyangl: Yes, all those three containers are used later in the analysis() and printDot() since DFST numbers are required to check ancestor relation between a pair of vertices. To get rid of those class fields I could either inline the getDFST() inside analysis(), or pass those containers as return values. What do you think @nunoplopes ?

zhengyangl:
@nunoplopes @aqjune all the reviews are addressed except the one I just commented, please let me know if there's any other concerns. :)

nunoplopes (approved):

PR #504, opened 2020-10-09
Reuse undef variables if allowed

aqjune:
This patch implements the third one of the optimizations described at https://github.com/AliveToolkit/alive2/pull/491#issuecomment-697299407

Given a program
```
br i1 cond, A, B
A:
  %v1 = undef
  ...
  br C
B:
  %v2 = undef
  ...
  br C
C:
  %v = phi (%v1, %v2)
```

The undef values at %v1 and %v2 can be encoded with the same SMT variable.
The reason is that, if cond is either true or false, only one of the branches is executed.
At the successor of A and B (which is C), the undef variable cannot be reused anymore.

This patch implements this optimization by tracking the latest undef variable id per basic block, increment the id whenever a fresh undef var is created, and merge maximum value at the successors.
Naturally, it explains the sharing between undef variables in returning blocks too.

I'll leave the experimental result after running is done here.

### DIFF
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -364,7 +364,7 @@ smt::expr State::mkFreshUndef(const smt::expr &e_ty, bool do_register) {
   }
 
   if (latest_unused_undef_id == undef_var_pool.size()) {
-    string_view name = "undef_" + to_string(latest_unused_undef_id);
+    string name = "undef_" + to_string(latest_unused_undef_id);
     newv = expr::mkFreshVar(name.data(), e_ty);
     undef_var_pool.emplace_back(newv);
   } else {
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -230,9 +230,8 @@ const StateValue& State::operator[](const Value &val) {
   }
 
   vector<pair<expr, expr>> repls;
-  for (auto &u : uvars) {
-    repls.emplace_back(u, expr::mkFreshVar("undef", u));
-  }
+  for (auto &u : uvars)
+    repls.emplace_back(u, mkFreshUndef(u));
 
   if (hit_half_memory_limit())
     throw AliveException("Out of memory; skipping function.", false);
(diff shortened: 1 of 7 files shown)


nunoplopes:
buildbot is failing due to a warning (it's a bug in the code)

ir/state.cpp
```diff
@@ -343,12 +342,48 @@ bool State::isUndef(const expr &e, const Value *used_by) const {
   return undef_vars.count(e) != 0;
 }
 
+smt::expr State::mkFreshUndef(const smt::expr &e_ty, bool do_register) {
+  expr newv;
+
+  auto same_type = [](const expr &e1, const expr &e2) {
```
nunoplopes, line 348: you should implement a method in expr to check type. There it's much cheaper (just check if sort() are ==)

aqjune:
The result of LLVM unit test: timeout 4340 -> 4363, maybe this change isn't friendly to Z3.
However there is a good news: a few `SMT Error: smt tactic failed to show goal to be sat/unsat (incomplete quantifiers)` error disappears.
Running single file benchmarks raise a lot of this error, and I think this patch is a step forward resolving it. I'll run single file benchmarks and see the result.

aqjune:
Single file program results: [results-undefopt2.zip](https://github.com/AliveToolkit/alive2/files/5358683/results-undefopt2.zip)
gzip has a nice performance boost (+40 correct), oggenc has mild boost, sqlite3 has 10 `failed to show goal to be sat/unsat` errors disappeared. ph7 has a mild timeout increase, but the overall impact seems good.

aqjune:
The CI error seems to be timeout failure of unit tests

nunoplopes:
This patch is a bit complicated.. The reuse of undef vars is done blindly so the results are always going to be mixed.
The purpose of reusing undef vars is so that when one of them gets instantiated all of the reused vars get instantiated in the same way (i.e., with the same expression/value). If undefs get reused in the "same place", it's likely quantifier instantiations are the same. Otherwise it's a matter of luck, who knows. Seems a bit random to me.

In your example, it's not clear to me we want to encode %v1 and %v2 with the same variable. We want to share undef vars across common BBs when merging paths, but not necessarily across different BBs. e.g., if %v1 gets instantiated with 42, why should %v2 be instantiated with 42 as well? Unless the expressions are equal modulo undef, then I agree they can be merged. But that's a different patch, that's not what's implemented here.

aqjune:
What I've thought was something like this:
Currently undef rewrite may cause exponential growth of undef variables during symbolic execution, and this patch addresses the issue by sharing undef variables whenever possible.
I think the result that quite a few `SMT Error: smt tactic failed to show goal to be sat/unsat (incomplete quantifiers)` disappears from sqlite3 shows its benefit.

aqjune:
What do you think? Since exponential growth of undef variables can happen, it has to be addressed.
Updates in strip_undef is limited to undefs in arguments, so I think it might not be helpful for large programs.

aqjune:
I'll experiment the impact of this PR.
Currently the machine that I used is out of order after blackout :/ I'll run it after it is fixed.

aqjune:
I'll see whether this patch can be revived by running the experiment again

aqjune:
Hmm, # correct pairs of sqlite3 decreases quite a bit, which is not expected :(

PR #506, opened 2020-10-12
Add support for llvm.assume with noundef op bundle

### LINKED ISSUE #505: Support llvm.assume with operand bundle
Example:
```
call void @llvm.assume(i1 true) [ "noundef"(i1 %x) ]
%x.fr = freeze i32 %x
; %x.fr is %x
```

aqjune:
Resolves #505

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -711,6 +711,8 @@ end:
             BB->addInstr(
               make_unique<Assume>(*get_operand(v), Assume::WellDefined));
           }
+        } else {
+          return error(i);
         }
       }
       PARSE_UNOP();
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -692,7 +692,7 @@ end:
 
   RetTy visitUnreachableInst(llvm::UnreachableInst &i) {
     auto fals = get_operand(llvm::ConstantInt::getFalse(i.getContext()));
-    return make_unique<Assume>(*fals, Assume::IfNonPoison);
+    return make_unique<Assume>(*fals, Assume::AndNonPoison);
   }
 
   RetTy visitIntrinsicInst(llvm::IntrinsicInst &i) {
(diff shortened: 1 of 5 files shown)


llvm_util/llvm2alive.cpp
```diff
@@ -692,15 +692,29 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
+      for (unsigned j = 0; j < n; ++j) {
+        auto bundle = i.getOperandBundleAt(j);
+        llvm::StringRef name = bundle.getTagName();
+        if (name == "noundef") {
+          // Currently it is unclear whether noundef bundle takes single op.
+          // Be conservative & simply assume it can take arbitrary num of ops.
+          for (unsigned j = 0; j < bundle.Inputs.size(); ++j) {
+            llvm::Value *v = bundle.Inputs[j].get();
+            BB->addInstr(
+              make_unique<Assume>(*get_operand(v), Assume::WellDefined));
```
nunoplopes, line 712: what if get_operand returns null?

llvm_util/llvm2alive.cpp
```diff
@@ -692,15 +692,29 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
 
   RetTy visitIntrinsicInst(llvm::IntrinsicInst &i) {
     switch (i.getIntrinsicID()) {
     case llvm::Intrinsic::assume:
     {
+      unsigned n = i.getNumOperandBundles();
+      for (unsigned j = 0; j < n; ++j) {
+        auto bundle = i.getOperandBundleAt(j);
+        llvm::StringRef name = bundle.getTagName();
+        if (name == "noundef") {
```
nunoplopes, line 706: should we take this opportunity to skip unknown bundle tags to avoid false-positives?
We have a few unit tests that fail because of unhandled bundle tags.

llvm_util/llvm2alive.cpp
aqjune: I think it should be non-null, because def should dominate operand bundles: https://godbolt.org/z/xqE4Tb
It is assumed as non-null in other places, e.g., `visitUnreachableInst`.

llvm_util/llvm2alive.cpp
aqjune: Yep, I'll update it

PR #507, opened 2020-10-12
Check whether src and tgt have same param attributes

### LINKED ISSUE #476: Check refinement of function/input attributes
Right now we say the following is correct though it's not:
```
define i32 @src(* dereferenceable(4) %p) {
  ...
}
=>
define i32 @tgt(* dereferenceable(8) %p) {
  ...
}
```

We need to ensure that no new attributes are added to fn or arguments

aqjune:
This patch resolves #476 by conservatively checking whether src/tgt have same param attributes.

### DIFF
--- a/ir/attrs.cpp
+++ b/ir/attrs.cpp
@@ -56,10 +56,33 @@ bool ParamAttrs::undefImpliesUB() const {
   return ub;
 }
 
+bool ParamAttrs::operator==(const ParamAttrs &rhs) const {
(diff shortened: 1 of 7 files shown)


nunoplopes:
We should check this at Alive2 IR level, otherwise you need to repeat this code in every single front-end. For example, the opt plugin is not fixed with this patch.

aqjune:
Aha.. okay, will update this patch

PR #513, opened 2020-10-16
Let clang tv print result at one file per pass

aqjune:
This patch makes clang tv print result at one file per pass. When the directory is set using `-mllvm -tv-report-dir=<dir>`, log files are dumped into the directory. If there are multiple functions in IR, each log has only one function as well. This is to support finer parallelism in the future.

Additionally, a simple patch that makes llvm2alive use utils.cpp's `out` rather than maintaining its own copy is made (the first commit). It was needed in this patch.

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -25,7 +25,6 @@ using llvm::LLVMContext;
 
 namespace {
 
-ostream *out;
 unsigned constexpr_idx;
 unsigned copy_idx;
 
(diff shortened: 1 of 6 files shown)


aqjune:
CI is failing because `stdc++fs` isn't at the dependency of alive-tv, etc. Should I put relevant functions to `tv.cpp`, or keep it at `files.h` and add the dependency?

aqjune:
Activate the per-pass print mode only when `-tv-report-per-pass` is given (otherwise the output has no change)
Move makeUniqueFilePath back to tv.cpp

PR #521, opened 2020-10-20
Don't undef rewrite if unused in the ctrl flow

aqjune:
This patch adds `unused_vars` to value analysis, to support the optimization discussed yesterday

### DIFF
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -43,6 +43,7 @@ static T intersect_set(const T &a, const T &b) {
 void State::ValueAnalysis::intersect(const State::ValueAnalysis &other) {
   non_poison_vals = intersect_set(non_poison_vals, other.non_poison_vals);
   non_undef_vals = intersect_set(non_undef_vals, other.non_undef_vals);
+  unused_vars = intersect_set(unused_vars, other.unused_vars);
 
   for (auto &[fn, interval] : other.ranges_fn_calls) {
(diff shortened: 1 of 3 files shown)


aqjune:
Running LLVM unit tests

ir/state.cpp
```diff
@@ -149,6 +151,16 @@ expr State::strip_undef_and_add_ub(const Value &val, const expr &e) {
            is_undef_cond(not_undef, var);
   };
 
+  auto is_if_undef_or_add = [&](const expr &e, expr &var, expr &not_undef) {
```
nunoplopes, line 154: this is unrelated to the main patch. Can you split this and benchmark the those patches independently pls? Otherwise we have no idea what's causing what.

aqjune:
Split the strip_undef optimization as a separate PR

nunoplopes:
thanks! the patch LGTM. Please let us know the perf results once finished!

aqjune:
Timeout: 4158 -> 4157 ... :/ 
Maybe reestimate its impact with C benchmarks after things are prepared?

nunoplopes:
I'll run on the bechmark suite I've here. Takes a few hours.

nunoplopes:
I see a decent increase in number of UNSAT & trivial queries. This is good, it verifies more stuff, even if # of times out decreases less dramatically.

aqjune:
I'm glad to hear that; thank you so much!

PR #523, opened 2020-10-20
Add more patterns to strip_undef_and_add_ub

aqjune:
Splitted from #521

### DIFF
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -281,7 +281,7 @@ expr State::strip_undef_and_add_ub(const Value &val, const expr &e) {
       // (ite (bvule (bvadd c (ite (= #b0 isundef_%var) %var undef)) %val)
       //       #b1 #b0)
       if (is_if_undef_or_add(lhs, val, not_undef, newe) && !has_undef(rhs)) {
-        expr cond = rhs == expr::mkInt(-1, rhs)
+        expr cond = rhs == expr::mkInt(-1, rhs);
         addUB(not_undef || cond);
         if (cond.isFalse())
           mark_notundef(val);
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -259,12 +259,12 @@ expr State::strip_undef_and_add_ub(const Value &val, const expr &e) {
     }
 
     if (c.isULE(lhs, rhs)) {
-      uint64_t n;
+      int64_t n;
 
       // (ite (bvule val (ite (= #b0 isundef_%var) %var undef)) #b1 #b0)
       if (is_if_undef_or_add(rhs, val, not_undef, newe) && !has_undef(lhs)) {
         addUB(not_undef || lhs == 0);
-        if (lhs.isUInt(n) && n != 0)
+        if (lhs.isInt(n) && n != 0)
           mark_notundef(val);
         return expr::mkIf(lhs.ule(newe), a, b);
       }
@@ -272,7 +272,7 @@ expr State::strip_undef_and_add_ub(const Value &val, const expr &e) {
       // (ite (bvule (ite (= #b0 isundef_%var) %var undef) val) #b1 #b0)
       if (is_if_undef_or_add(lhs, val, not_undef, newe) && !has_undef(rhs)) {
         addUB(not_undef || rhs == expr::mkInt(-1, rhs));
-        if (rhs.isUInt(n) && n != -1)
+        if (rhs.isInt(n) && n != -1)
           mark_notundef(val);
         return expr::mkIf(newe.ule(rhs), a, b);
       }
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -151,6 +151,16 @@ expr State::strip_undef_and_add_ub(const Value &val, const expr &e) {
            is_undef_cond(not_undef, var);
   };
 
+  auto is_if_undef_or_add = [&](const expr &e, expr &var, expr &not_undef) {
+    // (ite (= #b0 isundef_%var) %var undef)
+    // (bvadd const (ite (= #b0 isundef_%var) %var undef))
+    expr a, b;
+    return is_if_undef(e, var, not_undef) ||
+           (e.isAdd(a, b) &&
+            ((b.isConst() && is_if_undef(a, var, not_undef)) ||
+             (a.isConst() && is_if_undef(b, var, not_undef))));
+  };
+
   expr c, a, b, lhs, rhs;
 
   // two variants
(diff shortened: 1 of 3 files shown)


nunoplopes:
Can you please give a proof in SMT2 that this is correct?
e.g. I've used this for one of the cases `(x <= y)`, with both x,y possibly undef input vars:
```smt2
(declare-const x (_ BitVec 4))
(declare-const y (_ BitVec 4))
(declare-const undef1 (_ BitVec 4))
(declare-const undef2 (_ BitVec 4))
(declare-const isundef_x (_ BitVec 1))
(declare-const isundef_y (_ BitVec 1))

(push)
(assert
(or
  (and (= #b0 isundef_x) (= #b0 isundef_y))
  (and (= #b0 isundef_x) (= x #x8))
  (and (= #b0 isundef_y) (= y #x7))
))

(assert
(not
(= (bvsle (ite (= #b0 isundef_x) x #x0)
          (ite (= #b0 isundef_y) y #x0))
   (bvsle (ite (= #b0 isundef_x) x undef1)
          (ite (= #b0 isundef_y) y undef2))))
)
(check-sat)
(get-model)
(pop)

(assert
(not
(or
  (and (= #b0 isundef_x) (= #b0 isundef_y))
  (and (= #b0 isundef_x) (= x #x8))
  (and (= #b0 isundef_y) (= y #x7))
)))

(assert
(forall ((undef1 (_ BitVec 4)) (undef2 (_ BitVec 4)))
(= (bvsle (ite (= #b0 isundef_x) x #x0)
          (ite (= #b0 isundef_y) y #x0))
   (bvsle (ite (= #b0 isundef_x) x undef1)
          (ite (= #b0 isundef_y) y undef2))))
)
(check-sat)
(get-model)
```

aqjune:
My head is too overloaded to try this now (and actually I became to think that my patch is slightly incorrect too :/ )
I'll write a proof in a fresh state tomorrow.

ir/state.cpp
```diff
@@ -187,76 +199,91 @@ expr State::strip_undef_and_add_ub(const Value &val, const expr &e) {
...
     if (c.isULE(lhs, rhs)) {
+      uint64_t n;
+
       // (ite (bvule val (ite (= #b0 isundef_%var) %var undef)) #b1 #b0)
-      if (is_if_undef_or_add(rhs, val, not_undef) && !has_undef(lhs)) {
+      if (is_if_undef_or_add(rhs, val, not_undef, newe) && !has_undef(lhs)) {
         addUB(not_undef || lhs == 0);
-        mark_notundef(val);
-        return expr::mkIf(lhs.ule(val), a, b);
+        if (lhs.isUInt(n) && n != 0)
```
aqjune, line 267: This condition was necessary.

aqjune:
Running LLVM unit tests...
Actually, I couldn't understand the meaning of the proof yet..
Could you explain the high level structure of the proof please?

tests/alive-tv/undef/strip-add-eq.srctgt.ll
```diff
@@ -0,0 +1,26 @@
...
+  br i1 %c, label %A, label %B
+A:
+  %w = add i32 %x, 2
+  ret i32 %w
+B:
+  %w2 = add i32 %y, 2
+  ret i32 %w2
+}
+
+; CHECK: (ite (= %x %y) (bvadd #x00000002 %x) (bvadd #x00000002 %y)) / true
```
nunoplopes, line 26: this is brittle, as the order of elements can change at any time. It's controlled by Z3.
Isn't there a better way of checking that the optimization is working? maybe not, but this way isn't very good.

aqjune: Maybe I can simplify this example to return `x + 2` only - the order wouldn't matter then.

aqjune: Since it could be `2 + x` too, made it simply `x`

nunoplopes:
Ok, the proof is trying to prove that the original constraint to check if an expression is not undef is equivalent to the new expression.
The old expression U is e(x,y,0) = e(x,y,undef). The new expression U' is the or condition above.
Then the proof is split in two, one for each implication:
 1) if the new expr says it's not undef, the old expr should also say it's not undef.  For SMT it's smth like: `∃x, y, u . U'(x,y) ∧ ¬U(x,y,u)` is unsat

2) if the new expr says it's undef, the old expr should also say it's undef. In SMT: `∃x, y . ∀u . ¬U'(x,y) ∧ U(x,y,u)` is unsat (there's no `u` that makes the value of the expression different from when we replace undef with 0, i.e., the expression has a single value for all values of `u`).

aqjune:
Thanks..!
Hope this proof makes sense: [proof.zip](https://github.com/AliveToolkit/alive2/files/5417117/proof.zip)

LLVM unit test is still running; it takes a bit of time because I had to run master as well. Likely to be done in an hour

nunoplopes:
> Thanks..!
> Hope this proof makes sense: [proof.zip](https://github.com/AliveToolkit/alive2/files/5417117/proof.zip)
> 
> LLVM unit test is still running; it takes a bit of time because I had to run master as well. Likely to be done in ~an hour~ several hours, maybe tomorrow morning here :)

the proof makes sense, but it's missing many cases. You changed all of the previous code, so we need to prove every single case again.
BTW, from which benchmark does this pattern come from?

aqjune:
The previous cases are still there. The diffs are results of using `is_if_undef_or_add` instead of `is_if_undef`.
LLVM unit tests timeout: 4156 -> 4152
I remember from the last Skype talk that you mentioned this pattern happened from sqlite3, but I'll check from my side.

aqjune:
Eh... I just started sqlite3 because machines have been full, I'll leave the stat here when it's done.

aqjune:
sqlite3 consists of 455 .c files, and running for 11 hours validated 103 files.
With this patch, +1,031 cases are triggered. For comparison, 25,748 cases are triggered when `is_if_undef` was only used.

nunoplopes:
@aqjune can we merge this? do you have the smt2 proofs for this?

aqjune:
Yes, this is an updated proof containing all combinations of these cases:
```
// op = eq, sle, ule
x op y
x op (y + c)
(x + c) op y
(x + c) op (y + c2)
```
[proof2.zip](https://github.com/AliveToolkit/alive2/files/5470345/proof2.zip)
No regression found from LLVM unit tests

nunoplopes:
> Yes, this is an updated proof containing all combinations of these cases:
> 
> ```
> // op = eq, sle, ule
> x op y
> x op (y + c)
> (x + c) op y
> (x + c) op (y + c2)
> ```
> 
> [proof2.zip](https://github.com/AliveToolkit/alive2/files/5470345/proof2.zip)
> No regression found from LLVM unit tests

The proofs look great, thank you!

Issue #522, opened 2020-10-20
Encoding I/O builtins as unknown fn calls in InstCombine pass


aqjune:
Since I/O builtin calls like `printf` are optimized at InstCombinePass only, it is practically okay to regard them as unknown fn calls in non-InstCombine passes.

However, this requires maintaining two versions of Alive2 Functions - the one encoded with `io_nobuiltin=true` and another one encoded with `io_nobuiltin=false` at each pass.
Otherwise, it is impossible to recover the other version when needed.

For example: assumpe JumpThreading, InstCombine, SimplifyCFG are applied consecutively (F, F', F'', F''' are Alive2 functions) 

F -- JumpThreading -- F' -- InstCombine -- F'' -- SimplifyCFG -- F'''

F, F''' should be built with `io_nobuiltin=true`, but what about F' and F''?

Two possible solutions:

(1) Get the list of passes in advance somehow and use it. But I can't find the way to get it. :/ Do we need a separate invocation of clang process with no TV enabled?
(2) By default, get functions with `io_nobuiltin=true` and maintain it. When the pass is InstCombine, get functions with both `io_nobuiltin=true` and `false`, use the latter one to verify, store the former one for the upcoming passes. Question is its soundness; I think it is okay.

nunoplopes:
Why not store the flag in FnCall itself?
Skip verification if tgt fncall has known fn = true. But do verify otherwise.
This only fails if instcombine may remove a fncall from src. which I guess it probably does remove `printf("")`.
So we may need a global flag in Function?

aqjune:
Aha.. understood, makes sense. Then we don't need to generate Alive2 Function twice.

Will make a patch for this; after this, I believe I can give the first shot for various C benchmarks that we discussed.

The list of programs:
glib, PHP, cpython, sqlite3

Maybe we can run C benchmarks from SPECCPU 2017 too:
500.perlbench_r, 502.gcc_r, 505.mcf_r, 525.x264_r, 557.xz_r, 519.lbm_r, 538.imagick_r, 544.nab_r

Actually I want to start with small benchmarks in SPECCPU 2017 today, because it is straightforward to compile (no dependency issue, custom setting, etc)

What Crellvm had: SPECCPU 2006 C benchmarks + sendmail, emacs, gimp, ghostscript

aqjune:
Irrelevant with this specific issue, but related with clang plugin:

[...]

PR #524, opened 2020-10-20
clang plugin: encode I/O calls as unknown fncalls except when running…

aqjune:
… InstCombine (issue #522 ).

1. With this patch, known_call() now returns `FnKnown`, `FnUnknown`, or `FnDependsOnOpt`. The last option means some optimizations know this fn, some opts don't.
2. If it is `FnDependsOnOpt`, llvm2alive creates `FnCall` instance with valid = `DependsOnFlag`.
3. FnCall::toSMT returns {} if valid = `DependsOnFlag` and Alive2 Function says it should be invalid (as discussed in #522).

### DIFF
--- a/ir/function.h
+++ b/ir/function.h
@@ -62,6 +62,8 @@ class Function final {
   unsigned bits_pointers = 64;
   unsigned bits_ptr_offset = 64;
   bool little_endian = true;
+  // If this is true, FnCalls having DependsOnFlag become valid unknown fn calls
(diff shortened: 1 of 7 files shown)


aqjune:
Okay, mcf_r of SPECCPU 2017 is now running, and I see many unsupported features due to function pointer calls. I'll share the report when it's done.
I *think* supporting function pointer calls shouldn't be very hard, considering our function call encoding? It is really natural to expand.

PR #531, opened 2020-10-22
Add fabs

aqjune:
This patch adds fabs.
LLVM unit tests running...


ir/instr.cpp
```diff
@@ -872,6 +873,11 @@ StateValue UnaryOp::toSMT(State &s) const {
       return v.cttz(expr::mkInt(-1, v)) + expr::mkUInt(1, v);
     };
     break;
+  case FAbs:
```
LebedevRI, line 876: Looking at FNeg handling above, doesn't this need fast-math flags?

aqjune: You're right, thanks..! Forgot that call instructions can take fast-math flags.

llvm_util/llvm2alive.cpp
```diff
@@ -64,28 +64,6 @@ string_view s(llvm::StringRef str) {
   } while (0)
 
 
-FastMathFlags parse_fmath(llvm::Instruction &i) {
```
aqjune, line 67: Moved to `utils.h/cpp` to make it visible to `known_fns.cpp`
I chose `utils.h/cpp` because utils seems to be the place where auxiliary functions are gathered

aqjune:
Unit tests okay except the failure that I shared

ir/instr.cpp
```diff
@@ -872,6 +873,14 @@ StateValue UnaryOp::toSMT(State &s) const {
...
+    // TODO
```
nunoplopes, line 877: please fill this in. We don't need more TODOs in the code.

aqjune:
fast math flag added

PR #532, opened 2020-10-22
Add returned attribute

aqjune:
This patch adds support for `returned` attribute.
My machine for LLVM unit test is preoccupied with the updated fabs patch, I will leave report here after testing fabs and this patch is done

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2270,6 +2270,16 @@ StateValue Return::toSMT(State &s) const {
   if (attrs.has(FnAttrs::NoReturn))
     s.addUB(expr(false));
 
+  if (auto *val_returned = s.getFn().getReturnedInput()) {
+    // LangRef states that return type must be valid operands for bitcasts,
+    // which cannot be aggregate type.
+    assert(!dynamic_cast<AggregateType *>(&val->getType()));
+    auto &[v_returned, np_returned] = s[*val_returned];
+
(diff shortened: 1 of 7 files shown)


aqjune:

nunoplopes:
We need to support this in fncalls as well, like:
```
%y = f(returned %x)
=>
f(returned %x)
%y = %x
```

aqjune:
It seems `returned` has already been supported in the case: `tests/alive-tv/attrs/returned.src/tgt.ll` has such inputs.

ir/instr.cpp
```diff
@@ -2270,6 +2270,16 @@ StateValue Return::toSMT(State &s) const {
   if (attrs.has(FnAttrs::NoReturn))
     s.addUB(expr(false));
 
+  if (auto *val_returned = s.getFn().getReturnedInput()) {
+    // LangRef states that return type must be valid operands for bitcasts,
+    // which cannot be aggregate type.
+    assert(!dynamic_cast<AggregateType *>(&val->getType()));
+    auto &[v_returned, np_returned] = s[*val_returned];
```
nunoplopes, line 2277: uhm, this evaluates the inputs multiple times. Can we improve this?

nunoplopes: or maybe it doesn't matter; multiple returns aren't common.

PR #533, opened 2020-10-23
Add -tv-elapsed-time

aqjune:
This PR adds `-tv-elapsed-time` that prints the running time of TV.

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -150,7 +150,7 @@ llvm::cl::opt<bool> opt_io_nobuiltin(
 llvm::cl::opt<bool> opt_elapsed_time(
     "tv-elapsed-time",
     llvm::cl::desc("Print the elapsed time"),
-    llvm::cl::init(false));
+    llvm::cl::cat(TVOptions), llvm::cl::init(false));
 
 struct FnInfo {
   Function fn;
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -156,7 +156,6 @@ struct FnInfo {
   Function fn;
   unsigned order;
   std::string fn_tostr;
-  float elapsed_time = 0;
 };
 
 ostream *out;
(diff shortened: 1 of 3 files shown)


aqjune:
Exclude the time taken by LLVM optimizations

nunoplopes:
can you use util/stopwatch.h instead pls?

aqjune:
Hi, can I get this merged? This is useful for estimating the time spent on translation validation only.

nunoplopes:
I'm failing to see the point of this patch. It's measuring the aggregated time of TV across all functions, right?
So it's: process run time - optimization time. Since optimization time is so small when compared to TV time, can't you just use the process run time?
Unless you want to do per function pair, which would make more sense to me. We have a time around vcgen, but not around refinement formula creation nor SMT solving. We miss runaways in those two.

aqjune:
I agree that per function time info should be more useful; updated the patch as suggested
.

tv/tv.cpp
```diff
@@ -142,6 +148,7 @@ optional<smt::smt_initializer> smt_init;
 optional<llvm_util::initializer> llvm_util_init;
 TransformPrintOpts print_opts;
 unordered_map<string, FnInfo> fns;
+unordered_map<string, float> fns_elapsed_time;
```
aqjune, line 151: Declared a separate map here because if a function `f` has unsupported instructions `fns` may not have a slot for `f`.

aqjune:
The output looks like this:
```
----------------- ELAPSED TIME ------------------
  _Z2anIP1pE1iT_: 0.0 s
  _ZN1hIPFivEEC1EiS1_: 0.1 s
  _ZN1lC2IZ1avE3$_0EET_: 0.0 s
  _Z2auIP1pEv1oT_: 0.3 s
  _ZN1hIPFivEEC2EiS1_: 1.3 s 
  _ZN1FIP1pE2apES1_: 0.0 s
  _ZN1lC1IZ1avE3$_0EET_: 0.0 s
  _Z1av: 0.1 s
  <TOTAL>: 1.8 s
```

PR #534, opened 2020-10-23
experimental is not needed when compiler is clang

aqjune:
When clang is used to compile tv.cpp, `__GNUC__` is set to 4, causing `experimental/filesystem` to be used.
This invokes `undefined symbol: _ZNSt12experimental10filesystem2v17__cxx114path14_M_split_cmptsEv` error when clang tv is used in Linux with debug info on.

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -25,7 +25,7 @@
 #include <unordered_map>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


nunoplopes:
I just removed support for gcc 8.

PR #541, opened 2020-10-26
clang plugin: add -tv-batch to verify optimizations in a batch

aqjune:
This is a patch that adds `-tv-batch` for clang plugin that batches supported optimizations and verify them at once.

Currently WIP; running benchmarks to see whether it works successfully

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -84,6 +84,12 @@ llvm::cl::opt<bool> opt_smt_skip("tv-smt-skip",
                                  llvm::cl::cat(TVOptions),
                                  llvm::cl::init(false));
 
+llvm::cl::opt<bool>
+  opt_batch("tv-batch",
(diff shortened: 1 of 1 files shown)


aqjune:
Closing this, since this won't be used for the experiment.

nunoplopes:
But is the patch close to being finished?
This is still a nice feature to have. We can finish it later. Just wondering if the patch is worth it keeping for later or if the approach doesn't work at all?

aqjune:
Um, I guess this patch might be still slightly far from being prepared due to the function duplication issue - #522 .
Function duplication now isn't an urgent issue because bitcodes are pre-generated and alive-tv process is created per function pairs. So maybe it can be done after the submission..? :) Then, this can be reopened.

aqjune:
I'll revive this patch after the deadline.

PR #544, opened 2020-10-27
Fix issue #542

aqjune:


regehr:
I can confirm that this fixes the crashes I saw while running our clang plugin on files from llvm/lib/Analysis

PR #547, opened 2020-10-27
TV: group all options into a grop, to make --help more readable

LebedevRI:
It's already done for all other options, just not these.


nunoplopes:
The indentation doesn't seem very inform to me, but..
Thanks for the patch!

LebedevRI:
> The indentation doesn't seem very inform to me, but..

Wasn't me, that's what clang-format gave me..

> Thanks for the patch!

Thank you!

PR #550, opened 2020-10-30
buglist.md: add undef category

aqjune:
Adds undef category to distinguish bugzilla reports that will disappear when undef is removed from LLVM.


nunoplopes:
I don't see the point of this really. It may induce people in error, thinking of reduced value of those bugs.

aqjune:
I was slightly embarassed; I thought the undef category was natural because the dashboard's graph was showing the result with/without undef as well.
Anyway, I'll maintain it locally then.

Issue #477, opened 2020-09-01
Semantics of readnone function calls is wrong


fhahn:
It seems like function calls to `readnone` functions lead Alive2 to verify invalid transformations as correct.

The example below verifies as correct if `%f.1` is replaced by `true`, which is *not* correct, as it is always `false`.

If the first call to `@use` is removed, Alive2 correctly rejects the transformation as invalid. The use of functions to insert uses to check is quite common in LLVM tests and this could lead to invalid transformations to not being flagged up.

Full example: https://alive2.llvm.org/ce/z/9agPfJ
```
declare void @use(i1) readnone

define void @src(i32 %x, i32 %y) {
entry:
  %c.1 = icmp ule i32 %x, %y
  br i1 %c.1, label %bb1, label %bb2

bb1:
  ret void

bb2:
  call void @use(i1 true)
  %f.1 = icmp ule i32 %x, %y
  call void @use(i1 %f.1)
  ret void
}
```

nunoplopes:
Ok, I see what's the issue: src calls already `use(true)` so in terms of UB that gets covered (assuming that readnone functions trigger UB for the same set of inputs).
src also calls `use(false)` but the target doesn't. This is ok because if anything we are removing UB. Calling `use(true)` instead doesn't add any more UB. See this example where Alive2 also says it's correct (introduction of fn calls to readnone): https://alive2.llvm.org/ce/z/r-F8_i
Removing fn calls also works: https://alive2.llvm.org/ce/z/T2MqPy

We may have an issue with the understanding of the semantics (there's even a TODO in the code 😀). To summarize:
 - Can readnone calls be removed? If not, why?
 - Can readnone calls be inserted? If not, why?
 - (extra question): do readnone functions always return? (if not we can't introduce them)

My understanding was that readnone functions are pure (only depend on the inputs) and don't modify any state. Hence you could remove & add them. If they can modify some state, then we need to understand what's the exact semantics.
Any suggestions/opinions/examples on the above?  I'm of course happy to implement whatever the right thing is 😀

fhahn:
Ah, I think I see what is going on with my original test case. Alive is basically ignoring/removing the second `@use` call because it does not modify any state and the result is not used?

One thing to note is that `readnone` is limited to memory visible to the program. In some places in the LangRef also just mention visible to to caller and explicitly call out that there could be an unwind implementation that does not use memory accessible by the caller.

Intersting questions, below my interpretations:

[...]

PR #551, opened 2020-10-31
make fn call range empty if function never reads memory

aqjune:
This is my solution of https://github.com/AliveToolkit/alive2/issues/477#issuecomment-719904013 .

IIUC, `analysis.ranges_fn_calls` states how many other functions have been called before the current function call is reached.
This is used to prune out pairwise equivalence encoding between function calls by ignoring calls from different input memory state.

However, if a function never reads memory, outputs are equivalent regardless of the input memory.
`analysis.ranges_fn_calls` should not prune out the equivalence encoding of such no-read calls in this case.

### DIFF
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -693,8 +693,8 @@ State::addFnCall(const string &name, vector<StateValue> &&inputs,
     auto call_data_pair
       = calls_fn.try_emplace(
           { move(inputs), move(ptr_inputs),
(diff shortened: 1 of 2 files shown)


aqjune:
The example is reduced from gzip.

nunoplopes:
This is not a real fix. I can take this as an optimization, but the memory refinement needs fixing.

PR #552, opened 2020-11-02
step 2,  fix phi nodes inside loops

zhengyangl:
@nunoplopes Hi Nuno, I have one question, how do we deal with multiple loops share one common header? Should we canonicalize the loop  before unrolling?

### DIFF
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -58,6 +58,45 @@ void BasicBlock::replaceTargetWith(const BasicBlock *from,
   }
 }
 
+void BasicBlock::replacePhiSrcWith(const string &from,
+                                   const string &to) {
+  for (auto &i : instrs()) {
+    if (Phi *phi = dynamic_cast<Phi*>(const_cast<Instr*>(&i)))
+      phi->replaceSourceWith(from, to);
+  }
+}
+
+void BasicBlock::removePhiSrc(const BasicBlock *bb) {
+  for (auto &i : instrs()) {
+    if (Phi *phi = dynamic_cast<Phi*>(const_cast<Instr*>(&i))) {
+      phi->removeValue(bb->getName());
+    }
+  }
+}
+
+void BasicBlock::removePhiSrcsExcept(const BasicBlock *bb) {
+  for (auto &i : instrs()) {
+    if (Phi *phi = dynamic_cast<Phi*>(const_cast<Instr*>(&i))) {
+      for (auto &src : phi->sources())
+        if (bb->getName() != src)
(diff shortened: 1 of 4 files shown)


nunoplopes:
Unit tests please? What tests does this fix?

LebedevRI:
Relevant, but bikeshedding/derailing comment: i'd love to be able to prove things like this: https://godbolt.org/z/dKxzKx

nunoplopes:
> Relevant, but bikeshedding/derailing comment: i'd love to be able to prove things like this: https://godbolt.org/z/dKxzKx

That's a good test. It should work with this patch. @zhengyangl can you test it as well pls?

regehr:
@zhengyangl Nuno is correct, you have to provide tests to go along with patches. Just provide us with a subset of the tests that you used to validate the patch in the first place. You can also add XFAIL tests for things that don't work yet. The unit test infrastructure that we have makes this all fairly easy.

nunoplopes:
The algorithm seems incorrect to me. It only updates phi's BBs predecessors, but not the value. It needs to update both.

nunoplopes:
> The algorithm seems incorrect to me. It only updates phi's BBs predecessors, but not the value. It needs to update both.

ok, I take that back. Values are updated with rauw before.
But then I don't see why the algorithm is so complicated. A linear pass over the BBs in order should do the trick? The information for the predecessor is kinda there already through the value.
Anyway, once we have your unit tests I can test my idea.

zhengyangl:
Yeah, will send a new PR for test cases in 1 hour.

nunoplopes:
@LebedevRI your test case is now supported; see here: https://github.com/AliveToolkit/alive2/blob/master/tests/alive-tv/loops/ctlz.srctgt.ll
The unroll factor = max number of iterations the loop may run for.

All single-exit loops are now supported (modulo bugs). If something doesn't work, please let us know 😀

LebedevRI:
@zhengyangl @nunoplopes thank you!

regehr:
This is super cool! However, I've bumped alive2 at our compiler explorer and I don't understand what I'm seeing. src is unrolled properly, but tgt doesn't seem to be. Am I missing something?

https://alive2.llvm.org/ce/z/oHkpVe

regehr:
ack nevermind, I misread. I thought the optimization was reducing the number of iterations but it's just eliminating the loop entirely.

LebedevRI:
> @zhengyangl @nunoplopes thank you!

Moar loop idioms!!!!
https://reviews.llvm.org/D91038

PR #553, opened 2020-11-02
test cases for inner phi rewiring

zhengyangl:
@nunoplopes , in this pull request, all of the positive test cases can not pass identity check for src (-src-fn=src -tgt-fn=src), and all of them crashes with "ERROR: Precondition is always false". So I marked them SKIP-IDENTITY temporarily. I'll keep looking until they are fixed.


nunoplopes:
> @nunoplopes , in this pull request, all of the positive test cases can not pass identity check for src (-src-fn=src -tgt-fn=src), and all of them crashes with "ERROR: Precondition is always false". So I marked them SKIP-IDENTITY temporarily. I'll keep looking until they are fixed.

Ok, can you investigate what's the issue pls?

zhengyangl:
Yeah I am looking now.

zhengyangl:
@nunoplopes , Hi Nuno, I noticed that when RHS get unrolled, it generates #sink. This line is setting pre_tgt to false, leading to `ERROR: Precondition is always false` https://github.com/AliveToolkit/alive2/blob/4db2e1556ce4b824a4eca7d11e3b91f527ccbd28/tools/transform.cpp#L374

Also it is not only happening in the four test cases in this PR, rather it happens in all the existing cases when RHS is getting unrolled. Is the tgt_state.sinkDomain() call correct?

aqjune:
Hi Zhengyang,
If the reason of `precondition is false` is due to the sink domain, it means all possible control flows are falling to #sink, which might not be the case that we want.

I think it is more related with just a python script that makes src and tgt run twice. It doesn't pass the unroll arguments: https://github.com/AliveToolkit/alive2/blob/4db2e1556ce4b824a4eca7d11e3b91f527ccbd28/tests/lit/lit/formats/alive2test.py#L97
Let me write a patch for this.

aqjune:
No, it wasn't related with `alive2test.py`.

I think the reason is that simply the loop isn't unrolled enough.
Running `alive-exec` on the examples raises the following error:
```
ERROR: Function doesn't reach a return statement
```

And the `precondition is false` is a valid error in this case.

zhengyangl:
@aqjune Thanks for the explanation! I've fixed the test cases so that they'll get enough unrolled.

PR #557, opened 2020-11-05
Add support for align attribute

aqjune:
This PR adds support for align attribute.

It is analogous to nonnull; violating the alignment condition is UB.


nunoplopes:
We really need to refactor this attribute stuff.. The amount of code we need to merge call site/func decl attributes is crazy.

aqjune:
+1
To make it beautiful updates in LLVM side is required as well; we need more functions in Function and CallBase.
Good thing is that introduction of those will help refactor LLVM as well.
I'll leave this as an issue.

PR #559, opened 2020-11-05
Recompute dom tree after each unroll

zhengyangl:
- Recompute dom tree after each unroll
- Jump over empty bb (#sink) in CFG::edge_iterator::next()


nunoplopes:
ok, let's see if recomputing becomes a bottleneck or not. We could map new blocks to old ones instead, but not worth the pain unless it becomes slow.

PR #563, opened 2020-11-06
bits_program_pointer should be initialized in advance

aqjune:
This PR resolves Transforms/GVN/pr25440.ll.

`Function::unroll()` can create an alloca whose size is `Memory::getStoreByteSize`, but pointer's store size is returning zero because `bits_program_pointer` isn't initialized yet.


aqjune:
The failure at Transforms/NewGVN/pr31594.ll is also due to this

PR #565, opened 2020-11-06
a possible solution for pr32838.ll crash

aqjune:
Transforms/NewGVN/pr32838.ll crashes when building a dominator tree, and the reason is really interesting.

IIUC, in DomTree, a parent DomTreeNode's order should be larger than the child's order (note that the entry's order is the maximum).

But, idom finding algorithm always starts with the first element of a predecessor block, and the predecessor block *can* have a ~~larger~~ smaller order than the current block.

This caused a crash in `DomTree::intersect`'s while loop.

But interesting thing is that, the paper (https://www.cs.rice.edu/~keith/EMBED/dom.pdf ) was using the predecessor-starting algorithm.
Would it be a bug in the paper?

### DIFF
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -863,8 +863,8 @@ void DomTree::buildDominators(const CFG &cfg) {
       auto &b_node = doms.at(b);
       if (b_node.preds.empty())
         continue;
-      
-      auto new_idom = b_node.preds.front();
+
+      DomTreeNode *new_idom = nullptr;
       for (auto p : b_node.preds) {
         if (p->dominator != nullptr) {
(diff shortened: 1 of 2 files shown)


aqjune:
Transforms/GVN/pr25440.ll is related with this - in the master branch, it causes an infinite loop.
With this patch, domtree is correctly built.
a
cfg:
foo

nunoplopes:
The algorithm is correct; this is not the right fix.
The issue is that the function is being called with a non-top-sorted graph. I've an easy fix, but I'm contemplating if there's something better.

aqjune:
Sorry for my confusion, I must have checked whether it was topologically sorted first :(

PR #570, opened 2020-11-07
if phi uses another phi in the block, it is previous iteration's one

aqjune:
Consider this example:
```
loop:
  %i = phi i32 [0, %entry], [%i.next, %loop]
  %j = phi i32 [0, %entry], [%i, %loop] 
  %i.next = add i32 %i, 1
  %c = icmp ne i32 %i, 1
  br i1 %c, label %loop, label %exit
exit:
  ret i32 %j
```

In the first iteration, %i and %j are both 0.
In the second iteration, %i is 1, but %j is still 0, because it is the value of %i of the previous iteration.
See here: https://godbolt.org/z/dTWePq
The returning value should be 0.

This patch fixes `cloneBB` to correctly handle this case.

### DIFF
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -390,17 +390,16 @@ cloneBB(Function &F, const BasicBlock &BB, const char *suffix,
         unordered_map<const Value*, vector<pair<BasicBlock*, Value*>>> &vmap) {
   string bb_name = BB.getName() + suffix;
   auto &newbb = F.getBB(bb_name);
-  set<const Value *> phis_from_org_bb; // phi nodes from the original BB
+  unordered_set<const Value *> phis_from_orig_bb;
 
   for (auto &i : BB.instrs()) {
     if (dynamic_cast<const Phi *>(&i))
-      phis_from_org_bb.insert(&i);
+      phis_from_orig_bb.insert(&i);
 
     auto d = i.dup(suffix);
     for (auto &op : d->operands()) {
       auto it = vmap.find(op);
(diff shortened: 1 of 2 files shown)


aqjune:
Running LLVM unit tests to see the impact of this

aqjune:
No regression, resolves Transforms/IndVarSimplify/no-iv-rewrite.ll and Transforms/InstCombine/gep-combine-loop-invariant.ll

nunoplopes:
This patch is wrong for unroll > 2:
```
%loop:
  %i = phi i32 [ 0, %entry ], [ %i.next, %loop ]
  %j = phi i32 [ 0, %entry ], [ %i, %loop ]
  %i.next = add i32 %i, 1
  %c = icmp ne i32 %i, 1
  br i1 %c, label %loop#2, label %exit

%loop#2:
  %i#2 = phi i32 [ %i.next, %loop ]
  %j#2 = phi i32 [ %i, %loop ]
  %i.next#2 = add i32 %i#2, 1
  %c#2 = icmp ne i32 %i#2, 1
  br i1 %c#2, label %loop#3, label %exit

%loop#3:
  %i#3 = phi i32 [ %i.next#2, %loop#2 ]
  %j#3 = phi i32 [ %i, %loop#2 ]   <------- should be %i#2
  %i.next#3 = add i32 %i#3, 1
  %c#3 = icmp ne i32 %i#3, 1
  br i1 %c#3, label #sink, label %exit
```

PR #572, opened 2020-11-07
a possible solution for illegal-addr-modes.ll: refresh users after cloning bbs

aqjune:
Transforms/LoopStrengthReduce/illegal-addr-modes.ll fails because unroll fails to replace cloned instructions' operands.

`getUsers()` is being called before `cloneBB()`, so newly created instructions aren't there.

`illegal-addr-modes.srctgt.ll` is a reduced example having small diff only.
After unrolling, `%p.0` in tgt's `%p.1 = getelementptr i8, i8* %p.0, i32 %i` should be replaced with a newly inserted phi.
But, after unrolled twice, `%p.1` in the second iteration still uses `%p.0`.

I hope this is the correct direction :)

### DIFF
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -583,7 +583,8 @@ void Function::unroll(unsigned k) {
 
       for (auto &[user, user_bb] : I->second) {
         // users inside the loop have been patched already
-        if (bbmap.count(user_bb))
+        if (find(unrolled_bbs.begin(), unrolled_bbs.end(), user_bb) !=
+            unrolled_bbs.end())
           continue;
 
         // insert a new phi on each dominator exit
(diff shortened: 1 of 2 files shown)


aqjune:

nunoplopes:
This patch introduces an extra `%p.0.next#phi#0` which is unused.

nunoplopes:
First part of the fix: https://github.com/AliveToolkit/alive2/commit/d9ff0d5cb4b43059c48cbb17e543f83cda5b4c60
cloneBB needs fixing now; will review #570 now to see if that's sufficient to fix this one as well.

nunoplopes:
This patch is not the right direction; cloneBB needs fixing instead. Working on it.

PR #575, opened 2020-11-09
Add tests from issue571

aqjune:


tests/alive-tv/loops/two-backedges-fail.srctgt.ll
```diff
@@ -0,0 +1,29 @@
+; TEST-ARGS: -src-unroll=2
+; SKIP-IDENTITY
```
nunoplopes, line 2: if you add -tgt-unroll=2 you don't need to skip identity.  It's always good to test identity.

PR #577, opened 2020-11-11
add more bug reproducers

aqjune:
This PR contains 18 more src/tgt pairs harvested from bugzilla.

Finding a bug from `pr1014.srctgt.ll` required precise encoding of escaped local blocks, so I wrote `pr1014-nolocal.srctgt.ll` and checked that a bug was found.
`pr22727.srctgt.ll` shows that it was actually not a bug.


nunoplopes:
thank you!

Issue #569, opened 2020-11-06
unrolling-related crash


regehr:
reduced from a test case due to YARPGen

```
regehr@home:~/tmp/Dpcei1NgDC/reduce$ cat foo.cpp
void a() {
  for (;;)
    for (int b; b;)
      ;
}
regehr@home:~/tmp/Dpcei1NgDC/reduce$ ~/alive2/build/scripts/alive++ -O3 foo.cpp -w -c -mllvm -tv-tgt-unroll=2 -mllvm -tv-src-unroll=2
-- 1. (anonymous namespace)::TVInitPass : Skipping
-- 2. EntryExitInstrumenterPass : Skipping
-- 3. ModuleToFunctionPassAdaptor<llvm::EntryExitInstrumenterPass>
-- 4. InferFunctionAttrsPass : Skipping
-- 5. SimplifyCFGPass

----------------------------------------
define void @_Z1av() {
%entry:
  %b = alloca i64 4, align 4, dead
  br label %for.cond

%for.cond:
  %0 = bitcast * %b to *
  start_lifetime * %0
  br label %for.cond1

%for.cond1:
  %1 = load i32, * %b, align 4
  %tobool = icmp ne i32 %1, 0
  br i1 %tobool, label %for.body, label %for.cond.cleanup

%for.body:
  br label %for.cond1#1#2

%for.cond1#1#2:
  %1#1#2 = load i32, * %b, align 4
  %tobool#1#2 = icmp ne i32 %1#1#2, 0
  br i1 %tobool#1#2, label %for.body#1#2, label %for.cond.cleanup

%for.body#1#2:
  br label %for.cond1#exit

%for.cond1#exit:
  %1#exit = load i32, * %b, align 4
  %tobool#exit = icmp ne i32 %1#exit, 0
  br i1 %tobool#exit, label #sink, label %for.cond.cleanup

%for.cond.cleanup:
  %2 = bitcast * %b to *
  free * %2 unconstrained
  br label %for.end

%for.end:
  br label %for.cond#2

%for.cond#2:
  %0#2 = bitcast * %b to *
  start_lifetime * %0#2
  br label %for.cond1#2

%for.cond1#2:
  %1#2 = load i32, * %b, align 4
  %tobool#2 = icmp ne i32 %1#2, 0
  br i1 %tobool#2, label %for.body#2, label %for.cond.cleanup#2

%for.body#2:
  br label %for.cond1#1#2#2

%for.cond1#1#2#2:
[...]

PR #578, opened 2020-11-11
tv: recreate tgt function since preprocess may have touched it

aqjune:
This addresses #522, and resolves the crash described at https://github.com/AliveToolkit/alive2/issues/569 .

### DIFF
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -457,14 +457,14 @@ static auto getPhiPredecessors(const Function &F) {
   return map;
 }
 
-void Function::unroll(unsigned k) {
+bool Function::unroll(unsigned k) {
   if (k == 0)
-    return;
+    return false;
 
   LoopAnalysis la(*this);
   auto &roots = la.getRoots();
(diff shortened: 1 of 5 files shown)


nunoplopes:
did you measure the impact of this?
Does tracking whether preprocess changed anything help? or is this just a mega change for nothing?

aqjune:
I thought it might be helpful for a scenario that runs llvm unit test with smt-skip and unroll=0 when want to list unsupported functions, but the experiment says it only has about 1% burden.
Didn't mean to just dump a change, sorry if it looked like that.

nunoplopes:
> I thought it might be helpful for a scenario that runs llvm unit test with smt-skip and unroll=0 when want to list unsupported functions, but the experiment says it only has about 1% burden.
> Didn't mean to just dump a change, sorry if it looked like that.

It's fine; I was just wondering about the numbers because I see the preprocessor kick in a lot in practice (with real programs, less with unit tests).

PR #581, opened 2020-11-12
test: add inf-loop-unrolled/local-unescaped version of bugzilla reports

aqjune:
This PR adds infinite-loop-unrolled/local-block-unescaped version of four bugs: pr10067, pr23599, pr29034, pr36437

This shows that modulo inf loop / escaped local block issues, bugs can be detected.

Additionally, pr24179.srctgt.ll was working! infinite loop support wasn’t necessary to find a bug from this.

And a few miscellaneous changes:
- pr1014-nolocal.srctgt.ll: refactor a bit to minimize the change from the original bug report
- Minor comment updates at a few files


aqjune:
A few tests have a lot of metadata// should it be trimmed? :/ They are currently preserved from the original bug report.

nunoplopes:
> A few tests have a lot of metadata// should it be trimmed? :/ They are currently preserved from the original bug report.

yes, please. All that junk just makes debugging harder. The smaller the better.

PR #583, opened 2020-11-14
memory: ptr_alias entry should be properly dropped when merging

aqjune:
When merging two memory's `ptr_alias` in `Memory::mkIf`, entries that exist in `then` but not in `els` (and vice versa) should be dropped.

The attached test raises memory mismatch because ptr_alias in tgt is incorrectly cached that block id 0 does not alias with `@a`.

`store i64 ...`'s access size (8) is larger than `@a`(4), so `Memory::mayalias` answers that `%a2`'s bid (which is 0) does not alias with `@a`.
This is correct in the `unreachable` block, but not in the `exit` block. 

edit: actually I found the validity of this patch depends on what is the semantics of `ptr_alias`. If we say that `ptr_alias` is valid only when there was no UB, then I think this patch is valid. Otherwise, adding bid to `ptr_alias` based on `Memory::mayalias` is slightly controversial because `Memory::mayalias` considers offset as well.


nunoplopes:
I forgot I had fixed this already in the new memory branch. I just didn't have a test case.
I backported the fix: https://github.com/AliveToolkit/alive2/commit/4d1956d7cf93e66f65a076302cb8b59aadd0061c
Thanks for the test case! 😀

PR #584, opened 2020-11-18
MD_misexpect is gone

regehr:


regehr:
is this an OK fix, @LebedevRI ?

LebedevRI:
I suspect so, at least until someone comes up to fix the original patch upstream...

Issue #585, opened 2020-11-23
Unit tests for clang plugin


aqjune:
I think it would be great if we have unit tests that use clang plugin if `-DCLANG_PLUGIN` is set.

I'd like to revive https://github.com/AliveToolkit/alive2/pull/541 , but the patch is pretty complex, so I'd feel more comfortable if there is a test for this.

If this direction seems okay, I'll prepare a PR that adds support for .c/.cpp unit tests to lit.

aqjune:
(Maybe on Wed or after then, since I'll be offline for a few days)

regehr:
I think we should definitely have tests for this. I'd also like to add Compiler Explorer support for C/C++.

PR #588, opened 2020-11-25
clang tv: add lit test

aqjune:
This PR adds lit test for clang tv (#585 ).

It invokes `scripts/alivecc` or `scripts/alive++` if the file exists (otherwise it marks the test as unsupported).

Another minor edit in this PR is that `scripts/alivecc.in` is updated to use `.dylib` extension when used in Mac.

### DIFF
--- a/tests/lit/lit/formats/alive2test.py
+++ b/tests/lit/lit/formats/alive2test.py
@@ -49,9 +49,9 @@ class Alive2Test(TestFormat):
   def __init__(self):
     self.regex_errs = re.compile(r";\s*(ERROR:.*)")
     self.regex_xfail = re.compile(r";\s*XFAIL:\s*(.*)")
-    self.regex_args = re.compile(r"(;|//)\s*TEST-ARGS:(.*)")
-    self.regex_check = re.compile(r"(;|//)\s*CHECK:(.*)")
-    self.regex_check_not = re.compile(r"(;|//)\s*CHECK-NOT:(.*)")
+    self.regex_args = re.compile(r"(?:;|//)\s*TEST-ARGS:(.*)")
+    self.regex_check = re.compile(r"(?:;|//)\s*CHECK:(.*)")
+    self.regex_check_not = re.compile(r"(?:;|//)\s*CHECK-NOT:(.*)")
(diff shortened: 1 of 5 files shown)


regehr:
oh no, I totally forgot about dylib, thanks!!

tests/lit/lit/formats/alive2test.py
```diff
@@ -49,9 +49,9 @@ class Alive2Test(TestFormat):
   def __init__(self):
     self.regex_errs = re.compile(r";\s*(ERROR:.*)")
     self.regex_xfail = re.compile(r";\s*XFAIL:\s*(.*)")
-    self.regex_args = re.compile(r";\s*TEST-ARGS:(.*)")
-    self.regex_check = re.compile(r";\s*CHECK:(.*)")
-    self.regex_check_not = re.compile(r";\s*CHECK-NOT:(.*)")
+    self.regex_args = re.compile(r"(;|//)\s*TEST-ARGS:(.*)")
```
nunoplopes, line 52: you can use non-capturing parentheses here to avoid all the changes below
e.g. `(?:;|//)

tests/lit/lit/formats/alive2test.py
```diff
@@ -124,19 +134,26 @@ def execute(self, test, litConfig):
...
+    if clang_tv and exitCode != 0:
+      # clang tv should not exit with non-zero even if validation fails.
+      # Otherwise it will stop a build system such as `make`.
       return lit.Test.FAIL, output
 
     expect_err = self.regex_errs.search(input)
     if expect_err is None and xfail is None and chk is None and chk_not is None:
       # If there's no other test, correctness of the transformation should be
       # checked.
+      # In case of clang tv, it may have multiple results, so ignore validation
```
nunoplopes, line 153: this is kinda dangerous. why do we want to support this? Is this needed for any of the tests? if so, can you change them so this is not needed?

aqjune: issue569.cpp needed this, because one of the passes raised precondition is false error.
I modified issue569.cpp to manually check 'transformation seems to be correct' keyword to avoid the issue.

PR #592, opened 2020-11-26
fix tests

aqjune:
Fixes tests that are broken after updates of sret/byval in LLVM.
Also, this fixes one test broken by #590 and adds another test about align and dereferenceable.


tests/alive-tv/opt-memory/dereferenceable2.srctgt.ll
```diff
@@ -0,0 +1,15 @@
...
+  %v1 = load i32, i32* %p, align 4
+  ret i32 %v1
+}
+
+define i32 @tgt(i32* dereferenceable(2) %p) {
+  %v1 = load i32, i32* %p, align 4
+  ret i32 %v1
+}
+
+; TODO - since %p is dereferenced with a load with align 4, min_access_size can be
```
nunoplopes, line 13: leaving TODOs is not useful. No one is grepping for TODO in the code.
I'm removing this file. This case it probably doesn't show up in real life anyway.

PR #594, opened 2020-11-26
change attr's align field to store the original value

aqjune:
The recent two crashes are due to the inconsistent use of attr's align field.

The field is storing log(alignment), but ir/value.cpp was assuming the value was simply storing the alignment.

This PR changes attr's align field to store the original value

### DIFF
--- a/ir/attrs.h
+++ b/ir/attrs.h
@@ -15,11 +15,11 @@ public:
                    ReadOnly = 1<<3, ReadNone = 1<<4, Dereferenceable = 1<<5,
                    NoUndef = 1<<6, Align = 1<<7, Returned = 1<<8 };
 
-  ParamAttrs(unsigned bits = None) : bits(bits), align(1) {}
+  ParamAttrs(unsigned bits = None) : bits(bits) {}
 
   uint64_t derefBytes; // Dereferenceable
   uint64_t blockSize;  // exact block size for e.g. byval args
-  uint64_t align;      // Align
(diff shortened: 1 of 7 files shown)


ir/attrs.h
```diff
@@ -9,6 +9,7 @@ namespace IR {
 
 class ParamAttrs final {
   unsigned bits;
+  uint64_t align;      // power-of-2 in bytes
```
aqjune, line 12: It was uint64_t, but char type would be sufficient for this; I'll address this in this PR if okay.

tools/transform.cpp
```diff
@@ -677,7 +677,7 @@ static void calculateAndInitConstants(Transform &t) {
         continue;
 
       uint64_t align = i->hasAttribute(ParamAttrs::Align) ?
-          1ull << i->getAttributes().align : 1;
+          i->getAttributes().getAlign() : 1;
```
nunoplopes, line 680: should getAlign just return 1 when not set?

nunoplopes:
why don't we just change the field to store the alignment in bytes and default to 1?
storing in log2 doesn't seem to help at all. The only place we need the bits is in memory.cpp, but the API seems t take bytes, so no issues.

aqjune:
I thought log2 was used in the past to compactly store the alignment value?
But I have no strong opinion, and it was uint64_t (which was a bit strange).
Anyway, I'll change the field to store the value.

PR #596, opened 2020-11-27
consider fncall param aligns when calculating access size

### LINKED ISSUE #480: Take advantage of align function/argument attributes
Continuation of https://github.com/AliveToolkit/alive2/pull/442
This should take align into account as an optimization.

aqjune:
This PR resolves #480 .

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1565,19 +1565,28 @@ uint64_t FnCall::getMaxAccessSize() const {
 }
 
 FnCall::ByteAccessInfo FnCall::getByteAccessInfo() const {
-  bool has_deref = getAttributes().has(FnAttrs::Dereferenceable);
-  if (!has_deref) {
-    for (auto &arg : args)
-      if (arg.second.has(ParamAttrs::Dereferenceable)) {
-        has_deref = true;
-        break;
-      }
+  auto &retattr = getAttributes();
+  bool has_deref = retattr.has(FnAttrs::Dereferenceable);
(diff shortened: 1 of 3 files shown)


aqjune:
No regression found from LLVM unit tests

ir/instr.cpp
```diff
@@ -1565,19 +1565,28 @@ uint64_t FnCall::getMaxAccessSize() const {
...
-      }
+  auto &retattr = getAttributes();
+  bool has_deref = retattr.has(FnAttrs::Dereferenceable);
+  uint64_t bytesize = has_deref ? gcd(retattr.derefBytes, retattr.align) : 0;
+
+  for (auto &arg : args) {
+    if (!arg.first->getType().isPtrType())
+      continue;
+
+    if (arg.second.has(ParamAttrs::Dereferenceable)) {
```
nunoplopes, line 1576: what if the function takes 2 pointers, one with dereferenceable/align and one without? Then min access needs to be 1, right?

aqjune: I think in the case min access can be larger than 1, but due to a non-trivial reason.
I'll describe why I think so.

`call f(dereferenceable(4) align 4 %p)` is equivalent to a dummy load followed by the call:

```
load i32 %p, align 4 // a dummy load
call f(%p)
```

The dummy load contributes to the min access size, as a real load instruction does.

However, for the `call f(%p)`, I think it shouldn't set the min access size to 1; otherwise it deoptimizes too many things.
I think we can assume that a call updates the memory with min access size granularity.

Assuming that this dummy load idea makes sense, I think pointer arguments without dereferenceable shouldn't update the min access size.

PR #599, opened 2020-11-29
this patch allows the clang plugin to speedup by running TV tasks in

regehr:
subprocesses; it uses the GNU make jobserver to limit the number of
concurrently executing tasks.

http://make.mad-scientist.net/papers/jobserver-implementation/

MISSING IN THIS PR: getting the output straight. currently it is all jumbled together by concurrent processes. I will take care of that part next, but since the changes were getting largish I wanted to see about getting this merged first.

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -1,12 +1,13 @@
 // Copyright (c) 2018-present The Alive2 Authors.
 // Distributed under the MIT license that can be found in the LICENSE file.
 
-#include "llvm_util/llvm2alive.h"
 #include "ir/memory.h"
+#include "llvm_util/llvm2alive.h"
 #include "smt/smt.h"
 #include "smt/solver.h"
 #include "tools/transform.h"
 #include "util/config.h"
+#include "util/parallel.h"
 #include "util/stopwatch.h"
 #include "util/version.h"
 #include "llvm/ADT/Any.h"
@@ -162,6 +163,16 @@ llvm::cl::opt<unsigned> opt_tgt_unrolling_factor(
     llvm::cl::desc("Unrolling factor for tgt function (default=0)"),
     llvm::cl::cat(TVOptions), llvm::cl::init(0));
 
+llvm::cl::opt<bool> parallel_tv(
+    "tv-parallel",
+    llvm::cl::desc("Distribute TV load across cores (requires GNU Make, default=false)"),
+    llvm::cl::cat(TVOptions), llvm::cl::init(false));
+
+llvm::cl::opt<int> max_subprocesses(
+    "tv-max-subprocesses",
+    llvm::cl::desc("Maximum children this clang instance will have at any time (default=128)"),
+    llvm::cl::cat(TVOptions), llvm::cl::init(128));
+
 struct FnInfo {
   Function fn;
   unsigned order;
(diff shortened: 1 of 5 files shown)


regehr:
hmm, I'll look into the travis failures.

in the meantime here's what my group's big machine looks like when running TV on GNU make itself.
Screenshot from 2020-11-29 15-49-54

tv/tv.cpp
```diff
@@ -162,6 +163,19 @@ llvm::cl::opt<unsigned> opt_tgt_unrolling_factor(
...
+llvm::cl::opt<bool> parallel_tv(
+    "tv-parallel",
+    llvm::cl::desc("Distribute TV load across cores (requires GNU Make, default=false)"),
+    llvm::cl::cat(TVOptions), llvm::cl::init(false));
+llvm::cl::opt<int> max_subprocesses(
+    "tv-max-subprocesses",
+    llvm::cl::desc("Maximum children this clang instance will have at any time (default=128)"),
+    llvm::cl::cat(TVOptions), llvm::cl::init(128));
+#else
+bool parallel_tv = false;
```
nunoplopes, line 176: I think this can be enabled for both opt & clang plugins

util/parallel.cpp
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) 2018-present The Alive2 Authors.
+// Distributed under the MIT license that can be found in the LICENSE file.
+
+#include "llvm/Support/ErrorHandling.h"
```
nunoplopes, line 4: this part of the code can't depend on LLVM

util/parallel.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include <cstdlib>
+#include <cstring>
+#include <fcntl.h>
+#include <iostream>
+#include <sys/wait.h>
+#include <unistd.h>
+
+namespace parallel {
+
+char jobserver_token;
```
nunoplopes, line 16: these can be made static (outside the namespace)

util/parallel.cpp
regehr: ugh thanks yes I was just figuring this out on my own :(

tv/tv.cpp
regehr: got it.

I think it might as well be turned on by default since none of the parallel stuff will work at all unless the jobserver is detected, but we can switch this later if we want.

regehr:
want to add a comment that there's a defect in the jobserver protocol that I can't work around and it is annoying.

the problem is that when you do a parallel make, the jobserver flags are unconditionally passed to child processes in the environment, but the file descriptors mentioned are not necessarily open. that is, if make doesn't think a child process is a recursive make (and this is the default behavior on modern version of make) it will close the file descriptors. so far there's no problem -- the code I wrote gracefully handles the case where the file descriptors are mentioned in the environment, but cannot be used.

the bad thing is that if the file descriptors are being used, but some other purpose, there's no way to tell. this can cause really confusing problems. 

I cannot see any way around this without modifying GNU make, it is just not a robust design.

the solution in practice is to use a GNU make with a one line hack that causes it to never close the jobserver file descriptors, but if you forget to do this, the resulting failures can be very confusing.

PR #600, opened 2020-11-30
Add support for align, nonnull assume operand bundle

### LINKED ISSUE #508: Add support for more llvm.assume bundles
e.g.. we miss support for align() assume bundle in AlignmentFromAssumptions/simple.ll

aqjune:
This PR resolves #508 .

This implements the semantics discussed in https://reviews.llvm.org/D90529#2597302 .


ir/instr.cpp
```diff
@@ -2373,48 +2373,109 @@ unique_ptr<Instr> Return::dup(const string &suffix) const {
...
+  case Range: {
+    // assume(val, low1, high1, low2, high2, ...)
+    // Should this also be UB if the input value is poison?
+    const auto &val = s[*args[0]];
+    expr comp(false);
+    for (unsigned i = 1; i < args.size(); i += 2) {
+      const auto &low = s[*args[i]].value;
+      const auto &high = s[*args[i + 1]].value;
+      // Q: Is this successful even if constants are very large?
+      bool wrap = low.sgt(high).simplify().isTrue();
```
nunoplopes, line 2446: don't simplify here; let this be an expression

llvm_util/llvm2alive.cpp
```diff
@@ -932,50 +937,20 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
-          }
+          operands.emplace_back(
+            get_operand(llvm::mdconst::extract<llvm::ConstantInt>(
+              Node->getOperand(op))));
         }
 
-        if (range)
-          BB->addInstr(make_unique<Assume>(*range, Assume::IfNonPoison));
+        auto kind = ID == LLVMContext::MD_align ? Assume::Align : Assume::Range;
+        BB->addInstr(make_unique<Assume>(*range, kind));
```
nunoplopes, line 953: should be move(operands) I believe

ir/instr.cpp
```diff
@@ -2373,48 +2373,109 @@ unique_ptr<Instr> Return::dup(const string &suffix) const {
...
+    const auto &vptr = s.getAndAddPoisonUB(*args[0]);
+    uint64_t align = *dynamic_cast<IntConst *>(args[1])->getInt();
+
+    Pointer ptr(s.getMemory(), vptr.value);
+    s.addUB(ptr.isAligned(align));
+    break;
+  }
+  case Range: {
+    // assume(val, low1, high1, low2, high2, ...)
+    // Should this also be UB if the input value is poison?
```
nunoplopes, line 2439: no, it's not. please remove this comment.
range is only valid if the value is non-poison. We've discussed this before.

ir/instr.cpp
```diff
@@ -2373,48 +2373,109 @@ unique_ptr<Instr> Return::dup(const string &suffix) const {
...
     s.addUB(np.implies(v != 0));
     break;
   }
   case WellDefined:
-    (void)s.getAndAddPoisonUB(*cond, true);
+    (void)s.getAndAddPoisonUB(*args[0], true);
+    break;
+  case Align: {
+    // assume(ptr, align)
+    const auto &vptr = s.getAndAddPoisonUB(*args[0]);
```
nunoplopes, line 2430: why is it UB to pass a poison ptr? is this explicit in LangRef?
An assume bundle with align is often used after inlining a function that had an align arg. So the semantics needs to be the same. i.e. calling f(align 4 poison) can only be UB iff assume align 4 poison is UB.
We need some investigation to decide on the semantics.

aqjune:
Dropping range and adding support for nonnull and align first since they are treated in the LangRef patch D90529

aqjune:
I'll turn this into ready if LLVM unit test has no regression

aqjune:
It has no regression from LLVM unit tests, so this semantics looks okay for intraprocedural optimizations.

aqjune:
If I have time, I'll look into Attributor to see whether it satisfies the LangRef semantics.

llvm_util/llvm2alive.cpp
```diff
@@ -744,6 +744,15 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
             BB->addInstr(
               make_unique<Assume>(*get_operand(v), Assume::WellDefined));
           }
+        } else if (name == "align") {
+          llvm::Value *ptr = bundle.Inputs[0].get();
+          llvm::Value *align = bundle.Inputs[1].get();
+          vector<Value *> args = {get_operand(ptr), get_operand(align)};
```
nunoplopes, line 750: you need to check the return value of get_operand() as it may fail.

ir/instr.cpp
```diff
@@ -2433,50 +2433,96 @@ unique_ptr<Instr> Return::dup(const string &suffix) const {
...
 
 expr Assume::getTypeConstraints(const Function &f) const {
-  if (kind == WellDefined)
+  switch (kind) {
+  case WellDefined:
     return true;
-  return cond->getType().enforceIntType();
+  case AndNonPoison:
+  case IfNonPoison:
+    return args[0]->getType().enforceIntType(1);
```
nunoplopes, line 2514: don't enforce i1 here. we've used assumes for non-bool types

ir/instr.cpp
```diff
@@ -2433,50 +2433,96 @@ unique_ptr<Instr> Return::dup(const string &suffix) const {
...
+    const auto &vptr = s.getAndAddPoisonUB(*args[0]);
+    uint64_t align = *dynamic_cast<IntConst *>(args[1])->getInt();
+
+    Pointer ptr(s.getMemory(), vptr.value);
+    s.addUB(ptr.isAligned(align));
     break;
   }
+  case NonNull: {
+    // assume(ptr)
+    const auto &vptr = s.getAndAddPoisonUB(*args[0]);
```
nunoplopes, line 2499: I thought we had defined `call @f(nonnull poison)` and `call @f(nonnull null)` to pass a poison pointer rather than giving UB?
Same for the alignment?

aqjune: For assume operand bundles, the consensus made in https://reviews.llvm.org/D90529#2597302 is to respect the current wording in LangRef's assume bundle, which is:
```
Just like for the argument of llvm.assume, if any of the provided guarantees are violated at runtime the behavior is undefined.
```
For function attributes, they have poison semantics unlike these assume bundle, as you said.

ir/instr.cpp
aqjune: Thanks, I'll update this as well as the get_operand check.

llvm_util/llvm2alive.cpp
```diff
@@ -741,9 +741,26 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
+          auto *aptr = get_operand(ptr), *aalign = get_operand(align);
+          if (!aptr || !aalign)
+            return error(i);
+
+          vector<Value *> args = {aptr, aalign};
+          BB->addInstr(make_unique<Assume>(move(args), Assume::Align));
+        } else if (name == "nonnull") {
+          llvm::Value *ptr = bundle.Inputs[0].get();
+          BB->addInstr(
+            make_unique<Assume>(*get_operand(ptr), Assume::NonNull));
```
nunoplopes, line 763: missing check here too

PR #602, opened 2020-12-02
clang tv: fix a use-after-free bug

aqjune:
The TargetLibraryInfoImpl temporary object passed to the constructor of TargetLibraryInfo is destroyed after the constructor is finished.
This causes use-after-free at llvm2alive calls afterward.

Adding John's reduced code as a unit test requires passing
```
-isysroot `xcrun --show-sdk-path`
```
to clang on Mac because the test is including a header file.
I'll add this test after the alivecc script is updated.

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -204,13 +204,14 @@ struct TVPass final : public llvm::FunctionPass {
     });
 
     llvm::TargetLibraryInfo *TLI = nullptr;
+    llvm::TargetLibraryInfoImpl TLIImpl;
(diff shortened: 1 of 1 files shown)


aqjune:
The test to be added should be:
```
#include <stdlib.h>

int a, b;
void c(void) {
   b;
   free(&a);
}
```

nunoplopes:
I guess this issue is reproduceable with any file under asan, right? It shouldn't need to include any system header.

regehr:
argh... I've run into this exact bug so many times when writing C++. it seems like it should work.

PR #603, opened 2020-12-03
run TV from our LLVM plugin in parallel by forking. provide two

regehr:
parallelism managers: one that just forks in uncontrolled fashion (for
testing and debugging -- don't run it on large inputs) and another
that uses the jobserver from GNU make to limit parallelism.

when output in not directed to a report directory, it all just gets
blatted to the terminal in mixed-up fashion by all the different
proceses. when output is directed to a report directory, it goes to a
lot of little individual files and then a utility fix_report_files
stitches it all back together.

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -204,6 +204,8 @@ bool is_clangtv = false;
 fs::path opt_report_parallel_dir;
 const char *parallel_subdir = "tmp_parallel";
 unique_ptr<parallel> parallelMgr;
+default_random_engine re;
+uniform_int_distribution<unsigned> rand;
 
 struct TVPass final : public llvm::FunctionPass {
   static char ID;
@@ -275,8 +277,6 @@ struct TVPass final : public llvm::FunctionPass {
       return false;
 
     if (parallelMgr) {
-      default_random_engine re;
-      uniform_int_distribution<unsigned> rand;
       random_device rd;
       re.seed(rd());      
       string newname;
@@ -387,8 +387,6 @@ struct TVPass final : public llvm::FunctionPass {
     fnsToVerify.insert(opt_funcs.begin(), opt_funcs.end());
 
     if (!report_dir_created && !opt_report_dir.empty()) {
-      default_random_engine re;
-      uniform_int_distribution<unsigned> rand;
       static bool seeded = false;
 
       if (!seeded) {
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -125,7 +125,7 @@ llvm::cl::opt<bool> opt_disable_undef_input(
     llvm::cl::desc("Alive: Assume function input cannot be undef"),
     llvm::cl::cat(TVOptions), llvm::cl::init(false));
 
-llvm::cl::list<std::string>
+llvm::cl::list<string>
     opt_funcs("tv-func",
               llvm::cl::desc("Name of functions to verify (without @)"),
               llvm::cl::cat(TVOptions), llvm::cl::ZeroOrMore,
@@ -184,7 +184,7 @@ llvm::cl::opt<int> max_subprocesses(
 struct FnInfo {
   Function fn;
   unsigned order;
-  std::string fn_tostr;
+  string fn_tostr;
 };
 
 ostream *out;
@@ -203,7 +203,7 @@ bool has_failure = false;
 bool is_clangtv = false;
 fs::path opt_report_parallel_dir;
 const char *parallel_subdir = "tmp_parallel";
-std::unique_ptr<parallel> parallelMgr;
+unique_ptr<parallel> parallelMgr;
 
 struct TVPass final : public llvm::FunctionPass {
   static char ID;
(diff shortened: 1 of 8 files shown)


regehr:
as you'll see, Nuno, there's some even worse stuff in the jobserver client than there was before. nevertheless, I think it's a workable solution and we shouldn't let the gross things bother us too much. and anyway, if someone wants to do a better jobserver, I think this provides good infrastructure.

tv/tv.cpp
```diff
@@ -182,7 +201,12 @@ bool showed_stats = false;
 bool report_dir_created = false;
 bool has_failure = false;
 bool is_clangtv = false;
-
+default_random_engine re;
+uniform_int_distribution<unsigned> rand;
+fs::path opt_report_parallel_dir;
+const char *parallel_subdir = "tmp_parallel";
+parallel *parallel;
```
nunoplopes, line 208: can we make this a unique_ptr so it doesn't leak?

tv/tv.cpp
```diff
@@ -288,19 +352,40 @@ struct TVPass final : public llvm::FunctionPass {
...
+    }
+
     fnsToVerify.insert(opt_funcs.begin(), opt_funcs.end());
 
     if (!report_dir_created && !opt_report_dir.empty()) {
-      static default_random_engine re;
-      static uniform_int_distribution<unsigned> rand;
       static bool seeded = false;
 
       if (!seeded) {
```
nunoplopes, line 391: you missed this bit of the seed while sharing the random_engine. Like this all processes will have the same seed and filenames will clash

regehr: ah, good catch. I've been using the mode where it just overwrites files so never saw that happen.

regehr:
well, what I have here dies at runtime with 
```
error: unable to load plugin '/home/regehr/alive2/build/tv/tv.so': '/home/regehr/alive2/build/tv/tv.so: undefined symbol: _ZN8parallelD2Ev'
```
this demangles to parallel::~parallel(), and I don't understand what's going on there, do you know the fix Nuno?

nunoplopes:
> well, what I have here dies at runtime with
> 
> ```
> error: unable to load plugin '/home/regehr/alive2/build/tv/tv.so': '/home/regehr/alive2/build/tv/tv.so: undefined symbol: _ZN8parallelD2Ev'
> ```
> 
> this demangles to parallel::~parallel(), and I don't understand what's going on there, do you know the fix Nuno?

you should just need to mark the destructor as virtual and override it and it should work I guess.
Not I dunno why it wants a destructor as these are PODs. Should work just fine without IIRC.

regehr:
something about putting it in a unique pointer necessitates the destructor. I no longer pretend to understand c++

regehr:
do you mind taking a quick stab at this? I tried a bunch of things such as what you suggest and just got more and more runtime errors

aqjune:
On my machine, giving an empty body to the destructor seems to work:
```
virtual ~parallel() {};
```
Interestingly, `_ZN8parallelD2Ev` still does not show up even if I run `nm -gU ./build/tv/tv.dylib | grep parallel`. :/ (for linux, the command might be slightly different, but I guess nm would still be used)

regehr:
thanks Juneyoung, I tried things like this and it didn't work, but I'll try more stuff tomorrow.

regehr:
@aqjune ok that works, thanks, that's embarrassing bet I tried many similar things, but apparently not this!

@nunoplopes this PR is ready for more review

nunoplopes:
I'm not sure I like the `include(...)` bit. I understand it was useful to debug and compare against the sequential mode, but after this stuff is debugged, why do we need it?
As long as functions aren't interleaved (which we can ensure), the order doesn't really matter I guess. Maintaining the order vs the additional hassle of having another dir with thousands of files (slow) and then preprocess them doesn't sound like a great tradeoff to me.
Is there any reason to go the multi-file way?

regehr:
> I'm not sure I like the `include(...)` bit. I understand it was useful to debug and compare against the sequential mode, but after this stuff is debugged, why do we need it?
> As long as functions aren't interleaved (which we can ensure), the order doesn't really matter I guess. Maintaining the order vs the additional hassle of having another dir with thousands of files (slow) and then preprocess them doesn't sound like a great tradeoff to me.
> Is there any reason to go the multi-file way?

so you're suggesting that we use file locking to atomically append results to the main file in sort of random order? this is easy, I just didn't think anyone would want to see that.

nunoplopes:
Thinking again, the order is useful for diffing between runs. Still would like to have a solution that doesn't involve thousands of files. The current approach is a bit annoying TBH.

regehr:
one option I thought about it having the master clang process grab output from children as they exit (though a pipe, I suppose) and then it prints all of this out at once before terminating. this isn't too difficult and I can work on it if you want.

nunoplopes:
Yeah, I think something like that would make the tool much more usable. It adds a bit of complexity to the code, as the master then needs to sort the output, but I think it's well worth it.

regehr:
Ok I'll work on this. I think there are only two things to worry about:
- a lot of open file descriptors for all of the pipes but hopefully modern OSes are OK about this
- possible memory consumption issues in the main clang if there's a huge amount of output (for example, a file containing a lot of functions where we unroll loops)

PR #605, opened 2020-12-04
add unit test support for opt plugin

### LINKED ISSUE #95: Add lit test support for opt plugin & alive-tv
Required for e.g. https://github.com/AliveToolkit/alive2/pull/88 because the parser doesn't have support for load/store, etc yet.

aqjune:
This PR resolves #95 : adds support for running lit testing with the opt Alive2 plugin.


aqjune:
Hopefully this won't make conflict with #603

PR #608, opened 2020-12-07
add get_outs/set_outs

aqjune:
I'm sorry for the delay, I had a few things that had to be done today.

This is a super small patch that adds getter/setter for `out` that can be used to change ostream that is used by llvm2alive.cpp .

I think `initializer` does not require ostream argument anymore, but it is left unchanged because it needs updates in tv.


nunoplopes:
@regehr is this what you need?

regehr:
yes I believe so -- at the moment I'm handling everything output from tv.cpp but some other things get printed in llvm_utils and I'm not capturing it and reordering it properly, swapping the output stream should fix that

PR #610, opened 2020-12-07
before, parallel clang-tv mode emits a lot of little files that need

regehr:
to be stitched together using an external utility. this patch does the
stitching internally. every child shares a pipe with the main clang
process and communicate their results back to it that way.


util/parallel.cpp
```diff
@@ -0,0 +1,174 @@
...
+      --subprocesses;
+  }
+  assert(subprocesses == 0);
+}
+
+void parallel::emitOutput(std::stringstream &parent_ss,
+                          std::ofstream &out_file) {
+  ensureParent();
+  std::string line;
+  std::regex rgx("^include\\(([0-9]+)\\)$");
```
nunoplopes, line 162: is this bazooka really needed? Isn't a string_view::starts_with() sufficient?

util/parallel.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include <fcntl.h>
+#include <fstream>
+#include <regex>
+#include <sstream>
+#include <string>
+#include <sys/wait.h>
+#include <unistd.h>
+
+// TODO: read this carefully and make sure it's safe when pids are
+// reused
```
nunoplopes, line 16: uhm, there's no issue ATM since you keep everything in a vector and vector idxs are used. So all good, right?

regehr: well, I had meant to sit down for 5 minutes and make sure I believed this argument. I think I do, I removed the comment.

but also if we were compiling a large file this array could get quite large and it's awkward to reuse slots. I was thinking about replacing it with an unordered_map indexed by PID. that should be totally fine as long as the index only lives for as long as the associated process.

util/parallel.cpp
regehr: I can revise if you want but I like the way it expresses all of the constraints at once instead of spreading them across a few lines of less-obvious code. 

also we're only going to do this operation once per fork + TV call, so it's hardly plausible that this would show up in our profile.

PR #613, opened 2020-12-08
give us a turned-off-by-default global mechanism for timing out a lon…

regehr:
…g TV job, but only when it's run as a subprocess by the LLVM plugin

I made this mechanism print the same thing as the other timeout mechanism, not 100% certain that's the correct decision since people may be curious which timeout is getting triggered. but this seems safest and ensures that our existing results-processing scripts will understand what occurred.

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -26,7 +26,9 @@
 #include <iostream>
 #include <memory>
 #include <random>
+#include <signal.h>
 #include <sstream>
+#include <unistd.h>
 #include <unordered_map>
 #include <utility>
 
@@ -177,6 +179,12 @@ llvm::cl::opt<int> max_subprocesses(
                    "(default=128)"),
     llvm::cl::cat(TVOptions), llvm::cl::init(128));
 
+llvm::cl::opt<long> subprocess_timeout(
+    "tv-subprocess-timeout",
+    llvm::cl::desc("Maximum time, in seconds, that a parallel TV call "
+                   "will be allowed to execeute (default=infinite)"),
+    llvm::cl::cat(TVOptions), llvm::cl::init(-1));
+
 struct FnInfo {
   Function fn;
   unsigned order;
@@ -200,6 +208,13 @@ bool is_clangtv = false;
 fs::path opt_report_parallel_dir;
 unique_ptr<parallel> parallelMgr;
 stringstream parent_ss;
+
+void sigalarm_handler(int) {
+  *out << "ERROR: Timeout\n\n";
+  parallelMgr->finishChild();
+  // this is a fully asynchronous exit, skip destructors and such
+  _Exit(0);
+}
 
 string get_random_str() {
   static default_random_engine re;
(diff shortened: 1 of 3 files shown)


tv/tv.cpp
```diff
@@ -177,6 +179,12 @@ llvm::cl::opt<int> max_subprocesses(
                    "(default=128)"),
     llvm::cl::cat(TVOptions), llvm::cl::init(128));
 
+llvm::cl::opt<long> subprocess_timeout(
+    "tv-subprocess-timeout",
+    llvm::cl::desc("Maximum time, in milliseconds, that a parallel TV call "
```
nunoplopes, line 184: I would suggest we switch this to seconds. This timeout is very macroscopic; it's unlikely we want to go down to ms.

tv/tv.cpp
```diff
@@ -200,6 +208,17 @@ bool is_clangtv = false;
 fs::path opt_report_parallel_dir;
 unique_ptr<parallel> parallelMgr;
 stringstream parent_ss;
+mutex output_mutex;
```
nunoplopes, line 211: this mutex is only locked, never unlocked

tv/tv.cpp
```diff
@@ -359,7 +381,9 @@ struct TVPass final : public llvm::FunctionPass {
     if (parallelMgr) {
       llvm_util_init.reset();
       smt_init.reset();
+      output_mutex.lock();
```
nunoplopes, line 384: Some mutex implementations will complain about destroying a locked mutex.
Leaking stuff all over is a pain for debugging, but since we usually don't have such problems, maybe just use _Exit() here as well?

tv/tv.cpp
regehr: can do. only reason I used milliseconds here was for consistence with our other timeouts.

tv/tv.cpp
regehr: sure. the reason I went with thread+mutex here is that I was worried that a unix signal would interact with Z3 or LLVM poorly in some way, a thread seemed safer.
in the signal version of this code, at this point we would simply mask off signals, avoiding leaking the thread+mutex. that may be a better way to go, I have no idea.

nunoplopes: I had crashes in Z3 in the past when trying to destroy a locked mutex (don't remember which platform, though).
So as-is will crash for sure. FWIW Z3 ATM only catches SIGINT. I've no idea about LLVM.
Not leaking when there's a timeout is just too hard. But at least it would be nice to not leak everything in the non-timeout case (for debugging). Hence calling _exit() in the non-timeout case isn't very nice.
We have at least these 2 alternatives:
 - add some sync mechanism such that the mutex can be unlocked before exit so it doesn't crash when deleted. This is going to look close to what we have in Z3 I guess.
 - Try out signals. Not portable to windows, but I don't think we care.

tv/tv.cpp
nunoplopes: I would rather go with seconds, otherwise we'll be counting bazillion zeros on the command line.

tv/tv.cpp
regehr: sure, I'll try these out and see which one feels nicer

tv/tv.cpp
regehr: done, and this is the only granularity supported by alarm() anyway :)

tv/tv.cpp
regehr: signals are very clean and also I've set it up so that we'll get crashes if anyone is already using SIGALRM

nunoplopes:
LGTM, just waiting for the build bots.
Travis became terribly slow for non-paying projects.. 🙄

PR #618, opened 2020-12-10
TV output doesn't really even work in parallel mode unless

regehr:
-tv-report-dir is specified, so let's enforce that with a nice error
message

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -389,6 +389,13 @@ struct TVPass final : public llvm::FunctionPass {
     if (initialized++)
       return false;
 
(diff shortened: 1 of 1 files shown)


nunoplopes:
Why? I thought the output was buffered and then written at the end, regardless of whether it's a file out cout.
So is this a bug or a fundamental problem?

regehr:
I didn't even check, it didn't seem like something people would want to do. I can make it work if you want.

nunoplopes:
I think it's useful for debugging the very least. Also, I never use the `-tv-report-dir` when compiling the single-file benchmarks; it's just 1 file anyways 😀

regehr:
ok, I'll make sure it works

PR #620, opened 2020-12-13
switch to module pass

aqjune:
After #616 is merged, this PR should be rebased.


nunoplopes:
It would be preferable to keep a function that validates 1 function and call that from runOnModule. That reduces the humongous diff size, and also allows parallel strategies to shrink the set of calls and call the validateFunction() (or whatever name) on the functions each worker is responsible for.

PR #624, opened 2020-12-17
support llvm.trap

aqjune:
This PR uses a function call to encode llvm.trap.

The benefit of encoding it as a function call is that it is easy to check refinement of llvm.trap and other no-return function calls.
I found that it becomes a bit non-trivial for me to do without function call encoding (for example, simply making it do `state.addNoReturn()` will allow the refinement).
We already have a nice framework that works really well for encoding the equivalence of function calls using UB.

A limitation of this encoding is that we can't impose `call llvm.trap()` to never have UB.
It can introduce false positive in theory (because llvm.trap has more behavior than people's thought), but.. I couldn't come up with an example that has false positive.

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -874,7 +874,7 @@ end:
     {
       FnAttrs attrs;
       attrs.set(FnAttrs::NoReturn);
-      attrs.set(FnAttrs::NoRead);
+      attrs.set(FnAttrs::NoWrite);
       return make_unique<FnCall>(*llvm_type2alive(i.getType()),
                                  "", "#trap", move(attrs));
     }
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -870,6 +870,14 @@ end:
         return error(i);
       RETURN_IDENTIFIER(make_unique<Free>(*b, false));
     }
+    case llvm::Intrinsic::trap:
+    {
+      FnAttrs attrs;
+      attrs.set(FnAttrs::NoReturn);
+      attrs.set(FnAttrs::NoRead);
+      return make_unique<FnCall>(*llvm_type2alive(i.getType()),
+                                 "", "#trap", move(attrs));
+    }
 
     // do nothing intrinsics
     case llvm::Intrinsic::dbg_addr:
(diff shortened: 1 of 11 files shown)


llvm_util/llvm2alive.cpp
```diff
@@ -870,6 +870,14 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
         return error(i);
       RETURN_IDENTIFIER(make_unique<Free>(*b, false));
     }
+    case llvm::Intrinsic::trap:
+    {
+      FnAttrs attrs;
+      attrs.set(FnAttrs::NoReturn);
+      attrs.set(FnAttrs::NoRead);
```
aqjune, line 877: Would trap also be NoWrite?
Since trap is noreturn, its output memory won't be used anyway, but I wasn't not sure whether we can say  it is really NoWrite. :/

regehr:
LangRef is not especially clear about the semantics of @llvm.trap, but I think it is completely safe to assume that it simply terminates execution without triggering UB

aqjune:
Thank you. Maybe it would be great if there is a way to enforce that a specific function call never raises UB in Alive2. It won't be trivial because UB is used to encode equivalence of side effects of function calls, but there is a `speculatable` attribute in IR.

aqjune:
No regression in LLVM unit tests.

llvm_util/llvm2alive.cpp
nunoplopes: I think NoWrite should be fine. We still have the UB part to ensure that trap is called.
BTW, can you add a test to ensure trap can't be refined with unreach, like:
trap
unreachable
=>
unreachable

nunoplopes: I'm less sure about noread. We need to double check the semantics. Is this ok?
```
store 1, foo
trap
=>
store 2, foo
trap
```

(please add this test as well)

aqjune: LangRef says llvm.trap may be lowered to a target-specific instruction, so it seems noread is a bit unsafe. :/ I removed it.
I checked that LLVM unit tests that contain `llvm.trap` passed after this change.

PR #635, opened 2020-12-26
Print output earlier

regehr:
buffered output from all children and then dumped it all at once at
the end. this hurt when using alivecc on huge files like sqlite3.c.

this patch dumps output as it goes along. the only tricky thing was
dealing with the fact that children complete out of order.

this also does just a tiny bit of cleanup on the parallel manager.

tested by running alivecc in serial and parallel modes and making sure they print mostly the same thing

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -312,14 +312,10 @@ struct TVPass final : public llvm::ModulePass {
 
       if (pid != 0) {
         /*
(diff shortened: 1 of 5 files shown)


regehr:
this is nice: I can do TV for sqlite3.c in about 6 hours on my huge machine once this patch is applied

PR #637, opened 2020-12-27
clangtv: cache TLIImpl, use scopedwatch only when -tv-elapsed-time is set

aqjune:


regehr:
LGTM assuming that this caching strategy is legal, thanks Juneyoung!

nunoplopes:
Why is this caching legal? TLI is per function, and this cache creates a single TLI for the whole module, right?
I've committed a simple patch to avoid the memory allocation, but I doubt that's the whole story for the slowdown. We need something else.

nunoplopes:
> Why is this caching legal? TLI is per function, and this cache creates a single TLI for the whole module, right?
> I've committed a simple patch to avoid the memory allocation, but I doubt that's the whole story for the slowdown. We need something else.

Ah, never mind, this caches just the function independent bit. I dunno if that's going to give any speedup, but let's check.

PR #638, opened 2020-12-28
two new parallel managers

regehr:
first, a trivial one that doesn't actually call fork, it just drops
all TV work onto the floor. the intent is to make it easy to examine
our sequential critical path by ignoring everything else.

second, a new jobserver based on a named pipe. this ends up being
quite a lot easier and simpler than interacting with the GNU make
jobserver which has some gotchas.

### DIFF
--- a/util/parallel_fifo.cpp
+++ b/util/parallel_fifo.cpp
@@ -1,0 +1,57 @@
+// Copyright (c) 2018-present The Alive2 Authors.
+// Distributed under the MIT license that can be found in the LICENSE file.
+
+#include "util/compiler.h"
+#include "util/parallel.h"
+#include <cassert>
+#include <cstdlib>
+#include <cstring>
+#include <fcntl.h>
+#include <string>
+#include <sys/stat.h>
+#include <unistd.h>
+
+using namespace std;
+
+bool fifo::init() {
+  ENSURE(parallel::init());
+  auto fifo_filename = getenv("ALIVE_JOBSERVER_FIFO");
+  if (!fifo_filename)
+    return false;
+  pipe_fd = open(fifo_filename, O_RDWR);
+  if (pipe_fd < 0)
+    return false;
+  return true;
+}
+
+void fifo::getToken() {
+  ENSURE(read(pipe_fd, &token, 1) == 1);
+}
+
+void fifo::putToken() {
+  ENSURE(write(pipe_fd, &token, 1) == 1);
+}
+
+tuple<pid_t, ostream *, int> fifo::limitedFork() {
+  assert(pipe_fd != -1);
+  auto res = parallel::limitedFork();
+  // child now waits for a jobserver token
+  if (get<0>(res) == 0)
+    getToken();
+  return res;
+}
+
+void fifo::finishChild(bool is_timeout) {
+  parallel::finishChild(is_timeout);
+  putToken();
+}
+
+void fifo::finishParent() {
+  /*
+   * finishParent() basically just blocks -- we'll give up our
+   * parallel execution token until it returns
+   */
+  putToken();
+  parallel::finishParent();
+  getToken();
+}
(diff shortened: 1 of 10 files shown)


util/parallel_fifo.cpp
```diff
@@ -0,0 +1,57 @@
...
+  parallel::finishChild(is_timeout);
+  putToken();
+}
+
+void fifo::finishParent() {
+  /*
+   * finishParent() basically just blocks -- we'll give up our
+   * parallel execution token until it returns
+   */
+  putToken();
```
nunoplopes, line 54: why does this implementation require overbooking tokens?
I remember there was a good reason to do so for the posix one, but does this shouldn't suffer from the same issues?

regehr: there's no requirement, it's just a policy choice to give up the token so someone else can run, since the parent is likely to be blocked for a while here. I originally did this because otherwise I was seeing a bunch of idle cores. it's against the rules for the posix jobserver (but appears to work just fine in practice) and it's not against the rules for our own  job server since I write the rules. we don't have to do it if you don't want, but there's no problem here.

nunoplopes: But the parent doesn't acquire a token, only the children, I think? I believe this is creating more tokens than the max specified number.

regehr: the posix jobserver specifies that the original process has an implicit token, it does not need to acquire or release it. we are free to say something similar, and I in fact I think this makes sense: the process exec'ed by the jobserver should get a token for free without having to acquire it. so here it's just giving the token back while it's blocked. 

look, should I just delete this? it clearly makes you unhappy at some level. I only put it in there since I observed it to give us more even usage of cores, it doesn't really matter, we can just run with a slightly larger N and then the end effect will be not all that different.

nunoplopes: I'm just trying to understand the invariants and the code 😀
If the jobserver has 5 tokens initially, then 5 children can be spawned and work in parallel. But then we hit this code that gives back 1 token and now some other process is able to get a token. We now have 6 child processes running concurrently, while we specified we wanted up to 5.
Is this true?
For the posix one, from what I understood the parent has one token, which is discounted from the pool. So the parent can only spawn 4 children at a time.

It makes sense to me that the parent has a token because it also consumes resources while it's forking and while it's doing all the sequential tasks. But this code is not acquiring a token AFAICT. Only the posix one is.

regehr: well, in my first version of the server I had it put n-1 bytes into the pipe for parallelism level n, but it seemed ugly so I removed it. so yes, if I don't put that back we could end up with n+1 processes running at a time.

there are two things we're trying to avoid:
1. spawning so many processes that we run out of RAM
2. under-utilizing a nice expensive multicore

a bit of overload doesn't violate either of these requirements. if we want to avoid this overload, then I can go back to the n-1. the requirements here are very soft so I would rather just land this patch than argue over details.

regehr: actually there is a fun invariant-related thing that happens when we run under ninja or GNU make. in either case, there are two independent pools of token. every actual clang process (that is, the parents not doing TV work) holds a token from make/ninja, and then TV subprocesses use the alive token pool.

when a parent clang process puts an extra token into the alive2 token pool, we can think of this as alive2 borrowing a token from the make/ninja token pool. the original token isn't actually being used (because the parent clang process only does this thing when it is blocking to wait for all of its children to terminate). so we might end up with quite a few extra alive processes at once, but these always come at the expense of the other token pool. we violate a local invariant but preserve a global one. I find this pleasing and I like how it all seems to work out.

regehr:
well, ugh, I'm trying to do TV for the MultiSource directory in test-suite and I'm seeing >500 runnable processes using this initial command line:
```
alive-jobserver -j64 make -j16
```
I don't understand how things get out of control like this. (I have already commented out the code we're debating above, just to make sure that it is not involved with this problem.)

regehr:
(problem in previous comment is a dumb bug in alivecc, I left the wrong environment variable there. it's not a bug in the jobserver. I'll fix this in next pull req)

PR #639, opened 2020-12-30
ok, this is a collection of small changes, I'm still trying to figure

regehr:
out the best way to do everything here:

- remove the GNU make jobserver client, it was fun to write but it's
  just too ugly and hard to use

- add a couple more environment variables to alivecc

- the script for running TV on test-suite was using two jobservers,
  oops -- stop doing this and also check for this condition and error
  out, in tv.cpp

- offset the number of tokens by -1 if we're exec'ing a process, in
  order to maintain our parallel invariant more accurately

- add a counter that prints the %age of the time llvm2alive works;
  this needs a corresponding change in stats.php to print the
  analogous statistic aggregated across all files, I hope someone
  else will do this so don't have to learn any php

- fix very rare SIGALARM-related crashes, apparently it is not
  possible to execute any non-trivial C++ in signal context, so cache
  the pipe file descriptor to avoid having to look it up using
  vector.back() -- this was really causing crashes! I don't pretend to
  understand the mechanism for this but the crashes are gone now

- instead of letting the parent process freely fork a lot of children
  that immediately block, block the parent from forking until a token
  is available. this was not possible using the GNU make jobserver
  since it could lead to deadlock.


nunoplopes:
I've patched the stats.php script to print vcgen failures. We already print enough information; it was just a mater of collecting the statistics.

PR #642, opened 2021-01-03
X86-specific shufflevector operations

aqjune:
To check whether transformations in Transforms/InstCombine/X86/x86-pshufb.ll is okay after the undef placeholder is replaced with poison, this patch adds support for three cases:

```
; First arg: vector data, second arg: element chooser
declare <16 x i8> @llvm.x86.ssse3.pshuf.b.128(<16 x i8>, <16 x i8>)                                                                                                                                                                                                                   
declare <32 x i8> @llvm.x86.avx2.pshuf.b(<32 x i8>, <32 x i8>)                                                                                                                                                                                                                        
declare <64 x i8> @llvm.x86.avx512.pshuf.b.512(<64 x i8>, <64 x i8>)                                                                                                                                                                                                                  
```

The specification that I referred:

llvm.x86.ssse3.pshuf.b.128: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=pshufb&techs=SSSE3&expand=5153
llvm.x86.avx2.pshuf.b: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=pshufb&techs=AVX2&expand=5156
llvm.x86.avx512.pshuf.b.512: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=pshufb&expand=5156,5159&techs=AVX_512

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -3594,7 +3594,7 @@ unique_ptr<Instr> InsertElement::dup(const string &suffix) const {
 
 
 ShuffleVector::ShuffleVector(
-    Type &type, std::string &&name, Value &v1, Value &v2, Op op,
+    Type &type, string &&name, Value &v1, Value &v2, Op op,
     vector<unsigned> mask_)
     : Instr(type, move(name)), v1(&v1), v2(&v2), op(op), mask(move(mask_)) {
   // Shufflevector operations other than LLVMIR_ShufVec do not take const mask
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -3615,7 +3615,7 @@ void ShuffleVector::print(ostream &os) const {
   const char *str = nullptr;
   switch(op) {
   case LLVMIR_ShufVec:   str = "shufflevector"; break;
-  case PShufB:  str = "pshuf"; break;
+  case PShufB:  str = "pshufb"; break;
   }
   os << getName() << " = " << str << " " << *v1 << ", " << *v2;
   if (op == LLVMIR_ShufVec)
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -3657,8 +3657,9 @@ StateValue X86PShufB::toSMT(State &s) const {
       vec->getType().getAsAggregateType());
   vector<StateValue> vals;
 
+  auto &mask_val = s[*mask];
   for (unsigned i = 0; i < vty->numElementsConst(); ++i) {
-    auto m = vty->extract(s[*mask], expr::mkUInt(i, 64));
+    auto m = vty->extract(mask_val, expr::mkUInt(i, 64));
     // mask chooses the element from its own slide, which is 16-elems wide
     auto idx = (m.value & expr::mkUInt(15, m.value))
                + expr::mkUInt(i & 0xF0, m.value);
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -3593,6 +3593,16 @@ unique_ptr<Instr> InsertElement::dup(const string &suffix) const {
 }
 
 
+ShuffleVector::ShuffleVector(
+    Type &type, std::string &&name, Value &v1, Value &v2, Op op,
+    std::vector<unsigned> mask_)
+    : Instr(type, move(name)), v1(&v1), v2(&v2), op(op),
+      mask(std::move(mask_)) {
+  // Shufflevector operations other than LLVMIR_ShufVec do not take const mask
+  assert((op != LLVMIR_ShufVec && mask.size() == 0) ||
+         (op == LLVMIR_ShufVec && mask.size() != 0));
+}
+
 vector<Value*> ShuffleVector::operands() const {
   return { v1, v2 };
 }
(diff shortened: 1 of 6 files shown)


aqjune:
A tricky part is its interaction with shufflevector because shufflevector may return undef, see:
```
define <16 x i8> @fold_with_undef_elts(<16 x i8> %InVec) {                                                                                                                                                                                                                            
%0:                                                                                                                                                                                                                                                                                   
  %1 = ssse3.pshuf.b.128 <16 x i8> %InVec, <16 x i8> { 0, 128, undef, 128, 1, 128, undef, 128, 2, 128, undef, 128, 3, 128, undef, 128 }                                                                                                                                               
  ret <16 x i8> %1                                                                                                                                                                                                                                                                    
}                                                                                                                                                                                                                                                                                     
=>                                                                                                                                                                                                                                                                                    
define <16 x i8> @fold_with_undef_elts(<16 x i8> %InVec) {                                                                                                                                                                                                                            
%0:                                                                                                                                                                                                                                                                                   
  %1 = shufflevector <16 x i8> %InVec, <16 x i8> { 0, poison, poison, poison, poison, poison, poison, poison, poison, poison, poison, poison, poison, poison, poison, poison }, 0, 16, 4294967295, 16, 1, 16, 4294967295, 16, 2, 16, 4294967295, 16, 3, 16, 4294967295, 16            
  ret <16 x i8> %1                                                                                                                                                                                                                                                                    
}                                                                                                                                                                                                                                                                                     
ERROR: Target's return value is more undefined                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                      
Example:                                                                                                                                                                                                                                                                              
<16 x i8> %InVec = < #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0), #x00 (0) >                                                                                                 
                                                                                                                                                                                                                                                                                      
Source:                                                                                                                                                                                                                                                                               
<16 x i8> %1 = < #x00 (0), #x00 (0), poison, #x00 (0), #x00 (0), #x00 (0), poison, #x00 (0), #x00 (0), #x00 (0), poison, #x00 (0), #x00 (0), #x00 (0), poison, #x00 (0) >                                                                                                             
                                                                                                                                                                                                                                                                                      
Target:                                                                                                                                                                                                                                                                               
<16 x i8> %1 = < #x00 (0), #x00 (0), #xfc (252, -4), #x00 (0), #x00 (0), #x00 (0), #xfc (252, -4), #x00 (0), #x00 (0), #x00 (0), #xfc (252, -4), #x00 (0), #x00 (0), #x00 (0), #xfc (252, -4), #x00 (0) >                                                                             
Source value: < #x00 (0), #x00 (0), poison, #x00 (0), #x00 (0), #x00 (0), poison, #x00 (0), #x00 (0), #x00 (0), poison, #x00 (0), #x00 (0), #x00 (0), poison, #x00 (0) >                                                                                                              
Target value: < #x00 (0), #x00 (0), #xfc (252, -4), #x00 (0), #x00 (0), #x00 (0), #xfc (252, -4), #x00 (0), #x00 (0), #x00 (0), #xfc (252, -4), #x00 (0), #x00 (0), #x00 (0), #xfc (252, -4), #x00 (0) >                                                                              
```
Should we say that these target-specific intrinsic can return poison as well if the chooser is non-negative & larger than element size?

llvm_util/llvm2alive.cpp
```diff
@@ -915,6 +916,25 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
     case llvm::Intrinsic::prefetch:
       return NOP(i);
 
+    // X86 intrinsics
```
nunoplopes, line 919: can you move this above the NOP ones pls?

ir/instr.cpp
```diff
@@ -3603,23 +3613,55 @@ void ShuffleVector::rauw(const Value &what, Value &with) {
 }
 
 void ShuffleVector::print(ostream &os) const {
-  os << getName() << " = shufflevector " << *v1 << ", " << *v2;
-  for (auto m : mask)
-    os << ", " << m;
+  const char *str = nullptr;
+  switch(op) {
+  case LLVMIR_ShufVec:   str = "shufflevector"; break;
+  case SSSE3_PShufB128:  str = "ssse3.pshuf.b.128"; break;
```
nunoplopes, line 3619: can we just a single pshuf? There's not need for all these AFAICT

ir/instr.cpp
```diff
@@ -3593,6 +3593,16 @@ unique_ptr<Instr> InsertElement::dup(const string &suffix) const {
 }
 
 
+ShuffleVector::ShuffleVector(
```
nunoplopes, line 3596: no std:: here

ir/instr.cpp
```diff
@@ -3603,23 +3613,55 @@ void ShuffleVector::rauw(const Value &what, Value &with) {
...
+  }
+  os << getName() << " = " << str << " " << *v1 << ", " << *v2;
+  if (op == LLVMIR_ShufVec)
+    for (auto m : mask)
+      os << ", " << m;
 }
 
 StateValue ShuffleVector::toSMT(State &s) const {
-  auto vty = v1->getType().getAsAggregateType();
+  auto vty = static_cast<const VectorType*>(v1->getType().getAsAggregateType());
```
nunoplopes, line 3630: we can't do a static cast on types because of Symbolic Types. And I don't think it's needed (see below).

ir/instr.cpp
```diff
@@ -3603,23 +3613,55 @@ void ShuffleVector::rauw(const Value &what, Value &with) {
...
 
-  for (auto m : mask) {
-    if (m >= 2 * sz) {
-      vals.emplace_back(UndefValue(vty->getChild(0)).toSMT(s).value, true);
-    } else {
-      auto *vect = &s[m < sz ? *v1 : *v2];
-      vals.emplace_back(vty->extract(*vect, m % sz));
+  auto x86_shuffle = [&]() {
+    for (unsigned i = 0; i < vty->numElementsConst(); ++i) {
+      auto mask = vty->extract(s[*v2], expr::mkUInt(i, 64));
```
nunoplopes, line 3636: aggregate type has an extract with an unsigned as arg

ir/instr.cpp
nunoplopes: also, rather than implementing new phsuf, why not lower them into shufflevector in llvm2alive?

nunoplopes: plus a add a test for the semantics pls

nunoplopes (changes requested):

ir/instr.cpp
aqjune: Test added.
pshufb states that the negative mask returns zero element unlike shufflevector, so I think it is slightly different from itl.

ir/instr.cpp
aqjune: Okay.
Interestingly, syntactically it accepts variable mask, but it crashes at SelDag. I'll update this

aqjune: Eh, I found that these intrinsics can sometimes take variable masks :/
We need VectorType's extract.
See test/Transforms/InstCombine/X86/x86-pshufb.ll 's demanded_elts_insertion().

ir/instr.cpp
aqjune: ok

ir/instr.cpp
aqjune: InsertElement::toSMT has the static cast too. Should I change it too?

llvm_util/llvm2alive.cpp
aqjune: moved

tests/alive-tv/vector/x86-pshufb.src.ll
```diff
@@ -0,0 +1,287 @@
+; ModuleID = 'x86-pshufb.ll'
+target triple = "x86_64-unknown-unknown"
+
+; These are copied from Transforms/InstCombine/X86/x86-pshufb.ll.
```
nunoplopes, line 4: please don't do this. The copyright belongs to LLVM. We shouldn't bring whole big files in.

ir/instr.cpp
nunoplopes: yes, please!

nunoplopes:
pshuf and shufflevector seem to share nothing. Should pshufb just be a new instruction?

tests/alive-tv/vector/x86-pshufb.src.ll
aqjune: Oh, I see. I'll write my own tests.

aqjune:
Yep, that makes sense. I'll update this.

ir/instr.cpp
nunoplopes: I take this back, sorry. Since that operand can only be a vector, there's no issue. The problem is when the operand may have more than one type (e.g., int or vector of ints).

aqjune: Oh, okay. No problem.

aqjune:
Hmm, the undef error still happens. :/ On my laptop, it happens when -disable-undef-input is given (otherwise TO).
[fold_with_undef_elts.zip](https://github.com/AliveToolkit/alive2/files/5770237/fold_with_undef_elts.zip)

ir/instr.cpp
```diff
@@ -3639,6 +3639,50 @@ unique_ptr<Instr> ShuffleVector::dup(const string &suffix) const {
...
+  os << getName() << " = pshufb " << *vec << ", " << *mask;
+}
+
+StateValue X86PShufB::toSMT(State &s) const {
+  auto vty = static_cast<const VectorType*>(
+      vec->getType().getAsAggregateType());
+  vector<StateValue> vals;
+
+  for (unsigned i = 0; i < vty->numElementsConst(); ++i) {
+    auto m = vty->extract(s[*mask], expr::mkUInt(i, 64));
```
nunoplopes, line 3661: you evaluate mask only once outside the loop

aqjune:
Resolving the undef issue seems to require defining our own semantics of pshufb w.r.t undef/poison, which needs a discussion..
Shall we revisit this issue after the memory related thing?

aqjune:
Hmm.. I think I won't have time for this patch, at least until the end of my military camp (which will be around the end of May). :) There are still ~3 weeks left, but I have bunch of things to do...
I'll revisit this patch when the shufflevector semantics is finally going to be fixed. It is currently blocked by the select -> and/or optimization.

Issue #493, opened 2020-09-27
Inconsistent implementation of nonnull & dereferenceable for function calls


nunoplopes:
In `unpack_inputs` we have the following code:
```cpp
Pointer p(s.getMemory(), move(value.value));
p.stripAttrs();
if (argflag.has(ParamAttrs::Dereferenceable))
  s.addUB(p.isDereferenceable(argflag.derefBytes, bits_byte / 8, false));
if (argflag.has(ParamAttrs::NonNull))
  s.addUB(p.isNonZero());
```

`value.non_poison` is never mentioned. It has to be, as `value.value` is not meaningful when `value.non_poison=false`

Then we have in `pack_return`:
```cpp
if (ty.isPtrType() && (isDeref || isNonNull)) {
  Pointer p(s.getMemory(), ret.value);
  s.addUB(ret.non_poison);
  if (isDeref)
    s.addUB(p.isDereferenceable(attrs.getDerefBytes()));
  if (isNonNull)
    s.addUB(p.isNonZero());
}
```

Here we assume nonnull & dereferenceable imply non-poison.
LangRef is omissive in this case.

It seems some optimizations in LLVM assume that a nonnull pointer might be pointer. For example, this transformation happens:
```llvm
%q = gep inbounds * %p, 1
%c = call i32 @f(* %q)
  =>
%q = gep inbounds * %p, 1
%c = call i32 @f(nonnull * %q)
```

We need:
- [ ] audit how LLVM relies on these attributes to understand the desired semantics
- [ ] clarify the semantics in LangRef
- [ ] fix `unpack_input` to mention `value.non_poison` somehow (depending on chosen semantics)
- [ ] fix `pack_return` if needed (depending on chosen semantics)

nunoplopes:
@aqjune you looked at this stuff not too long ago. Were there any conclusions back then? Can you handle this pls?

aqjune:
`poisonImpliesUB()` returns true when the attribute is nonnull or dereferenceable. `FnCall::toSMT` checks it and calls `getAndAddPoisonUB` instead. Currently we're assuming that passing poison to derefernceable and nonnull attributes is UB.
After ongoing patches on LLVM (malloc and GVN) has landed, I will send a mail to llvm-dev to clarify the semantics of nonnull and dereferenceable. A good news is that now we have noundef attribute in LLVM, so the current behavoir of nonnull/dereferenceable can be achieved by using noundef.

aqjune:
[...]

PR #649, opened 2021-01-17
make nonnull/align pass poison rather than raise UB

aqjune:
Since the nonnull patch (https://reviews.llvm.org/D90529) is likely to be landed soon..! (this Tuesday, unless there is a new request)
This is a patch that makes nonnull/align use the poison semantics rather than immediate UB.

LLVM unit tests running...

Relevant issue: #493

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1692,10 +1692,6 @@ static void unpack_inputs(State &s, Value &argv, Type &ty,
       assert(!argflag.poisonImpliesUB() || value.non_poison.isTrue());
 
       expr np(true);
-        // align and nonnull: the arg becomes poison if the condition isn't met
-      bool has_poison_attr =
-          argflag.has(ParamAttrs::Align) || argflag.has(ParamAttrs::NonNull);
-
       Pointer p(s.getMemory(), move(value.value));
       p.stripAttrs();
       if (argflag.has(ParamAttrs::Dereferenceable) ||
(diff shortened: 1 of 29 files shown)


aqjune:
BTW, I'm fine with landing this after the LangRef patch is landed

ir/instr.cpp
```diff
@@ -1687,19 +1687,30 @@ static void unpack_inputs(State &s, Value &argv, Type &ty,
 
   auto unpack = [&](StateValue &&value) {
     if (ty.isPtrType()) {
+      expr np = value.non_poison;
```
nunoplopes, line 1690: there's no need for this copy

ir/instr.cpp
```diff
@@ -1687,19 +1687,30 @@ static void unpack_inputs(State &s, Value &argv, Type &ty,
...
       p.stripAttrs();
       if (argflag.has(ParamAttrs::Dereferenceable) ||
           argflag.has(ParamAttrs::ByVal))
         s.addUB(
           p.isDereferenceable(argflag.getDerefBytes(), argflag.align, false));
-      else if (argflag.has(ParamAttrs::Align))
-        s.addUB(p.isAligned(argflag.align));
+      else if (argflag.has(ParamAttrs::Align)) {
+        // align and nonnull: the arg becomes poison if the condition isn't met
+        if (argflag.poisonImpliesUB())
```
nunoplopes, line 1700: instead of doing all this case splitting on poisonImpliesUB() which leads to a lot of code duplication, I suggest to collect all poison first and then move it to UB().

nunoplopes:
This patch is missing the corresponding changes to `Input::mkInput`

aqjune:
Yep, I'll address the comments
LLVM unit test result: has a failure at Transforms/InstCombine/call_nonnull_arg.ll , but my LLVM patch already fixes the failure as well

aqjune:
I had to update `Return::toSMT` as well.

I'm running LLVM unit tests again; will leave a comment when it's done

aqjune:
LLVM unit tests: has no regression except Transforms/InstCombine/call_nonnull_arg.ll (which is updated by the LLVM patch)

PR #654, opened 2021-01-26
factor out bid-based nonlocal block analysis

aqjune:
This breaks the symmetry of fncall's extra block and other nonlocals (the detail is sent as a mail separately)

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -35,13 +35,12 @@ using namespace util;
 
 //--- Functions for non-local block analysis based on bid ---//
 
-// If include_tgtonly is false, bid cannot be the global vars existing in target
-// only
-static bool nonlocal_is_globalvar(unsigned bid, bool include_tgtonly) {
-  if (bid < has_null_block)
-    return false;
-  return bid < has_null_block + num_globals_src ||
-         (include_tgtonly && num_nonlocals_src <= bid && bid < num_nonlocals);
+// If include_tgt is true, return true if bid is a global var existing in target
+// only as well
+static bool nonlocal_is_globalvar(unsigned bid, bool include_tgt) {
+  bool srcglb = has_null_block <= bid && bid < has_null_block + num_globals_src;
+  bool tgtglb = num_nonlocals_src <= bid && bid < num_nonlocals;
+  return srcglb || (include_tgt && tgtglb);
 }
 
 static bool nonlocal_is_constglb(unsigned bid) {
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -73,14 +73,7 @@ static bool always_noread(unsigned bid) {
 
 // bid: nonlocal block id
 static bool always_nowrite(unsigned bid) {
-  bool flag = bid <  num_consts_src + has_null_block ||
-              bid >= num_nonlocals_src + num_extra_nonconst_tgt;
-
-  // If a block cannot be read, it cannot be written as well
-  assert(!always_noread(bid) || flag);
-
-  // TODO: fncall mem cannot be accessed with loads and stores.
-  return flag;
+  return always_noread(bid) || is_constglb(bid);
 }
 
 
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -18,7 +18,9 @@ using namespace util;
 
   // Non-local block ids (assuming that no block is optimized out):
   // 1. null block: 0
-  // 2. global vars in source: 1 ~ num_globals_src
+  // 2. global vars in source:
+  //      1 ~ num_consts_src                    (constant globals)
+  //      num_consts_src + 1 ~ num_globals_src  (non-constant globals)
   // 3. pointer argument inputs:
   //      num_globals_src + 1 ~ num_globals_src + num_ptrinputs
   // 4. a block reserved for encoding the memory touched by calls:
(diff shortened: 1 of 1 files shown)


aqjune:
The goal of this patch is mainly for improving alias stats: this brings about 1.5% increase to % of blk ≤ 3 for ph7 and sqlite3.
I'm running experiment to see the performance change of the single file benchmark as well.

ir/memory.cpp
```diff
@@ -1443,7 +1443,7 @@ Memory::Memory(State &state) : state(&state), escaped_local_blks(*this) {
     return;
 
   next_nonlocal_bid
```
nunoplopes, line 1445: the function that updates this needs tweaking as well I think (`nextNonlocalBid`)

aqjune:
I'll make a mergeable version after submission.

aqjune:
I have so many PRS opened. :/ I'll close one by one later.

aqjune:
I found that it is helpful to first gather the primitive analysis on nonlocal bids that are spread over the code into one place.
For example, we frequently check whether a bid is a global variable using some primitive range checks. For this, I defined `nonlocal_is_constglb` and let it do the check.
I added a few assertions to check the sanity of these range checks as well.

I think there are quite a few things that we can do on a fncall memory block before diving into breaking symmetry.
For example, we can assume that the memory block is always alive and is non-heap as a global variable does. Maybe we can assume that it only stores integers as well?
As a next step, I'll implement these kinds of optimizations first.

ir/memory.cpp
```diff
@@ -18,15 +18,67 @@ using namespace util;
...
   // 4. a block reserved for encoding the memory touched by calls:
   //      num_globals_src + num_ptrinputs + 1
   // 5. nonlocal blocks returned by loads/calls:
   //      num_globals_src + num_ptrinputs + 2 ~ num_nonlocals_src - 1:
   // 6. global vars in target only:
-  //      num_nonlocals_src ~ num_nonlocals - 1
+  //      num_nonlocals_src ~ num_nonlocals_src + num_extra_nonconst_tgt - 1
+  //            (non-constant globals in target only)
+  //      num_nonlocals_src + num_extra_nonconst_tgt ~ num_nonlocals - 1
+  //            (constant globals in target only)
```
aqjune, line 34: This is a non-functional change and just the addition of a description about how bids are assigned.

ir/memory.cpp
```diff
@@ -999,7 +1050,7 @@ void Memory::mkAxioms(const Memory &tgt) const {
     return;
 
   auto nonlocal_used = [&](unsigned bid) {
-    return bid < tgt.next_nonlocal_bid || bid >= num_nonlocals_src;
+    return bid < tgt.next_nonlocal_bid || nonlocal_is_globalvar(bid, true);
```
aqjune, line 1053: The exact change should be
```
return bid < tgt.next_nonlocal_bid || (!nonlocal_is_globalvar (bid, false) && nonlocal_is_globalvar(bid, true))
```
, but I optimized it.

ir/memory.cpp
```diff
@@ -623,8 +675,8 @@ bool Memory::mayalias(bool local, unsigned bid0, const expr &offset0,
                       unsigned bytes, unsigned align, bool write) const {
   if (local && bid0 >= next_local_bid)
     return false;
-  if (!local && (bid0 >= (write ? num_nonlocals_src : numNonlocals()) ||
-                 bid0 < has_null_block + write * num_consts_src))
+  if (!local && (nonlocal_always_noread(bid0) ||
```
nunoplopes, line 678: this change is not NFC
the function only does `bid < has_null_block`

ir/memory.cpp
```diff
@@ -18,15 +18,67 @@ using namespace util;
...
+  //      num_nonlocals_src ~ num_nonlocals_src + num_extra_nonconst_tgt - 1
+  //            (non-constant globals in target only)
+  //      num_nonlocals_src + num_extra_nonconst_tgt ~ num_nonlocals - 1
+  //            (constant globals in target only)
+
+//--- Functions for non-local block analysis based on bid ---//
+
+// If include_tgt is true, return true if bid is a global var existing in target
+// only as well
+static bool nonlocal_is_globalvar(unsigned bid, bool include_tgt) {
```
nunoplopes, line 40: maybe rename to is_globalvar?

ir/memory.cpp
```diff
@@ -18,15 +18,67 @@ using namespace util;
...
+  bool srcglb = has_null_block <= bid && bid < has_null_block + num_globals_src;
```
nunoplopes, line 41: the first bit is redundant

ir/memory.cpp
```diff
@@ -18,15 +18,67 @@ using namespace util;
...
+  bool tgtglb = num_nonlocals_src <= bid && bid < num_nonlocals;
+  return srcglb || (include_tgt && tgtglb);
+}
+
+static bool nonlocal_is_constglb(unsigned bid) {
```
nunoplopes, line 46: I would rename the nonlocal_ bit

ir/memory.cpp
aqjune: Oops, I fixed it

ir/memory.cpp
aqjune: I think it is good to include non-null globals only.
It allows us to reason that bid s.t. is_globalvar(bid) is alive.

aqjune: it seems global vars are assumed to be live ones: line 655 in this file had a relevant comment
```
  // globals are always live
  if (local || (bid0 >= num_globals_src && bid0 < num_nonlocals_src)) {
```

PR #655, opened 2021-01-29
fix spaceship errors in mac

aqjune:
It is raising compile error due to this issue: https://clang.llvm.org/compatibility.html#dep_lookup

Fix this error by simply moving compare_iterators back to the end & forward declaring it. Adding a forward declaration is the first solution according to the link.

Additionally, make `operator<=>(const std::string &lhs, const std::string &rhs)` visible to others because it is used by some structs in state.h

### DIFF
--- a/util/spaceship.h
+++ b/util/spaceship.h
@@ -10,16 +10,6 @@
 
 namespace {
 
-template<typename T>
-struct is_pair_helper : std::false_type {};
-
-template<typename A, typename B>
-struct is_pair_helper<std::pair<A, B>> : std::true_type {};
-
-template<typename T>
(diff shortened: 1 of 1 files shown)


aqjune:
BTW, in my case, this required upgrading CommandLineTools.
This was simply done by following https://apple.stackexchange.com/a/375535 .

aqjune:
Huh, it is failing. Looking into this..

aqjune:
(will be back after dinner)

AlexDenisov:
That's a timely patch: I just hit exactly these compile errors on mac, which were gone as soon as I applied this patch!

nunoplopes:
the buildbots seem to be finally fixed! 🥳

PR #656, opened 2021-01-29
Fix CMake binary dir discrepancy

AlexDenisov:
When used as a standalone project, both CMAKE_BINARY_DIR and
PROJECT_BINARY_DIR yield the same value.
However, when Alive2 added as a sub-project (via add_subdirectory),
then CMAKE_BINARY_DIR and PROJECT_BINARY_DIR are different, which leads
to a build failure.

This patch fixes the discrepancy between two different values.

P.S. Alternative solution would be to add `${CMAKE_BINARY_DIR}` to the
`include_directories`, but I think the suggested solution is better since it
doesn't "pollute" the build directory of the parent project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,23 +49,23 @@ endif()
 
 find_package(Git REQUIRED)
 add_custom_command(
-  OUTPUT "${CMAKE_BINARY_DIR}/version_gen.h.tmp"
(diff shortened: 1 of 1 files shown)


AlexDenisov:
Forgot the most important part: the discrepancy is coming from the fact that the `version_gen.h` is stored in `CMAKE_BINARY_DIR`, while the include directories expect it to be in the `PROJECT_BINARY_DIR`.

PR #658, opened 2021-01-31
Lazily call preprocess to shorten vcgen time

aqjune:
This is a patch to lazily call preprocess to shorten VCGen time.

### DIFF
--- a/tools/transform.cpp
+++ b/tools/transform.cpp
@@ -392,12 +392,12 @@ check_refinement(Errors &errs, Transform &t, State &src_state, State &tgt_state,
   expr pre_tgt = pre_tgt_and();
 
   expr axioms_expr = axioms();
+  pre_tgt &= !sink_tgt;
+
   if (check_expr(axioms_expr && (pre_src && pre_tgt)).isUnsat()) {
     errs.add("Precondition is always false", false);
     return;
   }
-
-  pre_tgt &= !sink_tgt;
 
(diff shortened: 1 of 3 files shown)


tools/transform.cpp
```diff
@@ -436,15 +436,15 @@ check_refinement(Errors &errs, Transform &t, State &src_state, State &tgt_state,
     return;
   }
 
-  auto mk_fml = [&](expr &&refines) -> expr {
+  auto mk_fml = [&](const expr &refines) -> expr {
```
aqjune, line 439: std::bind wasn't working well for lambda functions that have arguments with rvalue reference.
The reason is that std::bind internally forwards the argument to a variable and calls the function: https://stackoverflow.com/a/34878230

nunoplopes:
This patch made me think that we need to shake things up in that function.
We produce all formulas upfront, which increase peak memory consumption. It would be best if we produced the formulas only for the query and then destroy them straight away to reclaim the memory.

aqjune:
That really makes sense..! I'll work on this.

aqjune:
I'll see whether there is no regression in single file benchmarks.
Since the diff is slightly big, I'm fine with having it reviewed after the submission.

aqjune:
No regression in bzip2/gzip/oggenc/ph7 (sqlite3 is running)

aqjune:
no regression in sqlite3 too

PR #659, opened 2021-01-31
Load ptr's partial order reduction should consider non_poison into account

aqjune:
When load returns a pointer, the pointer's bid is to be POR-ed.

However, if the returned pointer was poison, its bid has a bogus value. The POR condition should take this into account.


aqjune:
The two examples have identical src/tgt function; it would be great if alive2test.py supports this type of test as well. I'll leave this as an issue.

nunoplopes:
FYI this patch caused a 2x slowdown in sqlite3. I didn't track down yet what happened.

aqjune:
:o 2x slowdown is quite a lot. 
Should we try to optimize encoding of poison-ness of a loaded byte? Frequently I saw that a loaded value's non_poison formula is much larger than the value formula.
I don't have an idea that immediately comes upfront though.. and I wonder whether time will permit.

PR #661, opened 2021-02-04
add option to limit number of bits used for GEP offsets

fhahn:
This patch adds a max_gep_offset config option, to limit the maximum
value assumed for GEP offsets. This effectively provides an option to
limit the width of pointer arithmetic, which in turn speeds up
verification of programs using a lot of pointer arithmetic.

For regular arithmetic, a common advice is to limit the width of integer
types, but there is no easy way to  do the same thing for pointers. The
new option allows limiting the width of pointer arithmetic globally.

This should not impact soundness & completeness, unless the input
program has pointer arithmetic that does not fit in the specified width
(e.g. has a GEP that uses an offset outside the max range). So the users
of the option should think carefully when using the option.

Perhaps there is also a way to verify that the chosen value is not too
small?

### DIFF
--- a/tests/alive-tv/ptr-arithmetic-fail.srctgt.ll
+++ b/tests/alive-tv/ptr-arithmetic-fail.srctgt.ll
@@ -1,0 +1,58 @@
+; TEST-ARGS: -max-offset-in-bits=8
+; ERROR: Value mismatch
+
+define i4 @src(i8* %src, i8* %lower, i8* %upper, i8 %N, i8 %step) {
+entry:
+  %src.end = getelementptr inbounds i8, i8* %src, i8 %N
+  %cmp.src.start = icmp ult i8* %src, %lower
+  %cmp.src.end = icmp uge i8* %src.end, %upper
+  %or.precond.0 = or i1 %cmp.src.start, %cmp.src.end
+  br i1 %or.precond.0, label %trap.bb, label %step.check
+
+trap.bb:
+  ret i4 2
+
+step.check:
+  %step.pos = icmp uge i8 %step, 0
+  %step.ult.N = icmp ult i8 %step, %N
+  %and.step = and i1 %step.pos, %step.ult.N
+  br i1 %and.step, label %ptr.check, label %exit
+
+ptr.check:
+  %src.step = getelementptr inbounds i8, i8* %src, i8 %step
+  %cmp.step.start = icmp ult i8* %src.step, %lower
+  %cmp.step.end = icmp uge i8* %src.step, %upper
+  %or.check = or i1 %cmp.step.start, %cmp.step.end
+  br i1 %or.check, label %trap.bb, label %exit
+
+exit:
+  ret i4 3
+}
+
+define i4 @tgt(i8* %src, i8* %lower, i8* %upper, i8 %N, i8 %step) {
+entry:
+  %src.end = getelementptr inbounds i8, i8* %src, i8 %N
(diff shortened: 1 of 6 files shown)


nunoplopes:
Can you share some perf numbers? Does this help that much?
This thing is a bit dangerous. You can miss bugs because of it. Hence I'm asking if the speedup is worth it 😀

regehr:
another factor is that this might make counterexamples more readable, but of course we could accomplish that goal in other ways

(I have toyed with writing a counterexample search loop that iterates until the counterexample is as simple as possible. I'm not yet sure what definition of simple makes the most sense here but it would be things like minimizing integer values and offsets, and involving as few memory blocks as possible. I feel like this sort of thing might be a really nice way to improve the accessibility of alive2's results for end users without complicating any other parts of the tool)

nunoplopes:
> another factor is that this might make counterexamples more readable, but of course we could accomplish that goal in other ways
> 
> (I have toyed with writing a counterexample search loop that iterates until the counterexample is as simple as possible. I'm not yet sure what definition of simple makes the most sense here but it would be things like minimizing integer values and offsets, and involving as few memory blocks as possible. I feel like this sort of thing might be a really nice way to improve the accessibility of alive2's results for end users without complicating any other parts of the tool)

that sounds very cool indeed. Also, try to get away without undef & poison.
I built a tiny bit of infra-structure to handle over-approximations (missing the important refinement loop). Now you are doing under-approximations for readability, which makes total sense. The refinement loop should be unified later I hope.

regehr:
if there are a number of strategies for reducing counterexamples and we put them into a fixpoint loop then we will end up with something that could be accurately characterized as C-Reduce for counterexamples :)

fhahn:
> Can you share some perf numbers? Does this help that much?

My original motivating example is https://alive2.llvm.org/ce/z/iKWwYO

On my system, that takes around ~14s without the patch and with `-max-gep-offset-in-bits=8` it takes > 1s. Maybe I am just a bit too impatient, but I also want to verify slightly larger programs and it might also be helpful to share snippets on https://alive2.llvm.org/ce/ where the timeout cannot be increased.

And the programs I am looking at generally do not have loops with large constant bounds or large explicit offsets.

> This thing is a bit dangerous. You can miss bugs because of it. Hence I'm asking if the speedup is worth it 😀

Agreed, it needs to be handled/used with care. But I think in most cases it should be relatively safe, unless the code has loops with large constant bounds or explicit large offsets. Not sure if we could guard against that by setting up some constraints?



> another factor is that this might make counterexamples more readable, but of course we could accomplish that goal in other ways

That's an interesting point! Perhaps we could try to gradually lower the gep bit width on constructing counter examples.

nunoplopes:
Thanks for the example. > 10 speedup is impressive indeed. Your example gave me an idea on how to be less conservative, but I'll take care of that later.

On this patch, let me suggest a simpler implementation. In tools/transform.cpp there's this code:
```C++
bits_for_offset = min(round_up(max_geps, 4), (uint64_t)t.src.bitsPtrOffset());
```
You can just add something like:
```C++
bits_for_offset = min(bits_for_offset, config::max_offset_bits);
```

This change is less invasive and should accomplish exactly the same.

fhahn:
> ```c++
> bits_for_offset = min(bits_for_offset, config::max_offset_bits);
> ```
> 
> This change is less invasive and should accomplish exactly the same.

Done, thanks! I also added a test and a note that using this option may impact correctness.

PR #662, opened 2021-02-04
factor out logic to compare 2 versions of functions

fhahn:
This patch moves compareFunctions & the helpers exclusively used by it
to a new file (util/compare.h). This makes it easier to integrate
translation validation into other tools.

I am planning on using it in LLVM fuzzer using LLVM's FuzzMutate, which
uses translation validation to fuzz LLVM transformations for correctness
issues. I will share a patch as a follow-up.

### DIFF
--- a/llvm_util/utils.cpp
+++ b/llvm_util/utils.cpp
@@ -4,13 +4,22 @@
 #include "llvm_util/utils.h"
 #include "ir/constant.h"
 #include "ir/function.h"
+#include "llvm_util/llvm2alive.h"
+#include "smt/smt.h"
+#include "tools/transform.h"
+#include "llvm/ADT/Triple.h"
+#include "llvm/Analysis/TargetLibraryInfo.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/DataLayout.h"
 #include "llvm/IR/Function.h"
 #include "llvm/IR/GlobalVariable.h"
 #include "llvm/IR/Module.h"
 #include "llvm/IR/Operator.h"
+#include "llvm/IR/Type.h"
 #include "llvm/Support/raw_ostream.h"
+
+#include <iostream>
+#include <sstream>
 #include <unordered_map>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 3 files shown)


regehr:
@fhahn I would be super interested to hear more about this fuzzing effort at some point

nunoplopes:
As a side comment, I've delete the code to compare LLVM function signatures. That code was just temporary until we made alive itself more robust. It's not needed anymore (famous last words).

On the technical side, you need to move compare.* to the llvm_utils directory. Compiling Alive2 with LLVM is optional, and this directory is the only one where you can place code that needs LLVM (except for the tools).

Just realized that the bidirectional logic is now wrong because of the preprocessing going on. I can fix that later to avoid more conflicts.

fhahn:
> On the technical side, you need to move compare.* to the llvm_utils directory. Compiling Alive2 with LLVM is optional, and this directory is the only one where you can place code that needs LLVM (except for the tools).

I moved it to `llvm_utils/utils.h`. Does this place make sense?

> @fhahn I would be super interested to hear more about this fuzzing effort at some point

So far, I only hooked up the alive2 function check to libFuzzer for LLVM passes. I am not sure where to best put the code, but  I put up the code in https://reviews.llvm.org/D96654, in case you have any suggestions.

nunoplopes:
Are you sure this patch is what you need?
I thought for the fuzzer you would want to know if the transformation is wrong or no. You probably don't want any of the output, especially when the test passes. Just trying to understand what's the API that would be most helpful for your use case.

nunoplopes:
I've made a few changes to alive-tv as the bidirectional mode was broken (https://github.com/AliveToolkit/alive2/commit/aaa0c7b9d0f16f71074a217605162456db41453a). Plus a lot of refactoring on argument handling & so on.
This PR now has a lot of conflicts. But it's for a good cause. The refactored code should be in much better shape for reuse by other tools. So I'm closing this PR (as it essentially needs to be rewritten), but I'm happy to help extracting the relevant functionality for the fuzzer. The new `verify` function should be all you need. (I guess you are now waiting for the GSoC student)

fhahn:
> So I'm closing this PR (as it essentially needs to be rewritten), but I'm happy to help extracting the relevant functionality for the fuzzer. The new verify function should be all you need. (I guess you are now waiting for the GSoC student)

Thanks that sounds good to me! I'm indeed waiting to see if this gets picked up during GSoC or now.

Issue #643, opened 2021-01-04
State::getAndAddPoisonUB() doesn't simplify undef pointers


nunoplopes:
So I don't forget:
```llvm
%struct.va_list = type { i8* }

define i32 @src(%struct.va_list* %p) {
  %p2 = bitcast %struct.va_list* %p to i8*
  %a = va_arg i8* %p2, i32
  ret i32 %a
}
```

```
$ ./alive-tv -se-verbose aa.ll

%a = (va_arg_in_i32 (ite (= #b0 isundef_%p) %p (concat #b0 undef!2)) #x00) / ...
```

Issue #27, opened 2019-02-06
Simple transform times out


regehr:
Getting this with latest alive2 on OS X.

```
Johns-MacBook-Pro:build johnregehr$ ./alive foo2.opt
Processing foo2.opt..

----------------------------------------
  %a = add i64 %b, %b
=>
  %a = shl %b, 1

ERROR: Timeout

Johns-MacBook-Pro:build johnregehr$ ./alive foo3.opt
Processing foo3.opt..

----------------------------------------
  %a = add i65 %b, %b
=>
  %a = shl %b, 1

Doesn't type check!
Johns-MacBook-Pro:build johnregehr$ 
```

nunoplopes:
Ah, for types > 64 bits, you need to fully specify the types everywhere (Alive1 had the same limitation).
So this should work:
```
  %a = add i65 %b, %b
=>
  %a = shl i65 %b, 1
```

For the timeout, I need to investigate. It should be canonicalized anyway.

regehr:
ah, makes sense, thanks!

I think I never tried larger widths with Alive1, and only really care about them here since Souper sometimes sees them

manasij7479:
Current Alive2 is working fairly well with Souper, I am seeing less timeouts in general. Thanks!
(This particular example still seems to time out with i13 and above.)

regehr:
Now that alive2 is pretty mature and capable, I'd like to return and maybe spend some time working on performance issues. I have some time for this over the summer. For example the query below is one that I would naively expect to be fast, but it times out on volta, a very fast and otherwise idle machine. 
http://volta.cs.utah.edu:8080/z/CLAnvb

I haven't looked into this yet, but I assume that a quantifier is making it to bit blasting? Nuno, I know that you're miles ahead of me on this stuff-- do you have a bit of time to discuss what might be the best places to start attacking these issues?

```
define i16 @src(i16 %x) {
    %r = add i16 %x, %x
    ret i16 %r
}

define i16 @tgt(i16 %x) {
    %r = shl i16 %x, 1
    ret i16 %r
}
```

nunoplopes:
Right, that examples gets a quantifier due to the undef values.
It should be easy yes. The quantifier just needs to guess that the 3 undefs need to be equal to prove it correct. For some reason, Z3's MBQI algorithm is failing to find that instantiation.
[...]

PR #664, opened 2021-02-08
refactoring to undef-related functions

aqjune:
This is a non-functional change to undef-related functions, mainly changing function names or factoring out code to a separate function for readability.
I'll leave comments at the places where an explanation is necessary

I think this will help writing experimental codes for #27 and #643 .

### DIFF
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -146,48 +146,38 @@ static expr eq_except_padding(const Type &ty, const expr &e1, const expr &e2) {
   return result;
 }
 
-expr State::strip_undef_and_add_ub(const Value &val, const expr &e) {
-  if (isUndef(e)) {
+expr State::stripUndefAndAddUB(const Type &valty, const expr &e) {
+  if (isUndefVarInCurrentExec(e)) {
     addUB(expr(false));
     return expr::mkUInt(0, e);
   }
 
-  auto is_undef_cond = [](const expr &e, const expr &var) {
-    expr lhs, rhs;
-    // (= #b0 isundef_%var)
-    if (e.isEq(lhs, rhs)) {
-      return (lhs.isZero() && Input::isUndefMask(rhs, var)) ||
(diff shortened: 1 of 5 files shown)


ir/state.cpp
```diff
@@ -146,48 +146,38 @@ static expr eq_except_padding(const Type &ty, const expr &e1, const expr &e2) {
   return result;
 }
 
-expr State::strip_undef_and_add_ub(const Value &val, const expr &e) {
-  if (isUndef(e)) {
```
aqjune, line 150: I changed the name of this function to avoid ambiguity because there was another isUndef() at globals.h.

ir/state.cpp
```diff
@@ -199,15 +189,15 @@ expr State::strip_undef_and_add_ub(const Value &val, const expr &e) {
 
   // two variants
   // 1) boolean
-  if (is_if_undef(e, a, b)) {
+  if (is_input(e, a, b)) {
     addUB(move(b));
     return a;
   }
 
-  auto has_undef = [&](const expr &e) {
```
aqjune, line 207: All uses of has_undef are negating the result, so I changed this to well_defined.

ir/state.cpp
```diff
@@ -146,48 +146,38 @@ static expr eq_except_padding(const Type &ty, const expr &e1, const expr &e2) {
...
-    if (e.isEq(lhs, rhs)) {
-      return (lhs.isZero() && Input::isUndefMask(rhs, var)) ||
-             (rhs.isZero() && Input::isUndefMask(lhs, var));
-    }
-    return false;
-  };
-
-  auto is_if_undef = [&](const expr &e, expr &var, expr &not_undef) {
-    expr undef;
-    // (ite (= #b0 isundef_%var) %var undef)
```
aqjune, line 167: Pattern matching this formula is factored out to `Input::match`.

PR #665, opened 2021-02-08
Add .ident.ll test format

### LINKED ISSUE #660: Add a test format for identical src/tgt
There are quite a few alive2 tests that have src/tgt functions that are syntactically identical - it would be great if alive2test.py supports this kind of format.

aqjune:
This patch adds .ident.ll test format and updates tests that have identical src/tgt pairs to use this format.

To detect the list of tests that have identical pairs, I write this script, placed it at tests/alive-tv, and run it: https://gist.github.com/aqjune/85b696292fd4f1685a3f71f7d8e3390d

The output: https://gist.github.com/aqjune/7d03c926715dad4c498312d9ba2887e1

Interestingly this detected quite a few ill-formed .srctgt tests.
They had only one of `@src()`, `@tgt()`, or none of these.
I fixed these to be well-formed as well.

Resolves #660


tests/alive-tv/attrs/dereferenceable-shrink.srctgt.ll
```diff
@@ -6,6 +6,6 @@ define void @src(i8* %p) {
 }
 
 define void @tgt(i8* %p) {
-  call void @f(i8* dereferenceable(4) %p)
```
aqjune, line 9: This test wasn't shrinking the dereferenceable size (which isn't what the file name is saying).
alive2test.py is already running identity tests for .srctgt tests, so I fixed this test to shrink this.

tests/lit/lit/formats/alive2test.py
```diff
@@ -129,6 +126,8 @@ def execute(self, test, litConfig):
     cmd.append(test)
     if alive_tv_2:
       cmd.append(test.replace('.src.ll', '.tgt.ll'))
+    elif alive_tv_3:
+      cmd = cmd + [test, '-always-verify']
```
nunoplopes, line 130: uhm, doesn't this need to be `[test, test, --]` ?

aqjune: The other test is added by line 126.

PR #669, opened 2021-02-15
fix two things that block compilation using latest g++

regehr:
not totally sure about the second one...


nunoplopes:
the 2nd one is a more fundamental issue. will try to fix it later today or tmr.

Issue #666, opened 2021-02-13
Load/store's pointers are noundef


aqjune:
This patch has landed: https://reviews.llvm.org/D95238

aqjune:
I'll work on this tomorrow

aqjune:
Running LLVM unit tests with Alive2 is failing, and I guess this is due to LLVM's recent switch to enabling new pass manager by default. :/
I found that passing `-load-pass-plugin=...` to opt by default resolves this issue, at least in a few litmus tests. Currently, it is passed only when `$NPM_TV` is 1 (at opt-alive.sh).

aqjune:
I'll look into this Monday

PR #671, opened 2021-02-17
Resolve unit test failures after LLVM's switch to NPM

aqjune:
This is a patch to resolve the failures on my machine after NPM switch is enabled (https://github.com/AliveToolkit/alive2/issues/666#issuecomment-778779487).
After the NPM switch, all pass names in opt's command line arguments are parsed and passed into NPM (unless `-enable-new-pm=0` is given). For example, running `opt -tv -instcombine -tv` lookups `tv` and `instcombine` new passes.

This again causes a nasty problem that happened before (https://github.com/AliveToolkit/alive2/commit/c6d4b33daa177d64ab4b5947253dedd7dd61ac65 ). 
opt calls alive2's `registerPipelineParsingCallback("tv")` with dummy extra arguments to investigate whether `-tv` is a module-level pass or not.
This is problematic because our `registerPipelineParsingCallback("tv")` is not *idempotent*.
It adds `TVInitPass` in the first `registerPipelineParsingCallback` call only; the second call won't register TVInitPass.
Also, each call increments `TVNewPass::num_tv_pass_created++`, which is bad if it was a dummy call.

To resolve this, this patch makes `registerPipelineParsingCallback` idempotent.
To do this,

(1) I removed `TVInitPass`, and let simply the first run() of TV pass do the initialization.

(2) I made the TV pass instance do the reference counting at its constructor and destructor. If the number of references becomes zero, finalization is done.

Note that 
- This patch should not affect tests with opt + legacy pass manager.
- This patch affects clang tv because `TVInitPass` is removed; Now it starts validation earlier than before! In the past, there were a few passes like ForceFunctionAttr run before TVInitPass, and they couldn't be checked.

### DIFF
--- a/tests/opt-tv/npm-Os.opt.ll
+++ b/tests/opt-tv/npm-Os.opt.ll
@@ -5,4 +5,4 @@ define i32 @f() {
 }
 
 ; src and tgt are identical, but opt-alive should not validate anything
-; CHECK-NOT: Transformation
+; CHECK-NOT: Transformation seems
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -241,13 +241,13 @@ string get_random_str() {
   return to_string(rand(re));
 }
 
-struct TVPass final : public llvm::ModulePass {
+struct TVLegacyPass final : public llvm::ModulePass {
   static char ID;
   bool skip_verify = false;
   const function<llvm::TargetLibraryInfo*(llvm::Function&)> *TLI_override
     = nullptr;
 
-  TVPass() : ModulePass(ID) {}
+  TVLegacyPass() : ModulePass(ID) {}
 
   bool runOnModule(llvm::Module &M) override {
     for (auto &F: M)
@@ -402,8 +402,13 @@ struct TVPass final : public llvm::ModulePass {
   }
 
   bool doInitialization(llvm::Module &module) override {
+    initialize(module);
+    return false;
+  }
+
+  static void initialize(llvm::Module &module) {
     if (initialized++)
-      return false;
+      return;
 
     fnsToVerify.insert(opt_funcs.begin(), opt_funcs.end());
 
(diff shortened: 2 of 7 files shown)


tests/opt-tv/npm-Os.opt.ll
```diff
@@ -5,4 +5,4 @@ define i32 @f() {
 }
 
 ; src and tgt are identical, but opt-alive should not validate anything
-; CHECK-NOT: Transformation
+; CHECK-NOT: Transformation seems
```
aqjune, line 8: This was needed because it matched `WarnMissedTransformationsPass` in LLVM.

tv/tv.cpp
```diff
@@ -654,21 +658,10 @@ struct TVNewPass : public llvm::PassInfoMixin<TVNewPass> {
 
   void run(llvm::Module &M,
            const function<llvm::TargetLibraryInfo*(llvm::Function&)> &get_TLI) {
-    // Check TVFinalizePass::finalized again because TVNewPass can be
-    // called after TVFinalizePass registered by
-    // registerOptimizerLastEPCallback.
-    // It can happen when a default pipeline (O3, O2, ...) is used.
-    if (TVFinalizePass::finalized)
-      return;
```
aqjune, line 662: In the past, I found that this case never happened. I'll recheck whether it indeed is by running whole LLVM unit tests.

aqjune:
Before getting this reviewed, can someone check whether failures are reproducible with the recentmost LLVM commits..?
On my machine, simply running `ninja check` from alive2 was enough.

nunoplopes:
Alive2 & LLVM from this morning works for me.
But this patch seems an improvement anyway. It makes clang verify more, right?

aqjune:
Hmm, that's weird..
Yes, it allows a few very first passes to be verified. :)

PR #672, opened 2021-02-17
Resolve a crash when tv.so loaded by opt without tv

aqjune:
My previous PR (#671) introduced regressions in a few LLVM unit tests because my opt-alive updates caused interprocedural transformations to be verified. :(

The problem lies in here:
```
    # opt enables NPM by default.		
     if [[ $arg == "-enable-new-pm=0" ]]; then
       NPM_TV=0
     else		
       NPM_TV=1 << this is problematic		
     fi
```
Even if `$arg` was something like `-disable-output`, NPM_TV was again set to 1, causing tv.so to be (almost always) loaded and initiating the translation validation related things.

The added `opt-tv/noglobalopt.opt.ll` was failing before.
Another test `opt-tv/noglobalopt-legacypm-style.opt.ll` failed after my first version of the fix, so I added this.


aqjune:
Okay, ready to get reviewed.

PR #674, opened 2021-02-18
Dereferenced pointers are noundef

### LINKED ISSUE #666: Load/store's pointers are noundef

aqjune:
This PR resolves #666.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -3061,7 +3061,7 @@ void Load::print(ostream &os) const {
 }
 
 StateValue Load::toSMT(State &s) const {
-  auto &p = s.getAndAddPoisonUB(*ptr).value;
+  auto &p = s.getAndAddPoisonUB(*ptr, true).value;
   check_can_load(s, p);
   auto [sv, ub] = s.getMemory().load(p, getType(), align);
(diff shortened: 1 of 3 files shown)


ir/instr.cpp
```diff
@@ -3159,15 +3159,17 @@ void Memset::print(ostream &os) const {
...
   expr vptr;
   if (vbytes.isUInt(n) && n > 0) {
-    vptr = s.getAndAddPoisonUB(*ptr).value;
+    vptr = s.getAndAddPoisonUB(*ptr, true).value;
   } else {
     auto &sv_ptr = s[*ptr];
-    s.addUB((vbytes != 0).implies(sv_ptr.non_poison));
+    auto &sv_ptr2 = s[*ptr];
+    s.addUB((vbytes != 0).implies(
+        sv_ptr.non_poison && (sv_ptr.value == sv_ptr2.value)));
```
aqjune, line 3172: This naively encodes non-undefness, but I think we need to encapsulate this in a new function in State or somewhere that returns (value, non-undefness).
I can work on this if you want.

PR #676, opened 2021-02-20
tv shouldn't run if NPM is disabled or has tailcallelim

aqjune:
This resolves two regressions: Transforms/TailCallElim/opt-remarks-recursion.ll and Transforms/LoopUnrollAndJam/unroll-and-jam.ll .

opt-remarks-recursion.ll: tailcallelim is a function-level pass but unsupported by Alive2. I added a check for this.
unroll-and-jam.ll: it has tbaa enabled but still `module(tv)` was being added to -passes. We should check whether NPM is already set to 0 when adding `module(tv)`.

### DIFF
--- a/scripts/opt-alive.sh.in
+++ b/scripts/opt-alive.sh.in
@@ -33,8 +33,10 @@ for arg in $@; do
     passes=${arg/-*passes=/}
     passes_new=`@CMAKE_SOURCE_DIR@/scripts/rewritepass.py "$PASSREGISTRY" $passes`
     firstpass_level=`echo $passes_new | cut -d "(" -f1`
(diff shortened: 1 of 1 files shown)


aqjune:
In total, this PR resolves:
```
Transforms/LICM/promote-order.ll
Transforms/LICM/promote-tls.ll             
Transforms/LICM/scalar-promote-memmodel.ll 
Transforms/LICM/scalar-promote.ll          
Transforms/LoopUnrollAndJam/pragma-explicit.ll  
Transforms/LoopUnrollAndJam/unroll-and-jam.ll   
Transforms/TailCallElim/opt-remarks-recursion.ll
```

Issue #560, opened 2020-11-05
Refactor LLVM -> Alive2 attributes mapping


aqjune:
The code (which is at llvm2alive.cpp) isn't very beautiful now.

It is likely that this will involve adding a few functions to LLVM side and using them.

aqjune:
https://reviews.llvm.org/D97454

aqjune:
While making a bug fix patch for the align assume operand bundle issue, I found that it is possible to neaten llvm2alive.cpp without relying on the patch above. I'll make PR soon

PR #679, opened 2021-02-25
Add support for dereferenceable_or_null

aqjune:
Resolves a regression in Transforms/SimplifyCFG/UnreachableEliminate.ll

I'll leave the LLVM unit test result here when it's done

### DIFF
--- a/ir/attrs.h
+++ b/ir/attrs.h
@@ -14,11 +14,12 @@ public:
   enum Attribute { None = 0, NonNull = 1<<0, ByVal = 1<<1, NoCapture = 1<<2,
                    NoRead = 1<<3, NoWrite = 1<<4, Dereferenceable = 1<<5,
                    NoUndef = 1<<6, Align = 1<<7, Returned = 1<<8,
-                   NoAlias = 1<<9 };
+                   NoAlias = 1<<9, DereferenceableOrNull = 1<<10 };
 
   ParamAttrs(unsigned bits = None) : bits(bits) {}
 
   uint64_t derefBytes; // Dereferenceable
+  uint64_t derefOrNullBytes; // DereferenceableOrNull
   unsigned blockSize;  // exact block size for e.g. byval args
   unsigned align = 1;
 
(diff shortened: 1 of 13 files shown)


ir/attrs.h
```diff
@@ -46,14 +48,16 @@ class FnAttrs final {
...
+                   NoThrow = 1 << 10, NoAlias = 1 << 11, WillReturn = 1 << 12,
+                   DereferenceableOrNull = 1 << 13 };
 
   FnAttrs(unsigned bits = None) : bits(bits) {}
 
   bool has(Attribute a) const { return (bits & a) != 0; }
   void set(Attribute a) { bits |= (unsigned)a; }
 
   uint64_t derefBytes; // Dereferenceable
+  uint64_t derefOrNullBytes; // DereferenceableOrNull
```
aqjune, line 60: This has to be a separate field to precisely encode the semantics of dereferenceable_or_null.
ex) `define i8* dereferenceable(1) dereferenceable_or_null(2) @f()`

llvm_util/llvm2alive.cpp
```diff
@@ -337,6 +343,20 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
+             .hasParamAttr(argidx, llvm::Attribute::DereferenceableOrNull))
+          sz = i.getParamAttr(argidx, llvm::Attribute::DereferenceableOrNull)
+                .getDereferenceableOrNullBytes();
+        if (argidx < fn->arg_size())
+          sz = max(sz, fn->getParamDereferenceableOrNullBytes(argidx));
+        assert(sz);
+        attr.set(ParamAttrs::DereferenceableOrNull);
+        attr.derefOrNullBytes = sz;
+      }
+
```
aqjune, line 359: This should be simpler; issue: #560
I'm thinking of an API update in LLVM. If it is done, I'll make a patch here too.

aqjune:

PR #683, opened 2021-02-26
refactor code duplication in fn/paramattr encoding

aqjune:
This patch goes one step further and refactors code duplication in FnAttrs <-> ParamAttrs.

### DIFF
--- a/ir/attrs.cpp
+++ b/ir/attrs.cpp
@@ -9,6 +9,7 @@
 #include <cassert>
 
 using namespace std;
+using namespace smt;
 
 namespace IR {
 ostream& operator<<(ostream &os, const ParamAttrs &attr) {
@@ -87,6 +88,51 @@ uint64_t ParamAttrs::getDerefBytes() const {
   return bytes;
 }
 
+static pair<AndExpr, expr>
+encodePtrAttrs(const State &s, const expr &ptrvalue, uint64_t derefBytes,
+               uint64_t derefOrNullBytes, uint64_t align, bool nonnull) {
+  Pointer p(s.getMemory(), ptrvalue);
+  AndExpr UB;
+  expr non_poison = true;
+
+  if (nonnull)
+    non_poison &= p.isNonZero();
+
+  if (derefBytes || derefOrNullBytes) {
+    // dereferenceable, byval (ParamAttrs), dereferenceable_or_null
+    if (derefBytes)
+      UB.add(p.isDereferenceable(derefBytes, align));
+    if (derefOrNullBytes)
+      UB.add(p.isDereferenceable(derefOrNullBytes, align)() || !p.isNonZero());
+  } else if (align != 1)
+    // align
+    non_poison &= p.isAligned(align);
+
+  return make_pair(move(UB), move(non_poison));
+}
+
+pair<AndExpr, expr>
+ParamAttrs::encode(const State &s, const StateValue &val, const Type &ty) const {
+  AndExpr UB;
+  expr new_non_poison = val.non_poison;
+
+  if (ty.isPtrType()) {
+    auto [UBptr, npptr] =
+        encodePtrAttrs(s, val.value, getDerefBytes(), derefOrNullBytes, align,
+                       has(NonNull));
+    UB.add(move(UBptr));
+    new_non_poison &= move(npptr);
+  }
+
+  if (poisonImpliesUB()) {
+    UB.add(move(new_non_poison));
+    new_non_poison = true;
+  }
+
+  return make_pair(move(UB), move(new_non_poison));
+}
+
+
 bool FnAttrs::poisonImpliesUB() const {
   return has(Dereferenceable) || has(NoUndef) || has(NNaN) ||
          has(DereferenceableOrNull);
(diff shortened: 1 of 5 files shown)


ir/attrs.h
```diff
@@ -18,8 +22,8 @@ class ParamAttrs final {
 
   ParamAttrs(unsigned bits = None) : bits(bits) {}
 
-  uint64_t derefBytes; // Dereferenceable
-  uint64_t derefOrNullBytes; // DereferenceableOrNull
+  uint64_t derefBytes = 0; // Dereferenceable
+  uint64_t derefOrNullBytes = 0; // DereferenceableOrNull
```
aqjune, line 26: dereferenceable and dereferenceable_or_null cannot be zero (https://godbolt.org/z/WWdEs7 , https://godbolt.org/z/f8Kfz9), so it is sound to have zero as a value to indicate that there's no 'dereferenceable' attr.

ir/attrs.h
```diff
@@ -69,4 +73,12 @@ class FnAttrs final {
   friend std::ostream& operator<<(std::ostream &os, const FnAttrs &attr);
 };
 
+
+// Encode attributes for a non-aggregate value using UB and poison.
+// Attributes requiring more than UB or poison are not encoded.
+void encodeParamAttrs(const ParamAttrs &attrs, State &s, StateValue &value,
```
nunoplopes, line 79: can we move these to the respective classes?

TBH, I don't really like these functions changing StateValue directly. It's very non-obvious. Can we return a pair with the UB (AndExpr) and poison (maybe expr is enough)?

ir/attrs.h
```diff
@@ -3,10 +3,15 @@
 // Copyright (c) 2018-present The Alive2 Authors.
 // Distributed under the MIT license that can be found in the LICENSE file.
 
+#include "smt/exprs.h"
 #include <ostream>
 
 namespace IR {
 
+class State;
```
nunoplopes, line 11: state can be dropped now

ir/attrs.cpp
```diff
@@ -93,4 +143,31 @@ bool FnAttrs::undefImpliesUB() const {
...
+    UB.add(move(UBptr));
+    new_non_poison &= move(npptr);
+  }
+
+  if (poisonImpliesUB()) {
+    UB.add(move(new_non_poison));
+    new_non_poison = true;
+  }
+
+  return make_pair(move(UB), move(new_non_poison));
```
nunoplopes, line 170: `return { ... }` instead of make_pair

ir/value.cpp
```diff
@@ -218,38 +218,15 @@ StateValue Input::mkInput(State &s, const Type &ty, unsigned child) const {
...
-                !p.isNonZero());
-    } else if (hasAttribute(ParamAttrs::Align))
-      np_from_attr &= p.isAligned(attrs.align);
-  }
-
-  if (attrs.poisonImpliesUB()) {
-    s.addUB(move(np_from_attr));
-    np_from_attr = true;
-  }
+  auto [UB, non_poison] = attrs.encode(s, {expr(val), expr(true)}, ty);
```
nunoplopes, line 221: the non_poison part should take the `never_poison ? true : expr::mkBoolVar(np_name.c_str())` bit from below I guess

ir/attrs.cpp
```diff
@@ -83,6 +88,51 @@ uint64_t ParamAttrs::getDerefBytes() const {
   return bytes;
 }
 
+static pair<AndExpr, expr>
+encodePtrAttrs(const State &s, const expr &ptrvalue, uint64_t derefBytes,
```
nunoplopes, line 92: this one is an internal function. I would pass the UB/poison vars as arguments.

ir/attrs.h
aqjune: On my machine, it raised compile error due to the State argument in encode

ir/value.cpp
aqjune: never_poison considers `attrs.poisonImpliesUB()` into account, so in terms of correctness, it is fine.
I think this form is better because we can optimize away creation of `np_..` SMT variables. Otherwise attrs.encode will call `addUB(np_...)`.

aqjune: Actually this code becomes clearer if the non_poison part takes `never_poison ? true : expr::mkBoolVar(..)` as suggested.  If the performance cost of having the extra variable is almost negligible, I'll change this.

ir/attrs.h
aqjune: Oops, if the question was that State can be dropped from encode's argument, it is still necessary because encoding pointer attributes requires Memory.

ir/attrs.cpp
```diff
@@ -83,6 +88,45 @@ uint64_t ParamAttrs::getDerefBytes() const {
...
+
+  if (derefBytes || derefOrNullBytes) {
+    // dereferenceable, byval (ParamAttrs), dereferenceable_or_null
+    if (derefBytes)
+      UB.add(p.isDereferenceable(derefBytes, align));
+    if (derefOrNullBytes)
+      UB.add(p.isDereferenceable(derefOrNullBytes, align)() || !p.isNonZero());
+  } else if (align != 1)
+    // align
+    non_poison &= p.isAligned(align);
```
nunoplopes, line 109: this has to be UB.
a load from an unaligned pointer can crash some CPUs.

PR #686, opened 2021-03-05
Fix lifetime encoding to match the new langref wording

### LINKED ISSUE #433: GVN transformation regarding lifetime.start
https://godbolt.org/z/8D-TvV

```
define i8 @src() {
entry:
    %p = alloca i8
    call void @llvm.lifetime.start.p0i8(i64 1, i8* %p)
    store i8 1, i8* %p
    call void @llvm.lifetime.start.p0i8(i64 1, i8* %p)
    %v = load i8, i8* %p
    call void @llvm.lifetime.end.p0i8(i64 1, i8* %p)
    ret i8 %v
}

define i8 @tgt() {
  %p = alloca i8, align 1
  call void @llvm.lifetime.start.p0i8(i64 1, i8* %p)
  store i8 1, i8* %p, align 1
  call void @llvm.lifetime.start.p0i8(i64 1, i8* %p)
  call void @llvm.lifetime.end.p0i8(i64 1, i8* %p)
  ret i8 undef
}
```

LangRef says `A load from the pointer that precedes this intrinsic can be replaced with 'undef'.`

aqjune:
Fixes #433 

Conservatively, this patch still raises unsupported feature error if lifetime's pointer has alloca/non-alloca objects mixed.


RalfJung:
So how does this handle the acausal nature of `alloca`+`lifetime.start`, where the behavior of `alloca` depends on which operations are going to be performed in the future of the execution?

aqjune:
Hmm, I think we should raise unsupported instruction if the program contains a 'bad-looking' lifetimes. A lifetime call is good-looking if it exactly determines which alloca must be dead in the beginning.
I just realized that Alive2 was already raising the unsupported error for such lifetime intrinsic in the past, so it is consistent with Alive2's current policy as well.

ir/instr.h
```diff
@@ -475,16 +475,15 @@ class MemInstr : public Instr {
...
-        bool initially_dead)
-    : MemInstr(type, std::move(name)), size(&size), mul(mul), align(align),
-      initially_dead(initially_dead) {}
+  Alloc(Type &type, std::string &&name, Value &size, Value *mul, unsigned align)
+    : MemInstr(type, std::move(name)), size(&size), mul(mul), align(align) {}
 
   Value& getSize() const { return *size; }
   Value* getMul() const { return mul; }
   bool initDead() const { return initially_dead; }
+  void markAsInitiallyDead() { initially_dead = true; }
```
aqjune, line 486: Previously llvm2alive.cpp's lifetime.star/end encoding was splitted into two functions: hasLifetimeStart (which checks AllocaInst's uses) and visitIntrinsicInst (which adds StartLifetime instance)
I removed hasLifetimeStart, and made visitIntrinsicInst to update this initially-dead flag if lifetime.start uses the alloca.
This `markAsInitiallyDead` function is added to implement it.

ir/memory.cpp
```diff
@@ -1658,6 +1658,13 @@ void Memory::copy(const Pointer &src, const Pointer &dst) {
   dst_blk.val = *val();
 }
 
+void Memory::fillPoison(const expr &bid) {
+  Pointer p(*this, bid, expr::mkUInt(0, bits_for_offset));
+  expr blksz = p.blockSize();
+  memset(p.release(), IntType("i8", 8).getDummyValue(false),
+         move(blksz), bits_byte / 8, {}, false);
```
aqjune, line 1665: I guess this encoding can be optimized, but fillPoison() is called only when there are multiple lifetime.start/ends, or they are taking non-stack-allocated objects, etc. These cases are rare.

llvm_util/llvm2alive.cpp
```diff
@@ -579,12 +566,12 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
 
-    unordered_set<llvm::Value*> visited;
     // FIXME: size bits shouldn't be a constant
     auto size = make_intconst(DL().getTypeAllocSize(i.getAllocatedType()), 64);
-    RETURN_IDENTIFIER(make_unique<Alloc>(*ty, value_name(i), *size, mul,
-                        pref_alignment(i, i.getAllocatedType()),
-                        hasLifetimeStart(i, visited)));
+    auto alloc = make_unique<Alloc>(*ty, value_name(i), *size, mul,
+                      pref_alignment(i, i.getAllocatedType()));
+    allocs.emplace(&i, make_pair(alloc.get(), /*has lifetime.start?*/ false));
```
aqjune, line 573: This flag will be turned on by `getLifetimeKind` if it finds lifetime.start with a pointer to this alloca.

tests/alive-tv/lifetime/lifetime-noub.srctgt.ll
```diff
@@ -12,3 +23,4 @@ define void @f() {
   unreachable
 }
 
+; ERROR: Source is more defined than target
```
aqjune, line 26: This is a non-functional change; just merges lifetime-noub.src.ll/.tgt.ll into lifetime-noub.srctgt.ll

aqjune:
Now is the time to run LLVM unit tests.. It takes about 3 hrs, but if I don't forget I'll leave the result here when it's done.

aqjune:
No regression from LLVM unit tests

PR #687, opened 2021-03-05
fix build against latest llvm

regehr:


regehr:
oh no, I'm sorry-- I seem to have had an LLVM version problem on one of my machines, please revert this

PR #691, opened 2021-03-08
support lifetime undef or heap alloc

aqjune:
This patch resolves unsupported lifetime errors from:
```
InstCombine/deadcode.ll
InstCombine/malloc-free-delete.ll
MemCpyOpt/memset-memcpy-oversized.ll
DeadStoreElimination/MSSA/multiblock-multipath.ll
NewGVN/verify-memoryphi.ll
```

LoopVectorize/lifetime.ll still has unsupported lifetimes, but they look weird: https://godbolt.org/z/vf39f3
The input function already isn't looking good; lifetime.start/end is spread over entry and exit block.
The output function either has double lifetime.start or the pointer operand has `extract(...)` form.
I think this is a suboptimal form, and to help stackcoloring it should be fixed from LLVM side.

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -6,6 +6,7 @@
 #include "llvm_util/utils.h"
 #include "util/sort.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/Analysis/MemoryBuiltins.h"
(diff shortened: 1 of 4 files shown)


aqjune:
For lifetime(undef): https://github.com/llvm/llvm-project/commit/99890a225f80495468174becb876a560f4383cea has a relevant patch (which is old).

PR #693, opened 2021-03-08
refactor LLVM -> Alive2 attributes mapping

### LINKED ISSUE #560: Refactor LLVM -> Alive2 attributes mapping

aqjune:
This PR resolves #560 .

Running LLVM unit tests...

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -266,8 +266,7 @@ public:
     }
 
     auto call = make_unique<FnCall>(*ty, value_name(i),
-                                    '@' + fn->getName().str(), move(attrs),
-                                    approx);
+                                    '@' + fn->getName().str(), move(attrs));
     unique_ptr<Instr> ret_val;
 
     for (uint64_t argidx = 0, nargs = i.arg_size(); argidx < nargs; ++argidx) {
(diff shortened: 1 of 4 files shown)


llvm_util/llvm2alive.cpp
```diff
@@ -1213,13 +1115,89 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
         continue;
 
       default:
-        errorAttr(attr);
-        return nullopt;
+        // If it is call site, don't raise unsupported attribute error
+        if (!is_callsite) {
```
nunoplopes, line 1119: maybe let's change this so it never errors out.
Fncall should be tagged as approximated when an unknown attribute is used.

aqjune:
No regression in # of failures modulo approximation; approx part is added after the attrs.derefOrNullBytes fix

llvm_util/llvm2alive.cpp
aqjune: I updated this part, but the approximation reason should be passed?
Since I'm not familiar with the approximation part, I made the approx flag be passed and that's it.

aqjune:
I'm running LLVM unit tests with the approximation code included; will leave a comment here if any failure is found

llvm_util/llvm2alive.cpp
nunoplopes: the approximation code allow us to give a reason, yes. But the fncall instruction doesn't store a reason.
I think your code looks good as it is 😉

aqjune:
There's no new failure in LLVM unit tests on my machine

PR #694, opened 2021-03-11
add support for align assume op bundle with the extra param

aqjune:
This adds support for align assume op bundle with the extra parameter.

According to https://reviews.llvm.org/D90529#2613303, align with an extra param is used to model gcc's __builtin_assume_aligned.
align's semantics:
```
... with "align"(i8* ptr, i64 16, i64 12), ptr is aligned on 4 but ptr + 4 is aligned on 16.
```
__builtin_assume_aligned (from https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html):
```
void *x = __builtin_assume_aligned (arg, 32, 8);
means that the compiler can assume for x, set to arg, that (char *) x - 8 is 32-byte aligned.
```

It seems these two are equivalent, so I chose the 'subtraction' version (the gcc one) to encode the semantics.

I'll leave the LLVM unit tests result here when it's done.


llvm_util/llvm2alive.cpp
```diff
@@ -712,14 +713,35 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
-            // TODO: "align" assume operand bundle may take a third argument.
-            return error(i);
 
           auto *aptr = get_operand(ptr), *aalign = get_operand(align);
           if (!aptr || !aalign)
             return error(i);
 
+          if (bundle.Inputs.size() >= 3) {
+            assert(bundle.Inputs.size() == 3);
+            auto *adjustofs = get_operand(bundle.Inputs[2].get());
```
nunoplopes, line 723: This is probably guaranteed to be a constant always

LebedevRI:
I'm not sure off-hand if `add`/`sub` are equivalent there,
but `sub` is how LLVM originally lowered it:
https://godbolt.org/z/1q3c1E
So semantically LGTM.

llvm_util/llvm2alive.cpp
aqjune: I think so, I will fix this soon

aqjune:
There was a regression from Transforms/AlignmentFromAssumptions/simple.ll, and I think it conflicts with the definition that I wrote. The regression looks like this:
```
 28 define i32 @foo2a(i32* nocapture %a) nounwind uwtable readonly {
 29 entry:
 30   tail call void @llvm.assume(i1 true) ["align"(i32* %a, i32 32, i32 28)]
 31   %arrayidx = getelementptr inbounds i32, i32* %a, i64 -1
 32   %0 = load i32, i32* %arrayidx, align 4
 33   ret i32 %0
 34 
 35 ; CHECK-LABEL: @foo2a
 36 ; CHECK: load i32, i32* {{[^,]+}}, align 32
 37 ; CHECK: ret i32
 38 }
```

According to what was said before, `"align"(i32* %a, i32 32, i32 28)` means `%a + 4` is 32-bytes aligned.
Then, `%a - 4` cannot be 32-bytes aligned, but this test checks it.
Maybe I can ask people there.

llvm_util/llvm2alive.cpp
aqjune: I found that using `get_operand` was simpler... Instead I left an assertion there to make the understanding about align explicit .

nunoplopes:
> There was a regression from Transforms/AlignmentFromAssumptions/simple.ll, and I think it conflicts with the definition that I wrote. The regression looks like this:
> 
> ```
>  28 define i32 @foo2a(i32* nocapture %a) nounwind uwtable readonly {
>  29 entry:
>  30   tail call void @llvm.assume(i1 true) ["align"(i32* %a, i32 32, i32 28)]
>  31   %arrayidx = getelementptr inbounds i32, i32* %a, i64 -1
>  32   %0 = load i32, i32* %arrayidx, align 4
>  33   ret i32 %0
>  34 
>  35 ; CHECK-LABEL: @foo2a
>  36 ; CHECK: load i32, i32* {{[^,]+}}, align 32
>  37 ; CHECK: ret i32
>  38 }
> ```
> 
> According to what was said before, `"align"(i32* %a, i32 32, i32 28)` means `%a + 4` is 32-bytes aligned.
> Then, `%a - 4` cannot be 32-bytes aligned, but this test checks it.
> Maybe I can ask people there.

It seems there was a bug introduced between clang 11 and clang trunk.
`__builtin_assume_aligned` is lowered straight to assume's align bundle:  https://godbolt.org/z/3oGb9x
In clang 11, it was lowered to arithmetic+assume.

So, for the lowering from clang, it's using the gcc semantics, but for optimizations is negating the last operand. There's a bug in ether LLVM or clang. Maybe let's start by fixing LangRef so we agree on the semantics?  I'm surprised the patch went in without documentation..

jdoerfert:
TBH, I am still unsure if we should have 3-operand "align" bundles in the first place. If we make the offset explicit with a GEP we sidestep the problem. If we want to stick to 3-operand bundles, I'd vote for the GCC encoding throughout.

>  I'm surprised the patch went in without documentation..

https://llvm.org/docs/LangRef.html#assume-operand-bundles has the `align` example, though, 3-operand bundles are not mentioned.

aqjune:
> TBH, I am still unsure if we should have 3-operand "align" bundles in the first place. If we make the offset explicit with a GEP we sidestep the problem. If we want to stick to 3-operand bundles, I'd vote for the GCC encoding throughout.
> 

+1 for the GCC encoding: it brings the benefits from using assume operand bundle such as supporting one-use checks.
It is less explicit than using GEP, but I think its meaning can be properly delivered by LangRef.
If okay, I can submit a short LangRef patch for this to Phabricator.

llvm_util/llvm2alive.cpp
nunoplopes: Can you replace this code with `dyn_cast<Constantint>(..)->getZExtValue()`. It gets rid of the subtraction below. It's a constant; we can negate it here.

aqjune: Its value can be larger than 64 bits (`AlignmentFromAssumptions/simple.ll`).
I'll update this to use `getZExtValue` if its size is not larger than 64 bits.

llvm_util/llvm2alive.cpp
```diff
@@ -711,14 +712,47 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
+          if (!aptr)
             return error(i);
 
-          auto *aptr = get_operand(ptr), *aalign = get_operand(align);
-          if (!aptr || !aalign)
+          auto *align = llvm::cast<llvm::ConstantInt>(bundle.Inputs[1].get());
+          if (align->uge(UINT64_MAX))
             return error(i);
+          uint64_t align_value = align->getZExtValue();
+          auto *aalign = make_intconst(align_value, 64);
```
aqjune, line 723: I found that the existing code crashes when giving >64 bit constant for the first value:
```
// Assume::toSMT
   2489	  case Align: {
   2490	    // assume(ptr, align)
   2491	    const auto &vptr = s.getAndAddPoisonUB(*args[0]);
-> 2492	    uint64_t align = *dynamic_cast<IntConst *>(args[1])->getInt(); // getInt() crashes
```
I fixed the bitwidth to 64 to make `getInt()` succeed. This is okay because a constant larger than UINT64_MAX is guarded by the `if(..uge())` above.

nunoplopes, line 723: this change doesn't make sense. the code above is equivalent to get_operand. Without caching.

llvm_util/llvm2alive.cpp
```diff
@@ -711,14 +712,47 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
+
+          if (bundle.Inputs.size() >= 3) {
+            assert(bundle.Inputs.size() == 3);
+            auto *ofs =
+                llvm::dyn_cast<llvm::ConstantInt>(bundle.Inputs[2].get());
+            Value *adjust_ofs;
+
+            if (ofs->uge(UINT64_MAX)) {
```
nunoplopes, line 731: Does this happen in practice?

llvm_util/llvm2alive.cpp
```diff
@@ -711,14 +712,47 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
...
+            } else {
+              unsigned ofs_bw = ofs->getBitWidth();
+              adjust_ofs = make_intconst(-ofs->getZExtValue(), ofs_bw);
+            }
+
+            auto gep = make_unique<GEP>(
+                aptr->getType(),
+                "#align_adjustedptr" + to_string(alignopbundle_idx++),
+                *aptr, false);
+            gep->addIdx(1, *adjust_ofs);
```
nunoplopes, line 751: we can simplify this stuff by using -1ull here. as long as only support 64-bit ptrs it's fine.

llvm_util/llvm2alive.cpp
aqjune: Then should we change `Assume::toSMT` instead? `AlignmentFromAssumptions/simple.ll` is crashing due to 128 bit alignment :/

llvm_util/llvm2alive.cpp
aqjune: Maybe not, I'll remove this branch then

llvm_util/llvm2alive.cpp
aqjune: The difference is its bit size: `getInt()` crashes with larger-than-64bits int

aqjune:
Supporting the large integer thing is quite a hassle. :(

llvm_util/llvm2alive.cpp
nunoplopes: I see. These 128-bit tests are artificial; we can leave them as unsupported.
I would just revert part of the change above so it continues to use `get_operand`. And leave a check to bail out if bits > 64.

PR #695, opened 2021-03-14
Define fncall mem block as global, always-live, inaccessible one

aqjune:
This is a follow-up of #654 and treats the memory block for recording fncall's side effect as global, always live, and inaccessible one. This reduces the size of may-alias sets.
Also, this updates the disjointness of blocks to never consider the fncall mem block because its address is unobservable.

Note that things can be further improved by reordering the block ids; currently, `bid != expr::mkUInt(get_fncallmem_bid(), bid);` is individually added to the may-alias SMT formula. This can be improved by reassigning fncallmem_bid to be placed after the pointers from loads/fncalls/etc.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -23,10 +23,10 @@ using namespace util;
   //      num_consts_src + 1 ~ num_globals_src  (non-constant globals)
   // 3. pointer argument inputs:
   //      num_globals_src + 1 ~ num_globals_src + num_ptrinputs
-  // 4. a block reserved for encoding the memory touched by calls:
-  //      num_globals_src + num_ptrinputs + 1
-  // 5. nonlocal blocks returned by loads/calls:
-  //      num_globals_src + num_ptrinputs + 2 ~ num_nonlocals_src - 1:
+  // 4. nonlocal blocks returned by loads/calls:
+  //      num_globals_src + num_ptrinputs + 1 ~ num_nonlocals_src - 2:
+  // 5. a block reserved for encoding the memory touched by calls:
+  //      num_nonlocals_src - 1
   // 6. global vars in target only:
   //      num_nonlocals_src ~ num_nonlocals_src + num_extra_nonconst_tgt - 1
   //            (non-constant globals in target only)
@@ -60,14 +60,14 @@ static bool is_constglb(unsigned bid) {
 
 // Return true if bid is the nonlocal block used to encode function calls' side
 // effects
+static unsigned get_fncallmem_bid() {
+  assert(has_fncall);
+  return num_nonlocals_src - 1;
+}
 static bool is_fncall_mem(unsigned bid) {
   if (!has_fncall)
     return false;
-  return bid == num_globals_src + num_ptrinputs + 1;
-}
-static unsigned get_fncallmem_bid() {
-  assert(has_fncall);
-  return num_globals_src + num_ptrinputs + 1;
+  return bid == get_fncallmem_bid();
 }
 void ensure_non_fncallmem(const Pointer &p) {
   if (!p.isLocal().isFalse())
@@ -670,7 +670,9 @@ static set<expr> extract_possible_local_bids(Memory &m, const expr &eptr) {
 }
 
 unsigned Memory::nextNonlocalBid() {
-  return min(next_nonlocal_bid++, num_nonlocals_src-1);
+  unsigned next = min(next_nonlocal_bid++, num_nonlocals_src - 1 - has_fncall);
+  assert(!is_fncall_mem(next));
+  return next;
 }
 
 unsigned Memory::numLocals() const {
@@ -1006,7 +1008,7 @@ static expr mk_nonlocal_liveness_array() {
   return expr::mkInt(-1, num_nonlocals);
 }
 
-void Memory::mk_nonlocal_val_axioms(bool skip_consts) {
+void Memory::mkNonlocalValAxioms(bool skip_consts) {
   if (!does_ptr_mem_access)
     return;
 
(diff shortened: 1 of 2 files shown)


aqjune:
Running single file benchmarks + LLVM unit tests

aqjune:
Single file benchmarks: bzip2 has quite a bit of improvement, no noticeable difference in other bms
#Correct (syntactically equivalent pairs excluded) : 323 -> 337
#Timeout: 540 -> 528
#OOM: 196 -> 203, but this should be improved by the next patch (bid reordering)

ir/memory.cpp
```diff
@@ -58,17 +58,36 @@ static bool is_constglb(unsigned bid) {
   return false;
 }
 
+// Return true if bid is the nonlocal block used to encode function calls' side
+// effects
+static bool is_fncall_mem(unsigned bid) {
+  if (!has_fncall)
+    return false;
+  return bid == num_globals_src + num_ptrinputs + 1;
```
nunoplopes, line 66: use `get_fncallmem_bid()` here instead

ir/memory.cpp
```diff
@@ -58,17 +58,36 @@ static bool is_constglb(unsigned bid) {
...
+}
+static unsigned get_fncallmem_bid() {
+  assert(has_fncall);
+  return num_globals_src + num_ptrinputs + 1;
+}
+void ensure_non_fncallmem(const Pointer &p) {
+  if (!p.isLocal().isFalse())
+    return;
+  uint64_t ubid;
+  ENSURE(!p.getShortBid().isUInt(ubid) || !is_fncall_mem(ubid));
```
nunoplopes, line 76: should be an assert. ENSURE is assert + guaranteed side-effects

ir/memory.cpp
```diff
@@ -1039,6 +1066,16 @@ Memory::Memory(State &state) : state(&state), escaped_local_blks(*this) {
   if (has_null_block)
     alloc(expr::mkUInt(0, bits_size_t), bits_byte / 8, GLOBAL, false, false, 0);
 
+  if (has_fncall) {
+    // Fix the nonlocal block for encoding fn calls' side effects to be always
+    // alive & have GLOBAL type.
+    // Note that liveness of the block is already encoded by
+    // mk_nonlocal_liveness_array().
+    non_local_blk_kind.add(
```
nunoplopes, line 1074: is this bit any useful?
It will just make the formula for the non-simplified case bigger. The liveness of this block is kinda irrelevant AFAICT.

nunoplopes (commented):
This patch makes formula bigger in general because of the exception for the FN block everywhere. Unless this block is placed in the beginning or the end, this patch is not an obvious improvement. Can the block be moved to one of the ends?

ir/memory.cpp
aqjune: I think it is useful because this makes liveness bitvector after fn call simpler by keeping the live bit of this block.
It makes its interpretation consistent with `always_alive(unsigned)` as well.

ir/memory.cpp
aqjune: ENSURE is used here to resolve unused variable warning of ubid.

ir/memory.cpp
```diff
@@ -994,11 +1020,21 @@ void Memory::mk_nonlocal_val_axioms(bool skip_consts) {
...
+    if (has_fncall && state->isSource())
+      // exclude fncall mem block (src)
+      bid_upperbound = numNonlocals() - 2;
+    expr loadedptr_cond = !loadedptr.isLocal(false) &&
+                          !loadedptr.isNocapture(false) &&
+                          bid.ule(bid_upperbound);
+
+    if (has_fncall && !state->isSource())
+      // exclude fncall mem block (tgt): bid_upperbound is larger than fnbid
+      loadedptr_cond &= bid != expr::mkUInt(get_fncallmem_bid(), bid);
```
aqjune, line 1034: I couldn't remove this bit because `numNonlocals()` in tgt returns `num_nonlocals` which is larger than `numNonlocals()` in src.
Even if I couldn't totally remove this check, I think invocation of this can be shrunk because `mkNonlocalValAxioms` is encoding axioms of initial memory blocks in src and tgt even if they are equivalent? For example,
```
@glb = ...
define void @src() { .. }
define void @tgt() { .. }
```
The axiom of initial value of `@glb` is encoded in src and tgt twice; I think it is enough to encode src's one only.

aqjune:
Running experiment.. Since this does quite a bit of optimizations, I hope this can bring significant performance benefit to the camera ready version

ir/memory.cpp
nunoplopes: but that adds overhead to non-debug builds. just add a (void) after ubid f.ex.

ir/memory.cpp
```diff
@@ -1523,20 +1573,26 @@ StateValue Memory::load(const Pointer &ptr, const Type &type, set<expr> &undef,
     for (auto &p : all_leaf_ptrs(*this, val.value)) {
       auto islocal = p.isLocal();
       auto bid = p.getShortBid();
-      if (!islocal.isTrue() && !bid.isConst()) {
```
nunoplopes, line 1526: I think there's no change here anymore. If not, can you pls revert to the old so the diff becomes smaller?

ir/memory.cpp
nunoplopes: But it has the side-effect of increasing the size of the expression returned by `Pointer::getAllocType()`.
You can change the loop in `Memory::mkCallState()` instead to avoid the fncall bid.

nunoplopes: Ok, I take back. The current encoding uses UFs for non-local, so there's no formula size increase. Though I still prefer patchintg mkCallState. It's not worth to add this exception to the map.

ir/memory.cpp
nunoplopes: You are right that we shouldn't call this function from the memory constructor for the tgt. Nice catch!

But this function is used for the encoding of fn calls as well (called from `Memory::setState()`). We should probably move it to `Memory::mkCallState` so it's only called from src & once.

I suggest you revert these changes and we do the proper fix afterwards to avoid making this patch any more complicated.

aqjune:
revert changes

ir/memory.cpp
aqjune: I'll make the patch for removing the redundant constraint as a follow-up.
The diff here has fncall mem bid check only.

PR #696, opened 2021-03-18
Initial nonlocal val cannot contain ptr to fncall mem blk

### LINKED ISSUE #697: False-positive with function call & undef pointer input
```llvm
@glb = global i8* null, align 8

define void @fn(i8* %p) {
  %tobool = icmp ne i8* %p, null
  br i1 %tobool, label %if.then, label %if.end

if.then:
  call void @g(i8* undef)
  br label %if.end

if.end:
  %q = load i8*, i8** @glb, align 8
  ret void
}

declare void @g(i8*)
```

To repro needs:
 - global variable w/ pointer type
 - function call with undef ptr argument
 - call in a branch

```
$ alive-tv foo.ll -smt-to=100000

@glb = global 8 bytes, align 8

define void @fn(* %p) {
%0:
  %tobool = icmp ne * %p, null
  br i1 %tobool, label %if.then, label %if.end

%if.then:
  call void @g(* undef)
  br label %if.end

%if.end:
  %q = load *, * @glb, align 8
  ret void
}
=>
@glb = global 8 bytes, align 8

define void @fn(* %p) {
%0:
  %tobool.not = icmp eq * %p, null
  br i1 %tobool.not, label %if.end, label %if.then

%if.then:
  call void @g(* undef)
  br label %if.end

%if.end:
  ret void
}

Example:
* %p = undef

Source:
i1 %tobool = any
* %q = poison

SOURCE MEMORY STATE
===================
NON-LOCAL BLOCKS:
Block 0 >       size: 0 align: 1        alloc type: 0
Block 1 >       size: 8 align: 8        alloc type: 0
Block 2 >       size: 0 align: 4        alloc type: 0
Block 3 >       size: 0 align: 2        alloc type: 0
Block 4 >       size: 0 align: 16       alloc type: 0

Target:
i1 %tobool.not = #x1 (1)
```

reduced from sqlite3 @regehr

aqjune:
This fixes #697.

### DIFF
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -1030,7 +1030,7 @@ void Memory::mkNonlocalValAxioms(bool skip_consts) {
         assert(is_fncall_mem(upperbid));
         upperbid--;
       } else if (upperbid > get_fncallmem_bid()) { // target-only glb vars exist
-        bid_cond = bid != expr::mkUInt(get_fncallmem_bid(), bid);
+        bid_cond = bid != get_fncallmem_bid();
       }
     }
     bid_cond &= bid.ule(upperbid);
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -1026,10 +1026,10 @@ void Memory::mkNonlocalValAxioms(bool skip_consts) {
     expr bid_cond(true);
     if (has_fncall) {
       // initial memory cannot contain a pointer to fncall mem block.
-      if (state->isSource()) {
-        assert(is_fncall_mem(upperbid));
+      if (is_fncall_mem(upperbid)) {
         upperbid--;
-      } else if (upperbid > get_fncallmem_bid()) { // target-only glb vars exist
+      } else {
+        assert(!state->isSource()); // target-only glb vars exist
         bid_cond = bid != get_fncallmem_bid();
       }
     }
--- a/ir/memory.cpp
+++ b/ir/memory.cpp
@@ -1009,7 +1009,7 @@ static expr mk_liveness_array() {
   return expr::mkInt(-1, num_nonlocals);
 }
 
-void Memory::mk_nonlocal_val_axioms(bool skip_consts) {
+void Memory::mkNonlocalValAxioms(bool skip_consts) {
   if (!does_ptr_mem_access)
     return;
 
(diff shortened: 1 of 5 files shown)


ir/memory.cpp
```diff
@@ -1009,24 +1009,26 @@ static expr mk_liveness_array() {
...
-                             bid.ule(numNonlocals() - 1))));
-  }
+  Byte byte(*this, non_local_block_val[i].val.load(offset));
+  Pointer loadedptr = byte.ptr();
+  expr bid = loadedptr.getShortBid();
+
+  // num_nonlocals_src - 1 is reserved as a block for recording fn call's
+  // effects
+  unsigned bid_upperbound =
+      numNonlocals() - 1 - (state->isSource() && has_fncall);
```
aqjune, line 1026: For the target, `bid != get_fncallmem_bid()` makes the expression bigger; I have no idea if it's strictly better.

aqjune:
Oops, # correct of gzip decreases a bit; 904 -> 878. I'll look into this later

ir/memory.cpp
```diff
@@ -1009,24 +1009,26 @@ static expr mk_liveness_array() {
   return expr::mkInt(-1, num_nonlocals);
 }
 
-void Memory::mk_nonlocal_val_axioms(bool skip_consts) {
+void Memory::mkNonlocalValAxiom(unsigned i) {
```
nunoplopes, line 1012: This API change is not good IMHO. Especially for the new memory encoding that doesn't use multiple blocks.

ir/memory.cpp
nunoplopes: Isn't it needed for correctness?

aqjune:
The reason of the performance regression isn't clear ATM. :(
Since I'm going to use machines for reproducing numbers, further investigation may not be possible for a while. I'll revisit this issue after drawing new graphs is done.

ir/memory.cpp
aqjune: I think it is okay - a fncallmem is never accessed unless there is a pointer variable in IR that may alias with the fncallmem.
This is already restricted by the precondition in `load` and the pointer arguments, so giving a constraint to fncallmem is duplicated.

ir/memory.cpp
aqjune: Okay, I'll change the API.

ir/memory.cpp
aqjune: Okay, I think it is necessary because otherwise the POR precondition can always be false.

aqjune:
I left diff that fixes #697 only because other diffs were relevant to performance improvements.

ir/memory.cpp
```diff
@@ -1021,11 +1021,25 @@ void Memory::mk_nonlocal_val_axioms(bool skip_consts) {
     Byte byte(*this, non_local_block_val[i].val.load(offset));
     Pointer loadedptr = byte.ptr();
     expr bid = loadedptr.getShortBid();
+
+    unsigned upperbid = numNonlocals() - 1;
+    expr bid_cond(true);
+    if (has_fncall) {
+      // initial memory cannot contain a pointer to fncall mem block.
+      if (state->isSource()) {
```
nunoplopes, line 1029: what if it's target and there are no target-only glb vars? Shouldn't we decrease upperbid as well?
e.g.
```c++
if (is_fncall_mem(upperbid)) {
  --upperbid;
} else if (...) { }
```

aqjune: yes, right. I fixed this

PR #703, opened 2021-04-10
[RFC] undef ptr cannot be local if there's no local

### LINKED ISSUE #689: Alive2 fails to verify removal of alloca in function with undef ptr argument
```llvm
define void @src() {
  %c = alloca i8, align 1
  %call = call i32 @g(i32* undef)
  ret void
}

define void @tgt() {
  %call = call i32 @g(i32* undef)
  ret void
}

declare i32 @g(i32*)
```

```
$ alive-tv bug.ll

define void @src() {
  %c = alloca i64 1, align 1
  %call = call i32 @g(* undef)
  ret void
}
=>
define void @tgt() {
  %call = call i32 @g(* undef)
  ret void
}
```

Reduced from ph7 by @regehr

aqjune:
This is a proposal to fix #689 and many regressions found while compiling Redis.

If a function does not have any local vars, undef cannot have nonlocal bit set to one.

This patch does not fully implement vector or struct of undefs yet.
If this direction makes sense, I'll continue and finish this.

### DIFF
--- a/ir/memory.h
+++ b/ir/memory.h
@@ -158,7 +158,6 @@ class Memory {
   static bool observesAddresses();
   static int isInitialMemBlock(const smt::expr &e, bool match_any_init = false);
 
-  unsigned numLocals() const;
   unsigned numNonlocals() const;
 
   smt::expr isBlockAlive(const smt::expr &bid, bool local) const;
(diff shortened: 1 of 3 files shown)


nunoplopes:
Committed a fix here after investigation: https://github.com/AliveToolkit/alive2/commit/f6d62410ee51bcdfd145738f61d2787d850ae945

PR #710, opened 2021-04-21
split TVLegacyPass::runFunction into preparation & verify

aqjune:
To simplify implementing the batch option, this patch factors out verify part from `TVLegacyPass::runFunction`.
The new `verify` static function takes `Transform` object as well as `FnInfo` and has all the important part.
`TVLegacyPass::runFunction` mainly calls `llvm2alive` to fill the objects.

With this patch, batched opt becomes simpler because what it needs to do is to pre-build src/tgt pair and simply call this `verify` whenever an unsupported optimization is met.

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -84,6 +84,19 @@ void sigalarm_handler(int) {
   _Exit(0);
 }
 
+static void printDot(const Function &tgt, int n) {
+  if (opt_print_dot) {
+    string prefix = to_string(n);
+    tgt.writeDot(prefix.c_str());
+  }
+}
+
+static string toString(const Function &fn) {
+  stringstream ss;
+  fn.print(ss);
+  return ss.str();
+}
+
 struct TVLegacyPass final : public llvm::ModulePass {
   static char ID;
   bool skip_verify = false;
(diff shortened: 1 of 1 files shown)


tv/tv.cpp
```diff
@@ -131,37 +144,45 @@ struct TVLegacyPass final : public llvm::ModulePass {
...
-      t.tgt.writeDot(prefix.c_str());
-    }
+    bool regenerate_tgt = verify(t, I->second);
+
+    if (regenerate_tgt)
+      I->second.fn = *llvm2alive(F, *TLI);
+    else
+      I->second.fn = move(t.tgt);
+
+    I->second.fn_tostr = toString(I->second.fn);
```
nunoplopes, line 167: should we move this to the regenerate_tgt branch?

aqjune: I think it should be here - `I.second.fn_tostr` was containing src's syntax regardless of regenerate_tgt.

aqjune: Regarding this issue, I'm not sure whether the original code wasn't fully correct.. `fn_tostr` must have been refreshed, but it was missing.

nunoplopes: agreed the previous code wasn't good. But if we don't regenerate the code, then we don't need to regen the string either.

aqjune: I slightly tweaked the code to keep the previous string if regenerate_tgt is false.

tv/tv.cpp
```diff
@@ -131,37 +144,54 @@ struct TVLegacyPass final : public llvm::ModulePass {
...
+      tgt_tostr = toString(t.tgt);
+
+    bool regenerate_tgt = verify(t, I->second.n, src_tostr, tgt_tostr);
+
+    if (regenerate_tgt) {
+      I->second.fn = *llvm2alive(F, *TLI);
+      I->second.fn_tostr = toString(I->second.fn);
+    } else {
+      I->second.fn = move(t.tgt);
+      I->second.fn_tostr = move(tgt_tostr);
```
nunoplopes, line 173: this doesn't make sense. when regenerate_tgt is false, we can keep the old string. Because it means that either the two functions are equal or some error occurred (and in that case src isn't valid anymore).

aqjune: Yep, you're right; this isn't necessary.

Then would it be okay if I leave a comment here saying that updating fn_tostr is not necessary due to the reason?
Otherwise, it requires readers to do deep level of interprocedural analysis; to me doing it was painful sometimes.

aqjune: (after removing this move call, I mean)

nunoplopes: ofc

aqjune: Thx!

PR #711, opened 2021-04-23
clang plugin: add -tv-batch-opts to verify optimizations in a batch

aqjune:
This is a patch that adds -tv-batched-opts that batches supported optimizations and verify them at once.


aqjune:
In high-level: tv_batchutil.so is a new plugin used to emit batched optimization list.
tv.so can take the list via `-tv-batched-opts` flag and use it during tv.

util/parallel.cpp
```diff
@@ -235,6 +236,7 @@ bool parallel::emitOutput() {
       out_file << line << '\n';
     }
   }
+  out_file.flush();
```
aqjune, line 239: These flushes were necessary to avoid a buggy behavior that is exactly #720 when batching is enabled. When batching is disabled, this bug didn't appear again.
I investigated quite amount of time for this but I couldn't find why... adding flush calls here worked and passed Dongjoo's sanity checks for the printed logs.

aqjune:
Well, I found that `registerBeforeNonSkippedPassCallback` could be used..!
It is called before a pass is to be run, so I can get the name of the *next* pass.
I'm updating this patch to see whether this works.

aqjune:
`registerBeforeNonSkippedPassCallback` seems working, so I made the clang-tv to use it for both batched and non-batched tv as well.
For non-batched tv, this change is okay because the first and last transformation (Annotation2MetadataPass and ModuleToFunctionPassAdaptor) isn't interesting.

[list.txt](https://github.com/AliveToolkit/alive2/files/6796638/list.txt) shows how `registerBeforeNonSkippedPassCallback`("Before") and  `registerAfterPassCallback`("After") is called for a simple program with -O3.
You can see that sometimes "Before" or "After" appears consecutively; it is because the optimization pipeline is kind of a tree structure and before/after is done in pre/post-order traversal, IIUC.

util/parallel.cpp
aqjune: These extra flushes were still necessary. :(

nunoplopes: I've fixed the problem here: https://github.com/AliveToolkit/alive2/commit/0f41f08816619250c96934bb9169f8bb5f500745

The flush was in the wrong place.

nunoplopes:
> `registerBeforeNonSkippedPassCallback` seems working, so I made the clang-tv to use it for both batched and non-batched tv as well.
> For non-batched tv, this change is okay because the first and last transformation (Annotation2MetadataPass and ModuleToFunctionPassAdaptor) isn't interesting.

Can we have some of protection in case things change in the future? This is fragile.
For the last pass, can't we validate the IR at finalize or something like that?

BTW, did you run clang before and after this change to ensure the list of output is the same, modulo the last pass? Is the function pass name printing ok? because now you need to store the pass name and only print it in the next before call.


> My assumption is that the "non-leaf" passes (e.g. ModuleInlinerWrapperPass) don't actually do transformation, but their leaf nodes such as InlinerPass do, so they are safe to be regarded as no-op.
> But, this means we need to cleanup `skip_pass_list` so that it does not include such non-leaf passes... For example, `skip_pass_list` contains ModuleInlinerWrapperPass, but it should be removed. What do you think?

Sure, if that doesn't introduce any false-positives.

aqjune:
Will work on this this weekend

aqjune:
I fixed this patch to make clang-tv output print the same output if `-tv-batch-opts` is not given.

Here are the test I used (`simple.c`) and outputs of `opt-alive.sh`/`alivecc` before/after this patch: [clangtv.zip](https://github.com/AliveToolkit/alive2/files/7111203/clangtv.zip)
It also contains the output of `alivecc` after this patch is applied + `-tv-batch-opts` is set.

aqjune:
> > `registerBeforeNonSkippedPassCallback` seems working, so I made the clang-tv to use it for both batched and non-batched tv as well.
> > For non-batched tv, this change is okay because the first and last transformation (Annotation2MetadataPass and ModuleToFunctionPassAdaptor) isn't interesting.
> 
> Can we have some of protection in case things change in the future? This is fragile.
> For the last pass, can't we validate the IR at finalize or something like that?

I added an assertion checking whether the first pass is `Annotation2MetadataPass` when `-tv-batch-opts` is enabled.
For checking the last pass, I realized that `ClangTVFinalizePass` was already successfully invoking TVPass at the end of transformation regardless of the `-tv-batch-opts` flag, so it is okay.

> > My assumption is that the "non-leaf" passes (e.g. ModuleInlinerWrapperPass) don't actually do transformation, but their leaf nodes such as InlinerPass do, so they are safe to be regarded as no-op.
> > But, this means we need to cleanup `skip_pass_list` so that it does not include such non-leaf passes... For example, `skip_pass_list` contains ModuleInlinerWrapperPass, but it should be removed. What do you think?
> 
> Sure, if that doesn't introduce any false-positives.

Updated skip_pass_list by removing `ModuleInlinerWrapperPass` from it.

tv/tv.cpp
```diff
@@ -523,33 +590,52 @@ llvmGetPassPluginInfo() {
...
-          return &TLI_holder.emplace(*TLIImpl, &F);
-        };
-
-        tv.run(*const_cast<llvm::Module *>(unwrapModule(IR)), get_TLI);
-      };
-      // For clang tv, manually run TVPass after each pass
-      PB.getPassInstrumentationCallbacks()->registerAfterPassCallback(
-          move(clang_tv));
+
+      if (batch_opts) {
```
aqjune, line 594: This if stmt preserves the original behavior of clang-tv without batching.

nunoplopes:
crashes here..
```
clang: ../tv/tv.cpp:606: {anonymous}::llvmGetPassPluginInfo()::<lambda(llvm::PassBuilder&)>::<lambda(llvm::StringRef, llvm::Any)>: Assertion `P.str() == "Annotation2MetadataPass"' failed.
```

nunoplopes:
it crashed with -O0 only it seems. Fixed it! 😎

aqjune:
Great, thank you! :)

PR #713, opened 2021-04-23
fix pointer comparison to use integer addresses

aqjune:
As a follow-up, I'll implement possible optimizations.

I'm thinking of optimizing 'gep inbounds p, idx == null' first.


tools/transform.cpp
```diff
@@ -827,6 +827,9 @@ static void calculateAndInitConstants(Transform &t) {
           auto &t = bc->getType();
           has_vector_bitcast |= t.isVectorType();
           min_access_size = gcd(min_access_size, getCommonAccessSize(t));
+
+        } else if (auto *ic = dynamic_cast<const ICmp*>(&i)) {
+          has_ptr2int |= ic->isPtrCmp();
```
nunoplopes, line 832: this needs optimization later..

aqjune:
FYI: after this patch, a few LLVM unit test failures are added:

Pointer replacement:
```
Transforms/CorrelatedValuePropagation/phi-common-val.ll
Transforms/GVN/condprop.ll
```

Is free(ptr) nop if (int)ptr == 0?
```
Transforms/InstCombine/malloc-free-delete-dbginvar.ll
Transforms/InstCombine/malloc-free-delete.ll
```

Real bug? (gep null, idx != glb)
```
Transforms/InstSimplify/ConstProp/icmp-null.ll

201 @g2 = global 4 bytes, align 4                                                                                                                                                                                                             
202 @g = global 8 bytes, align 4                                                                                                                                                                                                              
203                                                                                                                                                                                                                                           
204 define i1 @null_gep_ne_global() {                                                                                                                                                                                                         
205 %0:                                                                                                                                                                                                                                       
206   %__constexpr_0 = ptrtoint * @g2 to i64                                                                                                                                                                                                  
207   %gep = gep * null, 8 x i64 %__constexpr_0                                                                                                                                                                                               
208   %cmp = icmp ne * %gep, @g                                                                                                                                                                                                               
209   ret i1 %cmp                                                                                                                                                                                                                             
210 }                                                                                                                                                                                                                                         
211 =>                                                                                                                                                                                                                                        
212 @g2 = global 4 bytes, align 4                                                                                                                                                                                                             
213 @g = global 8 bytes, align 4                                                                                                                                                                                                              
214                                                                                                                                                                                                                                           
215 define i1 @null_gep_ne_global() {                                                                                                                                                                                                         
216 %0:                                                                                                                                                                                                                                       
217   ret i1 1                                                                                                                                                                                                                                
218 }                                                                                                                                                                                                                                         
```

Unknown reason (I need to investigate):
```
Transforms/InstCombine/compare-alloca.ll ; needs to be checked; I think local block's address isn't quantified
Transforms/InstCombine/gep-custom-dl.ll ; needs to be checked
Transforms/InstCombine/getelementptr.ll
```

LebedevRI:
> Is free(ptr) nop if (int)ptr == 0?

yes.

nunoplopes:
I'm also getting this:
```
Failing Tests (1):
    Alive2 :: alive-tv/ptrcmp/D99642.srctgt.ll
```

aqjune:
The failure is weird.. I'm looking into it.

aqjune:
I'm running gzip to see the performance impact of this patch

aqjune:
For gzip, # of corrects didn't change very much. 
# timeout: decreased (829 -> 813), but SMT Error increased from 10 to 20

PR #718, opened 2021-04-25
optimize icmp to only compare offsets of two pointers if they are inbounds

aqjune:
Defined get_base_ptr (actually almost copied from may_be_nonlocal) because the old collect_gep_offsets could only look into constant offsets.

One scary thing was that `*.ident.ll` tests were not being checked. :(


ir/instr.cpp
```diff
@@ -1949,11 +1952,17 @@ StateValue ICmp::toSMT(State &s) const {
...
       Pointer lhs(s.getMemory(), av);
       Pointer rhs(s.getMemory(), bv);
-      if (use_provenance) {
+      switch (pcmode) {
+      case INTEGRAL:
+        return fn(lhs.getAddress(), rhs.getAddress(), cond);
+      case PROVENANCE:
         assert(cond == EQ || cond == NE);
         return cond == EQ ? lhs == rhs : lhs != rhs;
+      case OFFSETONLY:
```
nunoplopes, line 1961: uhm, is this any useful?
It may well end up with different exprs in src/tgt because of syntactic differences in the IR. If something can be folded, the smt exprs will do it.

ir/instr.h
```diff
@@ -228,21 +228,22 @@ class InsertValue final : public Instr {
...
+  PtrCmpMode pcmode = INTEGRAL;
   smt::expr cond_var() const;
 
 public:
   ICmp(Type &type, std::string &&name, Cond cond, Value &a, Value &b);
 
   bool isPtrCmp() const;
-  bool usesProvenance() const { return use_provenance; }
-  void setUseProvenance(bool flag) { use_provenance = flag; }
+  bool getPtrCmpMode() const { return pcmode; }
```
nunoplopes, line 245: typo: bool

Though I don't think it's worth to introduce offsetonly.

ir/instr.cpp
aqjune: I extracted offset because the meaning of ule/ult/... bids isn't clear. transform.cpp creates this mode when they point to the same base object, but without the knowledge comparing bids looks a bit counterintuitive.
I thought in terms complexity of expression it was equivalent. In which case does it cause syntactic difference?

aqjune: Well, comparing offsets only would be not very intuitive too. What I thought was to leave a comment at `OFFSETONLY` saying that this is meaningful only if the two pointers have same objects, but leaving it at `USEPROVENANCE` would work too.
In terms of syntactic difference I didn't understand yet.

tools/transform.cpp
```diff
@@ -538,6 +538,48 @@ static bool returns_local(const Value &v) {
...
+      continue;
+    }
+
+    if (auto s = dynamic_cast<Select*>(ptr)) {
+      todo.emplace_back(s->getTrueValue());
+      todo.emplace_back(s->getFalseValue());
+      continue;
+    }
+
+    if (base_ptr)
```
nunoplopes, line 573: what if it's equal to ptr?

aqjune: oops.. I'll fix it.

ir/instr.cpp
nunoplopes: this is ok, i hadn't read the whole patch.
Essentially you wanted to avoid reinstating the old Pointer comparison operators. This is fine as long as we don't compare ptrs of different objs, which I guess it should be the common case.

ir/instr.cpp
```diff
@@ -1949,11 +1952,17 @@ StateValue ICmp::toSMT(State &s) const {
...
+        return fn(lhs.getOffset(), rhs.getOffset(), cond);
+      default:
```
nunoplopes, line 1963: setting a default in a switch of a enum is not a great idea as it elides future warnings of missing cases handling

ir/instr.cpp
aqjune: yep, the old pointer comparison operators are not necessary yet I think.

ir/instr.cpp
aqjune: I moved it after the switch stmt.

Issue #717, opened 2021-04-25
Perf regression with integer-based ptr comparison


nunoplopes:
I'm seeing a significant perf regression after https://github.com/AliveToolkit/alive2/commit/2b4ee318b5b1f939439cf4fc65debc8096694163. Subsequent optimization didn't help much @aqjune.
gzip shows no regression. bzip2 (worst in perf), oggenc, ph7, sqlite3, gcc (worst in % drop in verified).

The most common pattern is comparison of a ptr arg with null, e.g.:
```c
if (p == null)
  return;
*p = 42;
```
(sometimes with a gep before the store/load)

I also see comparisons of return of fncalls with null. And comparison of loaded gvar with an argument.

A useful observation is that addresses are never observed in these functions. For the program above there are 4 cases:
1. p is null (early return)
2. p is OOB and zero (early return)
3. p is OOB and non-zero (UB)
4. p is in-bounds

The exact address of p is irrelevant. However, detection of case 2. requires to prove that `-offset == base_addr`.
Moreover, in these functions the overlapping of addresses between blocks is irrelevant, as they are not observed. (though important for counter-examples, but let's ignore that for a second).

Some ideas:
 - Skip emission of non-overlapping addresses axiom when addresses are not observed (with a more relaxed definition). This might be harder than it sounds as e.g. `p + i == null && q + i == null` cannot imply that p/q are equal.
 - Change non-local overlapping addresses axiom to use a quantifier when using more than x arguments, or when we think addresses aren't observed (no int2ptr/ptr2int or only 1). Hopefully Z3 will only instantiate the axioms lazily and therefore we save some formula size.
 - Specialize pointers to have an extra `is_zero` bit to cover functions that don't do int2ptr/ptr2int but do comparisons with null only (to cover case 2. above).
 - Split observability into local/non-local to avoid instantiating axioms unnecessarily

Other cases that can be optimized:
- [x] non-eq/ne between inbounds ptrs of same base ptr.

Other crazy ideas?

aqjune:
> A useful observation is that addresses are never observed in these functions. For the program above there are 4 cases:
> 
> 1. p is null (early return)
> 2. p is OOB and zero (early return)
> 3. p is OOB and non-zero (UB)
> 4. p is in-bounds

The second case seems problematic indeed.
A random thought is that in the source language (C) an OOB pointer cannot be passed as an input argument.
I don't know whether allowing it will miss any bug in practice... what do you think?

> 
[...]

PR #719, opened 2021-04-26
split has_ptr2int into local and nonlocal versions

aqjune:
As I mentioned in #717, if there is a pointer comparison between non-local pointers only, we don't need to encode addresses of local blocks, and vice versa.

This seems to happen in quite a few cases because alloca with struct type typically remains after SROA but they are not compared with other pointers.

I was so curious about its benefit so quickly implemented it.
I'll leave the experimental results here.


aqjune:
Closing this since its performance benefit isn't clear & I don't think I'll have time for this :/

PR #721, opened 2021-04-27
getMaxAllocSize should consider alignment

aqjune:
This fixes the regression of `Transforms/InstCombine/compare-alloca.ll`.

```
define i1 @src(i64* %arg) {
  %alloc = alloca i64
  %cmp = icmp eq i64* %arg, %alloc
  ret i1 %cmp
}

define i1 @tgt(i64* %arg) {
  ret i1 false
}
```

This failed because `bits_size_t` was too small.
It caused `%alloc` to be located in only one place in the memory, so properly guessing `%arg` could make the comparison always true.
`bits_size_t` was set to 5, so `%alloc` should be somewhere at `10000 ~ 11111`.
However, `%alloc`'s size is 8 and it has alignment 8, so only `10000` was the possible choice. `11000` doesn't work because adding 8 overflows.

To fix this, I updated `getMaxAllocSize` to consider alignment as well because it is the true size that occupies memory.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2622,14 +2622,22 @@ DEFINE_AS_RETZERO(Alloc, getMaxGEPOffset);
 DEFINE_AS_EMPTYACCESS(Alloc);
 DEFINE_AS_RETFALSE(Alloc, canFree);
 
+static uint64_t local_alloc_size(uint64_t size, unsigned align) {
+  // To find an available space, alignment should be considered as well.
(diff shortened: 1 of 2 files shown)


nunoplopes:
I added a more comprehensive fix that doesn't degrade the perf of the non-observed case. I added a couple of optimizations while at it.
https://github.com/AliveToolkit/alive2/commit/15c5d528b7ae41365b0f090566cd95940b1b1586

PR #722, opened 2021-04-27
clang plugin: the messages before getline should be clipped out

### LINKED ISSUE #720: clang plugin printing passes multiple times
When the parallel mode is used, clang plugin prints the same pass names multiple times:
```
-- 1. Annotation2MetadataPass
-- 2. ForceFunctionAttrsPass
....
-- 12. SROA
-- 1. Annotation2MetadataPass
-- 2. ForceFunctionAttrsPass
...
```

My option is:
```
export ALIVECC_REPORT_DIR=$ALIVE2_DIR/gzip.results
export ALIVECC_PARALLEL_UNRESTRICTED=1
export ALIVECC_SUBPROCESS_TIMEOUT=600
export ALIVECC_SMT_TO=60000
export ALIVECC_MAX_MEM=1024
export ALIVECC_QUIET=1

time $ALIVE2_DIR/build/alivecc -O3 -S -o -emit-llvm -o temp.ll gzip.c -mllvm --max-subprocesses=12
```
I used gzip.c from here: https://people.csail.mit.edu/smcc/projects/single-file-programs/

aqjune:
This fixes #720 by fixing `parallel::emitOutput()` to clip out the lines before the line read by `getline`.

An example: https://gist.github.com/aqjune/2ea18cefa4fc1fd98ba1e47ddeab5f97

I never knew that `sstream::str` has the full string even if it is being consumed :(

### DIFF
--- a/util/parallel.cpp
+++ b/util/parallel.cpp
@@ -206,6 +206,8 @@ bool parallel::emitOutput() {
   ensureParent();
   std::string line;
   std::regex rgx("^include\\(([0-9]+)\\)$");
+
+  auto befpos = parent_ss.tellg();
   while (getline(parent_ss, line)) {
     std::smatch sm;
     if (std::regex_match(line, sm, rgx)) {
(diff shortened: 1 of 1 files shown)


nunoplopes:
this is not the right fix. It should simply rewind the stream with https://en.cppreference.com/w/cpp/io/basic_istream/seekg

aqjune:
makes sense..! I'll use it

nunoplopes:
Actually I was wrong. The code needs to prune the prefix of the stream once in a while, otherwise the parent accumulates the whole thing in memory until the last child exits.
So your previous fix was good. Sorry, can you remove the last commit?

aqjune:
Well, to me computation time was more precious resource than memory so using seekg was actually better.
I reverted the last commit as requested.

PR #723, opened 2021-04-27
clang plugin: let stats cmd flag work by printing it when child exits

aqjune:
Currently clang plugin's `-mllvm -tv-alias-stats` isn't working because the statistics accumulated in the child process isn't being summed up to the parent process.

This is a small patch that makes children to print the statistics.


nunoplopes:
well, this is a bit suboptimal as it makes the behavior different to the sequential version. But fixing this properly isn't trivial either..

PR #738, opened 2021-08-05
Support for x86 PSRLs

zhengyang92:
Hi @nunoplopes , this pull request is the first patch of the X86 SIMD intrinsics semantics. This patch adds the support of llvm.x86.sse2.psrl.[d|w|q] and llvm.x86.avx2.psrl.[d|w|q]. These intrinsics are lowered into vpsrl[d|w|q] [1] by codegen [2]. I followed intel intrinsics guide when I was writing the semantics.

In this patch I added the X86IntrinBinOp class and this class will be the place for all the other X86 vector intrinsics with two operands. Please let me know if you like the style or not. I am going to add the other intrinsics if this patch is approved.

[1] https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=vpsrld&expand=6690 
[2] https://github.com/llvm/llvm-project/blob/8a557d8311593627efd08d03178889971d5ae02b/llvm/test/CodeGen/X86/avx2-intrinsics-x86.ll#L422


zhengyang92:
I have ran TV on the Transform directory. There are two Instcombine test cases that uses these operands and they are failing after the patch. There are no changes in other test cases. The problematic transformations are  https://github.com/llvm/llvm-project/blob/656001e7b2b939d9bce4fb58831d314dc67ddf7a/llvm/test/Transforms/InstCombine/X86/x86-vector-shifts.ll#L2717  and https://github.com/llvm/llvm-project/blob/d480f968ad8b56d3ee4a6b6df5532d485b0ad01e/llvm/test/Transforms/InstCombine/X86/x86-vector-shifts-inseltpoison.ll#L2717

```
define <4 x i32> @sse2_psrl_d_128_masked(<4 x i32> %v, <4 x i32> %a) {
; CHECK-LABEL: @sse2_psrl_d_128_masked(
; CHECK-NEXT:    [[TMP1:%.*]] = and <4 x i32> [[A:%.*]], <i32 31, i32 poison, i32 poison, i32 poison>
; CHECK-NEXT:    [[TMP2:%.*]] = shufflevector <4 x i32> [[TMP1]], <4 x i32> poison, <4 x i32> zeroinitializer
; CHECK-NEXT:    [[TMP3:%.*]] = lshr <4 x i32> [[V:%.*]], [[TMP2]]
; CHECK-NEXT:    ret <4 x i32> [[TMP3]]
;
  %1 = and <4 x i32> %a, <i32 31, i32 0, i32 undef, i32 undef>
  %2 = tail call <4 x i32> @llvm.x86.sse2.psrl.d(<4 x i32> %v, <4 x i32> %1)
  ret <4 x i32> %2
}
```

The transformation tries to rewrite psrls into paralleled lshrs. The optimization is fine except for the case where a[0] is a undef. If a[0] == undef, in src, psrl used it for one time, but, in tgt, the undef is broadcasted to all the other lanes by shufflevector and executing lshr vertically would introduce many uses of the undef. And under the “Target must be more defined source” criteria, this optimization is invalid. There is a straightforward fix, inserting a freeze before shufflevector will prevent the undef from being broadcasted.  

There are a few other test cases which won't pass TV due to undef input, similar to this one. What do you think? Should I report this?

tests/alive-tv/vector/x86/sse2_psrl_w-poison1.srctgt.ll
```diff
@@ -0,0 +1,10 @@
+define <8 x i16> @src(<8 x i16> %v) {
+  %1 = call <8 x i16> @llvm.x86.sse2.psrl.w(<8 x i16> %v, <8 x i16> <i16 0, i16 0, i16 poison, i16 0, i16 0, i16 0, i16 0, i16 0>)
+  ret <8 x i16> %1
+}
+
+define <8 x i16> @tgt(<8 x i16> %v) {
```
nunoplopes, line 6: this sort of semantics is not common in LLVM. We've been using element-wise poison semantics. Is there a reason to deviate from that?

ir/instr.h
```diff
@@ -940,6 +940,59 @@ class ShuffleVector final : public Instr {
...
+  /* llvm.x86.sse2.psrl.d */ sse2_psrl_d,
+  /* llvm.x86.sse2.psrl.q */ sse2_psrl_q,
+  /* llvm.x86.avx2.psrl.w */ avx2_psrl_w,
+  /* llvm.x86.avx2.psrl.d */ avx2_psrl_d,
+  /* llvm.x86.avx2.psrl.q */ avx2_psrl_q
+  };
+
+  // the shape of a vector is stored as <# of lanes, element bits>
+  static constexpr std::array<std::pair<unsigned, unsigned>, 6> shape_op0 = {
+  /* sse2_psrl_w */       std::make_pair(8, 16),
```
nunoplopes, line 956: use {} instead of make_pair

ir/instr.h
```diff
@@ -940,6 +940,59 @@ class ShuffleVector final : public Instr {
...
+  /* llvm.x86.sse2.psrl.w */ sse2_psrl_w,
+  /* llvm.x86.sse2.psrl.d */ sse2_psrl_d,
+  /* llvm.x86.sse2.psrl.q */ sse2_psrl_q,
+  /* llvm.x86.avx2.psrl.w */ avx2_psrl_w,
+  /* llvm.x86.avx2.psrl.d */ avx2_psrl_d,
+  /* llvm.x86.avx2.psrl.q */ avx2_psrl_q
+  };
+
+  // the shape of a vector is stored as <# of lanes, element bits>
+  static constexpr std::array<std::pair<unsigned, unsigned>, 6> shape_op0 = {
```
nunoplopes, line 955: there's really no reason to use array here vs normal C array

nunoplopes:
Stepping back for a second, why are you targeting this specific x86 instruction?
I'm very nervous about start adding support for assembly in an ad-hoc way and without a proper infrastructure to validate the semantics against CPUs. There are really complicated instructions and writing their spec without tests is a recipe for disaster.
Plus we really want to engineer this in a way that the spec is reusable between LLVM intrinsics and assembly. It's not worth to do it twice.

zhengyang92:
> Stepping back for a second, why are you targeting this specific x86 instruction?

I am working on my vector superoptimizer based on alive2 and I have already encoded about a hundred instructions in this way. I randomly chose those to make the pull request easy for you to review :)

tests/alive-tv/vector/x86/sse2_psrl_w-poison1.srctgt.ll
zhengyang92: The instruction uses the second vector operand as a 64 bit integer, whether the result is poison is not dependent on the element higher than 64 bit, am I right?

regehr:
> I'm very nervous about start adding support for assembly in an ad-hoc way and without a proper infrastructure to validate the semantics against CPUs. There are really complicated instructions and writing their spec without tests is a recipe for disaster.

Can you say a bit more about what this infrastructure might look like? One option will be to use Nader's interpreter to validate lots of litmus tests against execution on CPUs, assuming that we can find some nice collections of litmus tests for x86 vector instructions.

> Plus we really want to engineer this in a way that the spec is reusable between LLVM intrinsics and assembly. It's not worth to do it twice.

Just to be clear, are we talking about inline assembly, or are we talking about the new assembly semantics that we want for backend verification?

I've asked Zhengyang to look into factoring each instruction into the machine part + the LLVM part. I suppose that nearly all of the time, the LLVM part would only include the UB-related stuff.

nunoplopes:
The question 0 is can we reuse someone else's spec? The UIUC folks claimed that had a full spec of AVX, for example. That doesn't include poison, but LLVM's rules for poison can be mechanized easily.
Around 4/5 years ago I did a round and I couldn't find anything that could be reused easily, so maybe the answer is still no.

For testing, if we focus just of assembly instructions exposed as intrinsics right now, and if we assume we can run llc on a function with a single intrinsic and get it to use the respective assembly instruction, then it shouldn't be hard to do.
Something like:
 1) Fetch all LLVM intrinsics from the .td file and filter out those that return void or have no arguments.
 2) Test each one individually by creating a function that just calls the intrinsic with random arguments and prints the result.
 3) Run llc on the file & run it.
 4) Create an .ll file with the random intrinsic call as @src function and the output printed from the program as @tgt.
 5) Run alive-tv.

This method supports any UB, poison, crashes (eg, refinement from UB), etc. It needs a printer & random generator per data type. Pure random may be too naive, because it may end up triggering UB/poison cases all the time, hence we get a trivial refinement, but should be a good start. I guess a smarter way would be to even use alive to produce test cases that produce different results. But random is a great start.
I've written the spec of a bunch of AVX2 instructions in the past, and fuzzing uncovered plenty of bugs, hence my concern.

If we are going to write the spec of "hundreds" of instructions, I would rather write that in a way that we can use it for backend verification. The poison part is fine for both, as we can ignore it for assembly. The UB is LLVM only, and the machine may have other things like crashes, signals, whatever. We need to abstract the machine state so we can reuse the main part. Shouldn't be too hard; I think we can get here in a couple of iterations.
Alas, these vector intrinsics have no UB or crashes.

regehr:
ok @nunoplopes, this all makes sense and sounds good, but it's kind of another project and will take some time to get around to it, I suspect. @zhengyang92 want to close this for now?

zhengyang92:
thank you @nunoplopes , I'll close the PR for now.

Issue #650, opened 2021-01-18
writeonly fns cannot read memory, but fninputRefined violates it


aqjune:
Consider this function:
```
define void @f(i8* %p) {
  %call = call i8* @g1(i8* %p)
  call void @g2(i8* %call)
  ret void
}

declare i8* @g1(i8*)
declare void @g2(i8*) writeonly
```

Running alive-tv with an identity input raises `ERROR: Source is more defined than target`, and the reason is pretty interesting.

When encoding the refinement of g2's inputs, `Pointer::fninputRefined` is called. `fninputRefined` is checking `isBlockAlive().implies(other.isBlockAlive())`, but this isn't well-defined because g2 is a writeonly function.
`State::addFnCall` is filling the source's input memory with just an initial memory.

Should `Pointer::fninputRefined` take an additional flag which doesn't check liveness? `Pointer::fninputRefined` already has `is_byval_arg` flag.

nunoplopes:
nice catch!
Then I guess we do need to pass the memory. Even if a function only writes to memory, it has to write to valid blocks. So if it writes through an argument, that argument better be alive.
The whole problem is that the comparison function takes all fields of memory into account instead of a subset. e.g. for readonly it wouldn't compare the value field.

aqjune:
> Then I guess we do need to pass the memory. Even if a function only writes to memory, it has to write to valid blocks. So if it writes through an argument, that argument better be alive.

+1, it makes sense.

> The whole problem is that the comparison function takes all fields of memory into account instead of a subset. e.g. for readonly it wouldn't compare the value field.

I didn't fully understand this, could you elaborate a bit more please?

aqjune:
IIUC, calls_fn update in State::addFnCall should be fixed:
```
    auto call_data_pair
      = calls_fn.try_emplace(
          { move(inputs), move(ptr_inputs),
            reads_memory ? analysis.ranges_fn_calls
                         : State::ValueAnalysis::FnCallRanges(), <<< this and
            reads_memory ? memory : Memory(*this),            <<< this should consider `writes_memory` as well?
            reads_memory, argmemonly });
```

nunoplopes:
Right. the point is that doing it right now will regress performance because calls_fn relies on Memory::operator< which compares all fields. While for e.g. readonly we shouldn't be comparing the local/non-local values. We achieve that right now by creating a fresh memory.

aqjune:
[...]

PR #753, opened 2021-10-10
Encode the basic refinement of escaped local blocks

### LINKED ISSUE #435: ERROR: Precondition is always false
This code seems to not roundtrip to itself:
```
void use(char&);

void loop() {
    char storage[16] = {};
    for(char& c : storage) use(c);
}
```
https://godbolt.org/z/UPbp51

```
define void @_Z4loopv() {
%entry:
  %storage = alloca i64 16, align 16, dead
  %0 = gep inbounds * %storage, 16 x i64 0, 1 x i64 0
  start_lifetime * %0
  memset * %0 align 16, i8 0, i64 16
  %add.ptr = gep inbounds * %storage, 16 x i64 0, 1 x i64 16
  br label %for.body

%for.body:
  %__begin1.09 = phi * [ %0, %entry ], [ %incdec.ptr, %for.body ]
  call void @_Z3useRc(nonnull dereferenceable(1) * %__begin1.09)
  %incdec.ptr = gep inbounds * %__begin1.09, 1 x i64 1
  %cmp = icmp eq * %incdec.ptr, %add.ptr
  br i1 %cmp, label %for.cond.cleanup, label %for.body

%for.cond.cleanup:
  free * %0 unconstrained
  ret void
}
=>
define void @_Z4loopv() {
%entry:
  %storage = alloca i64 16, align 16, dead
  %0 = gep inbounds * %storage, 16 x i64 0, 1 x i64 0
  start_lifetime * %0
  memset * %0 align 16, i8 0, i64 16
  %add.ptr = gep inbounds * %storage, 16 x i64 0, 1 x i64 16
  br label %for.body

%for.body:
  %__begin1.09 = phi * [ %0, %entry ], [ %incdec.ptr, %for.body ]
  call void @_Z3useRc(nonnull dereferenceable(1) * %__begin1.09)
  %incdec.ptr = gep inbounds * %__begin1.09, 1 x i64 1
  %cmp = icmp eq * %incdec.ptr, %add.ptr
  br i1 %cmp, label %for.cond.cleanup, label %for.body

%for.cond.cleanup:
  free * %0 unconstrained
  ret void
}
ERROR: Precondition is always false

Summary:
  0 correct transformations
  0 incorrect transformations
  1 Alive2 errors
```

I wanted to play around and see what alive2 things about different memsets,
in particular dropping overflowing memset (should fail, but again fails with
`ERROR: Precondition is always false`), clamping overflowing memset (should pass,
but again fails with `ERROR: Precondition is always false`), etc.

aqjune:
This patch checks the bytes of local blocks escaped via fn calls (resolves #435 ).

This check is done when the local block ids are known to be constants (which is common). Otherwise, it fallbacks into the original checking.

The beginning point for reading the diff of this PR would be `Pointer::fninputRefined`.
If the pointer bids are constant and local, it now calls `Pointer::encodeLocalPtrRefinement` which is a new member function.

Among `Pointer::encodeLocalPtrRefinement`'s arguments, `readsbytes` is:
- if true, compare the blocks’ bytes as well as all other block attributes (which is implemented in `Memory::blockRefined`)
    * Note that, previously, Memory::blockRefined was receiving one non-local block id; in this PR, `bid_other` optional  argument is added. If it is `some(bid_other)`, refinement of two local blocks `bid` and `bid_other` is encoded.
    * Memory::blockRefined again calls refinement between bytes, which can be pointer bytes. It calls Pointer::refined, and `Pointer::refined`’s block_refined is still commentized, so recursive pointer refinement call does not happen.
- if false, compare block attributes except bytes (`Memory::blockPropertiesRefined`).

The byval encoding is updated to check the contents as well.
It follows the encoding of the cav paper.

### DIFF
--- a/tests/alive-tv/calls/writeonly-local.srctgt.ll
+++ b/tests/alive-tv/calls/writeonly-local.srctgt.ll
@@ -1,4 +1,5 @@
 ; This is a valid transformation, but alive-tv reports it as incorrect due to issue 650.
+; SKIP-IDENTITY
 
 define void @src() {
   %p = alloca i32
@@ -14,4 +15,5 @@ define void @tgt() {
 }
 
 declare void @writeonly(i32*) writeonly
+
 ; ERROR: Source is more defined than target


aqjune:
Ready to be reviewed.

aqjune:
The updated unit tests succeed, but I'll run LLVM unit tests and see how it goes (not sure I'll be back today).

aqjune:
Four new failures appear, and the reasons are as follows:

Transforms/DeadStoreElimination/captures-before-call.ll
Transforms/DeadStoreElimination/captures-before-load.ll
-> They fail due to https://github.com/AliveToolkit/alive2/issues/650 

Transforms/MemCpyOpt/callslot.ll
-> This fails because the pointer refinement defined in CAV isn't sufficient! :( Pointers of different offsets must be allowed.
```
define void @dest_is_gep_nounwind_call() {
%0:
  %dest = alloca i64 16, align 1
  %src = alloca i64 8, align 1
  %src.i8 = bitcast * %src to *
  %dest.i8 = gep * %dest, 16 x i64 0, 1 x i64 8
  call void @accept_ptr(* %src.i8)
  memcpy * %dest.i8 align 1, * %src.i8 align 1, i64 8
  ret void
}
=>
define void @dest_is_gep_nounwind_call() {
%0:
  %dest = alloca i64 16, align 1
  %src = alloca i64 8, align 1
  %dest.i8 = gep * %dest, 16 x i64 0, 1 x i64 8
  %dest.i81 = bitcast * %dest.i8 to *
  %dest.i812 = bitcast * %dest.i81 to *
  call void @accept_ptr(* %dest.i812)
  ret void
}
```

Transforms/LICM/promote-capture.ll
-> I think this is a bug in Alive2. A phi node is not successfully getting a value from the operand, or the counter example is not being correctly printed.
```
  %if:
    %c.inc = add i32 %c.inc2, 1
    br label %latch
  
  %latch:
    %c.inc1 = phi i32 [ %c.inc, %if ], [ %c.inc2, %loop ]
```
CEx:
```
    >> Jump to %if
  i32 %c.inc = #x00000001 (1)
    >> Jump to %latch
  i32 %c.inc1 = #x00000000 (0)
```

aqjune:
For the first two failures, I added tests/alive-tv/calls/writeonly-local.srctgt.ll to this PR that must succeed after the issue is fixed.

yotann:
These tests have invalid transformations involving escaped local blocks, but Alive2 doesn't detect the problems even with this PR:

- `alive-tv/bugs/pr1014.srctgt.ll`
- `alive-tv/bugs/pr23599.srctgt.ll`

PR #765, opened 2021-11-02
Signed comparison of pointers must compare underlying addrs

aqjune:
This fixes a bug due to an incorrect ptr comparison encoding.

Consider `p = gep inbounds p0, 0`, `q = gep inbounds p0, 1` where p0 points to the beginning of some block.

`icmp slt p, q` cannot be optimized to a comparison of their block offsets `icmp slt 0, 1` because the block p0's address may be 0x7fffffff (which is signed_int_max).
In this case, q=p+1 is 0x80000001, which is signed_int_min.

Therefore, `icmp slt p, q` is false, whereas `icmp slt 0, 1` is true.

The counter example of `slt.srctgt.ll` shows the case:

```
ERROR: Value mismatch                                
                                                           
Example:                                             
* %ptr = pointer(non-local, block_id=1, offset=5)    
                                                                                                                       
Source: 
* %first1 = pointer(non-local, block_id=1, offset=5)
* %last1 = pointer(non-local, block_id=1, offset=47)  
i1 %cmp = #x1 (1)              
                                                           
SOURCE MEMORY STATE      
===================                                                                                                    
NON-LOCAL BLOCKS:
Block 0 >       size: 0 align: 1        alloc type: 0   address: 0
Block 1 >       size: 64        align: 1        alloc type: 0   address: 103
                                                                                                                       
Target:                                                                                                                
* %first1 = pointer(non-local, block_id=1, offset=5)     <----- 103+5=108: less than sign_int_max (127)
* %last1 = pointer(non-local, block_id=1, offset=47)   <----- 103+47=150: overflows sign_int_max
i1 %cmp = #x0 (0)                                                                                                      
Source value: #x1 (1)                                    
Target value: #x0 (0)                                                                                                  
```

### DIFF
--- a/tests/alive-tv/ptrcmp/slt.srctgt.ll
+++ b/tests/alive-tv/ptrcmp/slt.srctgt.ll
@@ -13,5 +13,3 @@ entry:
   %cmp = icmp slt i8* %first1, %last1
   ret i1 %cmp
 }
-
-
--- a/tests/alive-tv/ptrcmp/slt.srctgt.ll
+++ b/tests/alive-tv/ptrcmp/slt.srctgt.ll
@@ -1,0 +1,17 @@
+define i1 @src(i8* %ptr) {
+entry:
+  %first1 = getelementptr inbounds i8, i8* %ptr, i32 0
+  %last1 = getelementptr inbounds i8, i8* %ptr, i32 42
+  %cmp = icmp slt i8* %first1, %last1
+  ret i1 %cmp
+}
+
+define i1 @tgt(i8* %ptr) {
+entry:
+  %first1 = getelementptr inbounds i8, i8* %ptr, i32 0
+  %last1 = getelementptr i8, i8* %ptr, i32 42
+  %cmp = icmp slt i8* %first1, %last1
+  ret i1 %cmp
+}
+
+
--- a/tools/transform.cpp
+++ b/tools/transform.cpp
@@ -1342,6 +1342,8 @@ static void optimize_ptrcmp(Function &f) {
 
     auto cond = icmp->getCond();
     bool is_eq = cond == ICmp::EQ || cond == ICmp::NE;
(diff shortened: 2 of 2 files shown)


nunoplopes:
btw, this patch exposed a bug in LLVM:
```
define i1 @gep_same_base_constant_indices(i8* %a) {
; CHECK-LABEL: @gep_same_base_constant_indices(
; CHECK-NEXT:    ret i1 true
;
  %arrayidx1 = getelementptr inbounds i8, i8* %a, i64 1
  %arrayidx2 = getelementptr inbounds i8, i8* %a, i64 10
  %cmp = icmp slt i8* %arrayidx1, %arrayidx2
  ret i1 %cmp
}
```

only correct if comparison is ult. Maybe mostly correct in practice as kernels often reserve half of the address space for themselves?

regehr:
well, not all kernels do that sort of thing, I believe both Linux and Windows support a 3 GB / 1 GB split on machines with 32-bit address space. I suspect on a machine with enough RAM it's not hard to cross this boundary, unless I'm misunderstanding something. is it worth it for me to make a 32-bit VM to demonstrate this?

nunoplopes:
Ah no, I think we agree it's a bug in LLVM. I'll report it next.

It was a good outcome that your testcase ended up exposing a bug in (an optimization in) Alive2 that was hiding a bug in LLVM 😀
`inbounds` is the main attribute we try to take advantage of. Maybe generating test cases where this is dropped can expose more issues.

regehr:
Ya, dropping that flag is something Yuyou's mutation engine will tend to do pretty often, so I think we'll have plenty of opportunity to find these.

I'm currently (slowly) dropping test cases that trigger known issues from a fork of the unit test suite, so hopefully this work well become less painful in the future. It's currently finding many instances of something silly where LLVM optimizes a 1-byte memcpy to a load/store pair, ugh.

PR #767, opened 2021-11-13
fptosi/fptoui must convert (-1.0, -0.0) to 0, not poison

aqjune:
This fixes bugs reported in https://reviews.llvm.org/D113543.

fptosi/fptoui's poison checking expressions were as follows:
```
      expr bv  = val.fp2sint(to_type.bits());
      expr fp2 = bv.sint2fp(val);
      // -0.0 is converted to 0 and then to 0.0, though -0.0 is ok to convert
      return { move(bv), val.isFPZero() || fp2 == val.roundtz() };
```
If the input (`val`) was -0.5, `bv` is 0, `fp2` is +0.0.
Therefore, `fp2 == val.roundtz()` is false because `val.roundtz()` is -0.0.

To fix this, the poison condition's first part (`val.isFPZero()`) is replaced with a range check.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1259,14 +1259,24 @@ StateValue ConversionOp::toSMT(State &s) const {
     fn = [](auto &&val, auto &to_type) -> StateValue {
       expr bv  = val.fp2sint(to_type.bits());
       expr fp2 = bv.sint2fp(val);
-      // -0.0 is converted to 0 and then to 0.0, though -0.0 is ok to convert
-      return { move(bv), val.isFPZero() || fp2 == val.roundtz() };
+      // -0.xx is converted to 0 and then to 0.0, though -0.xx is ok to convert
+      expr minus_one = expr::mkFloat(-1.0, val);
+      expr minus_zero = expr::mkFloat(-0.0, val);
+      return { move(bv),
+          (val.fogt(minus_one) && val.fole(minus_zero)) ||
+          fp2 == val.roundtz() };
     };
     break;
   case FPToUInt:
     fn = [](auto &&val, auto &to_type) -> StateValue {
       expr bv  = val.fp2uint(to_type.bits());
       expr fp2 = bv.uint2fp(val);
+      // -0.xx must be converted to 0, not poison.
+      expr minus_one = expr::mkFloat(-1.0, val);
+      expr minus_zero = expr::mkFloat(-0.0, val);
+      return { move(bv),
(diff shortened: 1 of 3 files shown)


ir/instr.cpp
```diff
@@ -1259,14 +1259,24 @@ StateValue ConversionOp::toSMT(State &s) const {
...
     fn = [](auto &&val, auto &to_type) -> StateValue {
       expr bv  = val.fp2uint(to_type.bits());
       expr fp2 = bv.uint2fp(val);
+      // -0.xx must be converted to 0, not poison.
+      expr minus_one = expr::mkFloat(-1.0, val);
+      expr minus_zero = expr::mkFloat(-0.0, val);
+      return { move(bv),
+          (val.fogt(minus_one) && val.fole(minus_zero)) ||
+          fp2 == val.roundtz() };
       return { move(bv), fp2 == val.roundtz() };
```
jrtc27, line 1280: This is dead

ir/instr.cpp
```diff
@@ -1259,14 +1259,24 @@ StateValue ConversionOp::toSMT(State &s) const {
     fn = [](auto &&val, auto &to_type) -> StateValue {
       expr bv  = val.fp2sint(to_type.bits());
       expr fp2 = bv.sint2fp(val);
-      // -0.0 is converted to 0 and then to 0.0, though -0.0 is ok to convert
-      return { move(bv), val.isFPZero() || fp2 == val.roundtz() };
+      // -0.xx is converted to 0 and then to 0.0, though -0.xx is ok to convert
+      expr minus_one = expr::mkFloat(-1.0, val);
+      expr minus_zero = expr::mkFloat(-0.0, val);
+      return { move(bv),
```
jrtc27, line 1265: For both of these I think the more "pure" approach of:
```
expr valrtz = val.roundtz();
return { move(bv), valrtz.isFPZero() || fp2 == valrtz };
```
would work?

ir/instr.cpp
aqjune: Fixed, thanks

ir/instr.cpp
aqjune: It works, but I expect the SMT solver's performance is better for the range check because comparison is simpler than casts.
But this may not be true due to an unknown reason, so I'll leave Nuno to decide.

nunoplopes: The roudtz() value is already computed, so you'd better use it rather than creating 2 extra comparisons.

PR #775, opened 2022-02-21
Fix LLVM build steps

pauloue:


regehr:
haha how'd we miss that

PR #782, opened 2022-04-02
Fix build with recent clang

LebedevRI:
Please do not kill the messenger.

Yes, this very much looks like insanity,
yet https://reviews.llvm.org/D119670 / https://github.com/llvm/llvm-project/commit/70b1f6de539867353940d3dcb8b25786d5082d63
i think basically implies that the former will be forbidden
in future versions of the standard.

In the end, i just want to be able to build alive to check my transform :)


nunoplopes:
Oh dear.. dictatorship is coming..
Thanks for the patch! I'll merge it now.

LebedevRI:
Sorry for bad news :(

PR #789, opened 2022-05-17
support the smul_fix intrinsic

regehr:
hi Nuno-- this needs more tests and it's only one of the intrinsics but I wanted to run a sanity check by you before I put too much more work into this, could you take a quick look and see if I've changed everything that I need to change to add a new instruction, etc? thanks...


ir/instr.cpp
```diff
@@ -1191,20 +1192,28 @@ StateValue TernaryOp::toSMT(State &s) const {
...
     break;
   case FShr:
-    fn = expr::fshr;
+    e = expr::fshr(a.value, b.value, c.value);
+    np = true;
+    break;
+  case SMulFix: {
+    e = expr::smul_fix(a.value, b.value, c.value);
+    auto mul_result = a.value.sext(a.bits()) * b.value.sext(a.bits());
+    np = mul_result == mul_result.trunc(a.bits()).sext(a.bits());
```
nunoplopes, line 1210: This looks like a vanilla overflow check. If so, you can use `mul_no_soverflow`.

nunoplopes (commented):
looks good to me.

regehr:
thanks. I'll close this and come back with a more complete patch

PR #795, opened 2022-05-24
Add optimization parameters specified in alive-tv

Hatsunespica:
Basic:

By using the same syntax as  LLVM "opt"
Instead of specifying -newgvn, and -licm, now we can use -passes="newgvn,loop-mssa(licm)" to invoke passes. Besides, alive-tv now can invoke any passes supported by opt. 

File added:
LLVMOptimizer.cpp

How it is implemented:
using the same syntax as opt, call PassBuilder::parsePassPipeline to get the ModulePassManager. Run on the module finally.

Possible improvement: 
Now because it doesn't have public isFunctionPassName() method, it's hard to implement the "optimizeFunction". Use "optimizeModule" instead


CMakeLists.txt
```diff
@@ -265,3 +266,4 @@ add_custom_target("check"
 if (BUILD_TV)
   add_dependencies("check" "alive-tv" "tv")
 endif()
+
```
nunoplopes, line 269: remove line

tools/alive-tv.cpp
```diff
@@ -58,6 +60,12 @@ llvm::cl::opt<std::string> opt_tgt_fn(LLVM_ARGS_PREFIX"tgt-fn",
   llvm::cl::desc("Name of tgt function (without @)"),
   llvm::cl::cat(alive_cmdargs), llvm::cl::init("tgt"));
 
+llvm::cl::opt<string>
+    optPass(LLVM_ARGS_PREFIX "passes",
+            llvm::cl::value_desc("optmization passes"),
```
nunoplopes, line 65: typo: optimization

tools/LLVMOptimizer.cpp
```diff
@@ -0,0 +1,32 @@
...
+LLVMOptimizer::LLVMOptimizer(std::string optArgs) {
+  this->optArgs = optArgs;
+  PB.registerModuleAnalyses(MAM);
+  PB.registerCGSCCAnalyses(CGAM);
+  PB.registerFunctionAnalyses(FAM);
+  PB.registerLoopAnalyses(LAM);
+  PB.crossRegisterProxies(LAM, FAM, CGAM, MAM);
+  
+  if(optArgs=="O2"){
+    MPM = PB.buildPerModuleDefaultPipeline(OptimizationLevel::O2);
```
nunoplopes, line 15: what about O1, O0, Oz, etc? Can't parsePassPipeline handle these?

nunoplopes (changes requested):
Good start, but please read the comments inline.
Also, could you please add test running a couple of optimizations? You can try to use `; CHECK: foo` to check that the output has the optimized function.

regehr:
Yuyou, I think the reason the CI builds are all broken is that Alive2 is designed to build either with our without LLVM, but you ignored this and made it look for LLVM header files when it's trying to build in standalone mode. Look more closely at the CMake stuff and you'll see how the LLVM stuff is protected by conditionals.

regehr:
Please always match the existing code formatting when doing this sort of work. In addition to what Nuno asked for, you want to have 4 more spaces in code like `if(optArgs=="O2"){`

tools/LLVMOptimizer.cpp
Hatsunespica: https://github.com/llvm/llvm-project/blob/main/llvm/tools/opt/opt.cpp#L794

LLVM opt uses if-else to take account teach scenarios. I can do the same thing here. What do you think?

tools/alive-tv.cpp
```diff
@@ -238,7 +245,9 @@ bool compareFunctions(llvm::Function &F1, llvm::Function &F2,
 }
 
 void optimizeModule(llvm::Module *M) {
-  llvm::LoopAnalysisManager LAM;
+  LLVMOptimizer optimizer(optPass);
+  optimizer.optimizeModule(M);
+  /*llvm::LoopAnalysisManager LAM;
```
regehr, line 250: probably don't comment out blocks of code like this -- either leave it as live code if we might want it or else remove it

tools/LLVMOptimizer.cpp
```diff
@@ -0,0 +1,32 @@
+#include "LLVMOptimizer.h"
+
+using namespace llvm;
+
```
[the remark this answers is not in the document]
nunoplopes: The drawback is the extra cost. But if you will reuse this code, then it's ok.

tools/LLVMOptimizer.h
```diff
@@ -0,0 +1,52 @@
...
+  llvm::FunctionAnalysisManager FAM;
+  llvm::CGSCCAnalysisManager CGAM;
+  llvm::ModuleAnalysisManager MAM;
+  llvm::PassBuilder PB;
+
+  llvm::FunctionPassManager FPM;
+  llvm::ModulePassManager MPM;
+
+public:
+  LLVMOptimizer(std::string optArgs);
```
nunoplopes, line 46: do we need all this OO interface or is it an overkill? The current usage creates an object, calls 1 function, an throws the object away. So the OO seems overkill.

tools/LLVMOptimizer.cpp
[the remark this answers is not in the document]
Hatsunespica: yes, I will use the code >_<

tools/LLVMOptimizer.h
Hatsunespica: Yes.. I think you are right. This class won't have too many extensions and two functions are enough.

Hatsunespica: How about static functions inside a class? or just functions

nunoplopes: what about a single function: `void optimize_module(llvm::Module *M)`?
(plus an `optimize_function` if you need it for your uses)

Hatsunespica: OK, I can do that. The return value is just for convenience.

tools/alive-tv.cpp
Hatsunespica: fixed!

tools/LLVMOptimizer.h
Hatsunespica: fixed

tools/alive-tv.cpp
Hatsunespica: fixed

tools/alive-tv.cpp
```diff
@@ -15,6 +16,7 @@
 #include "llvm/IR/Function.h"
 #include "llvm/IR/Module.h"
 #include "llvm/IR/LegacyPassManager.h"
+#include "llvm/IR/LegacyPassNameParser.h"
```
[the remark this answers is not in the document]
Hatsunespica: fixed

tools/LLVMOptimizer.h
```diff
@@ -0,0 +1,55 @@
...
+#include "llvm/Support/PrettyStackTrace.h"
+#include "llvm/Support/Signals.h"
+#include "llvm/Support/SourceMgr.h"
+#include "llvm/Transforms/IPO/PassManagerBuilder.h"
+#include "llvm/Transforms/Scalar/LICM.h"
+#include "llvm/Transforms/Scalar/NewGVN.h"
+#include "llvm/Transforms/Utils/Cloning.h"
+#include "llvm/Transforms/Utils/FunctionComparator.h"
+
+#include <algorithm>
```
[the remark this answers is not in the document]
Hatsunespica: fixed

Hatsunespica:
I ran with "ninja check" and all passes. I am checking how do add cases with different -passes parameters

nunoplopes:
> I ran with "ninja check" and all passes. I am checking how do add cases with different -passes parameters

buildbots are still failing.

tools/LLVMOptimizer.cpp
```diff
@@ -0,0 +1,75 @@
...
+llvm::Module *optimize_module(std::string optArgs, llvm::Module *M) {
```
nunoplopes, line 5: this is a lot of code duplication. You need to refactor this preamble into a static function.

tools/LLVMOptimizer.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+#include "llvm/IR/Function.h"
```
nunoplopes, line 2: none of these includes are needed

tools/LLVMOptimizer.h
```diff
@@ -0,0 +1,14 @@
...
+#include "llvm/IR/LegacyPassManager.h"
+#include "llvm/IR/Module.h"
+#include "llvm/IR/Verifier.h"
+#include "llvm/Passes/PassBuilder.h"
+#include "llvm/Support/Error.h"
+
+#include <string>
+
+
+llvm::Module *optimize_module(std::string optArgs,llvm::Module *M);
```
nunoplopes, line 13: this should take const char* as input to avoid making string copies.

nunoplopes (changes requested):

tools/LLVMOptimizer.h
Hatsunespica: I used llvm::StringRef for comparison convenience. Do you think it is acceptable?

tools/LLVMOptimizer.h
Hatsunespica: I rechecked and removed most duplicate headers.

tools/LLVMOptimizer.cpp
Hatsunespica: 0 0 Now my current implementation is optimize_function calls optimize_module. Do you think if I should remove optmize_function with only optimize_module remains?

tools/LLVMOptimizer.cpp
```diff
@@ -0,0 +1,42 @@
...
+  } else {
+    if (auto E = PB.parsePassPipeline(MPM, optArgs)) {
+      llvm::errs() << E << "\n";
+    }
+  }
+  MPM.run(*M, MAM);
+}
+
+void optimize_function(llvm::StringRef optArgs, llvm::Function *func) {
+  optimize_module(optArgs, func->getParent());
```
nunoplopes, line 41: this is wrong. it will optimize the whole module. Plus it is never used in our codebase. Do you need it?

tools/LLVMOptimizer.h
```diff
@@ -0,0 +1,5 @@
+#include "llvm/Passes/PassBuilder.h"
+#include "llvm/Support/Error.h"
+
+void optimize_module(llvm::StringRef optArgs,llvm::Module *M);
```
nunoplopes, line 4: fix spacing here

tools/LLVMOptimizer.h
```diff
@@ -0,0 +1,5 @@
+#include "llvm/Passes/PassBuilder.h"
```
nunoplopes, line 1: You don't need these includes. Plus add `#pragma once` & license header.

nunoplopes (changes requested):

tools/LLVMOptimizer.cpp
Hatsunespica: You are right. This function is only necessary inside my mutate tool. I don't think Alive2 needs it. Maybe we could add it back in a better way once we need it.

tools/LLVMOptimizer.h
Hatsunespica: If I delete "PassBuilder.h" I can't build it locally. 
I suspect it's my CMakeList.txt problem. Let me check it.

Hatsunespica: At the error message, it says 
"/home/runner/work/alive2/alive2/tools/LLVMOptimizer.h:1:10: fatal error: 'llvm/Passes/PassBuilder.h' file not found
#include "llvm/Passes/PassBuilder.h""

I don't know it can't find the header. Is it any problem that I should specify something in the Cmake.txt?

tools/LLVMOptimizer.cpp
Hatsunespica: By the way, I know it is optimizing whole module, but I can't find a proper way to implement it. T_T So I just decide let it to optimize whole module instead.

tools/LLVMOptimizer.h
nunoplopes: You just need to do a forward declaration. These are only pointers.

Hatsunespica: Hi Nuno, I am sorry. I don't think I get you... In my newest push, I removed function and added #pragma once and license. >_<

Hatsunespica: Oh Wait, better idea: This file is LLVM related so I should place it in /llvm_util/ rather /tool/
What do you think?

regehr:
@Hatsunespica here is an example of the sort of forward declaration Nuno is talking about 

https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Coroutines/CoroInternal.h#L19-L21

often you can do this instead of including a header file, this can lead to much faster compile times

nunoplopes:
> Now I removed the header, and use extern function instead at alive-tv.cpp. It passed all ninja checks!

You can't do that. That's not a good practice.
Also please respect file naming conventions, and sorting of lists.

Hatsunespica:
> > Now I removed the header, and use extern function instead at alive-tv.cpp. It passed all ninja checks!
> 
> You can't do that. That's not a good practice. Also please respect file naming conventions, and sorting of lists.

oh, I thought you asked me "You just need to do a forward declaration. These are only pointers." That's why I did this update. So what I should do is back to its previous version and add some tests?

nunoplopes:
> > > Now I removed the header, and use extern function instead at alive-tv.cpp. It passed all ninja checks!
> > 
> > 
> > You can't do that. That's not a good practice. Also please respect file naming conventions, and sorting of lists.
> 
> oh, I thought you asked me "You just need to do a forward declaration. These are only pointers." That's why I did this update. So what I should do is back to its previous version and add some tests?

https://en.wikipedia.org/wiki/Forward_declaration#Classes
https://stackoverflow.com/questions/4757565/what-are-forward-declarations-in-c

Hatsunespica:
> > > > Now I removed the header, and use extern function instead at alive-tv.cpp. It passed all ninja checks!
> > > 
> > > 
> > > You can't do that. That's not a good practice. Also please respect file naming conventions, and sorting of lists.
> > 
> > 
> > oh, I thought you asked me "You just need to do a forward declaration. These are only pointers." That's why I did this update. So what I should do is back to its previous version and add some tests?
> 
> https://en.wikipedia.org/wiki/Forward_declaration#Classes https://stackoverflow.com/questions/4757565/what-are-forward-declarations-in-c

Thanks  for your explanation. T_T. This is my first time to submit pull a request patch and thanks for your patient. I am adding test cases right now. What do you think of this version?

tools/alive-tv.cpp
```diff
@@ -269,6 +255,8 @@ llvm::Function *findFunction(llvm::Module &M, const string &FName) {
 }
 }
 
+void optimizeModule(llvm::Module *M, llvm::StringRef optArgs);
```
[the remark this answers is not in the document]
Hatsunespica: Aha I see what you mean! Thanks for the explanation!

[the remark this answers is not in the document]
Hatsunespica: should I add the function into llvm_util namespace?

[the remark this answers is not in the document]
nunoplopes: yes, thanks!

You can't do it with StringRef because it is passed by value. Alternatively, you can use C++'s string_view. We don't usually use LLVM's stuff so we stay as independent as possible.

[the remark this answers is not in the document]
Hatsunespica: Done!

Hatsunespica:
Hi Nuno, do you have any suggestions on where should I place the test cases?

PR #817, opened 2022-07-08
save IR to disk in the report directory from the plugin, if asked to

regehr:
the strategy is to save bitcode from the child process, meaning that the same module could end up being saved multiple times. we finesse this by writing to a temp file and then atomically renaming it.

I don't have a windows machine available to test this but I'll watch the buildbots...

the next part of this patchset will record command line options -- at that point we should have enough information to automatically run llvm-reduce on errors found by the plugin

### DIFF
--- a/llvm_util/cmd_args_list.h
+++ b/llvm_util/cmd_args_list.h
@@ -130,6 +130,11 @@ llvm::cl::opt<string> opt_report_dir(LLVM_ARGS_PREFIX "report-dir",
 bool report_dir_created = false;
 string report_filename;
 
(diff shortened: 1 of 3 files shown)


regehr:
hmmm... the use of tmpnam() elicits a linker warning on my linux box, let me know if I should just implement our own tmpnam, it's not like that's hard...

nunoplopes:
This patch stores the IR after the broken optimization AFAIU. We need to copy the module before each optimization just in case when the save-ir flag is on.

regehr:
ack, thanks

PR #820, opened 2022-07-19

regehr:
hopefully this patch is pretty obvious

I also made a related change which is to suppress `-` and `<stdin>` in generated file names, these are just a pain since their existence can break command line tools

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -13,12 +13,14 @@
 #include "llvm/ADT/Any.h"
 #include "llvm/ADT/Triple.h"
 #include "llvm/Analysis/TargetLibraryInfo.h"
+#include "llvm/Bitcode/BitcodeWriter.h"
 #include "llvm/IR/LegacyPassManager.h"
 #include "llvm/IR/Module.h"
 #include "llvm/Pass.h"
 #include "llvm/Passes/PassBuilder.h"
 #include "llvm/Passes/PassPlugin.h"
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/Transforms/Utils/Cloning.h"
 #include <fstream>
 #include <iostream>
 #include <memory>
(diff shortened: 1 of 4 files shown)


regehr:
(I can figure out how to test this if you want, but since this is probably only used by us, I'm fine for it to not be in the unit tests. I did test and make sure that the saved IR + the saved command line do reproduce soundness failures)

tv/tv.cpp
```diff
@@ -168,6 +194,9 @@ struct TVLegacyPass final : public llvm::ModulePass {
       return false;
     }
 
+    if (!MClone && first && opt_save_ir && !report_filename.empty())
```
nunoplopes, line 197: I think this condition is only true when you run the first optimization and for the first function.
If I run `-O2` and the bug is somewhere in the middle, the original IR (before any optimization) will be written to disk I think.

regehr: ok so we have to keep saving additional versions each time?

I wrote it like that to avoid cloning the module for every function but I'll find a better way to do that

nunoplopes: I think it's fine to keep cloning. I think we'll only enable the save-ir flag when we know there's a bug, so it's fine to slow things down a bit.

regehr: so you're thinking we we-run alive2 with this option after finding a crash? I was hoping to just leave this turned on all the time while bug-hunting...

regehr: but of course re-running is easy since we have the exact command line. maybe that's a better idea. let me play around with this some more.

nunoplopes: We need an option that allows to dump the IR right before the buggy optimization + the name of that optimization. That's what we need to reproduce the issue. (we may need some extra pass arguments, but those are probably not easy to recover automatically. Though a human can grab those easily from the overall cmd line I think)

regehr: well, alternatively we can just log the command line and also the original IR and then narrow things down more later on as part of reducing

PR #821, opened 2022-07-20

regehr:
ok this time I've included a silly perl script that reproduces all ~57 unsoundnesses that happen when running over the LLVM unit tests.

Nuno I would argue that this approach (saving the original IR, and not trying to grab it just before the unsound transform) is the correct one. or if I'm wrong, at least this is simpler and we should go with it for now.

you'll notice an apparently spurious reset of MClone in `finalize()`, this is necessary the prevent a crash in the module destructor, I don't know what's going on but I don't think it's our bug (Yuyou noticed some bugs in MClone where it does not completely separate the new module from the old one -- this could lead to a crash if the original module is destructed before the cloned one)

### DIFF
--- a/llvm_util/cmd_args_def.h
+++ b/llvm_util/cmd_args_def.h
@@ -39,6 +39,8 @@ if (!report_dir_created && !opt_report_dir.empty()) {
   if (!opt_overwrite_reports) {
     do {
       auto newname = fname.stem();
+      if (newname.compare("-") == 0 || newname.compare("<stdin>") == 0)
+        newname = "";
       newname += "_" + get_random_str(8) + ".txt";
       path.replace_filename(newname);
     } while (fs::exists(path));
--- a/llvm_util/cmd_args_list.h
+++ b/llvm_util/cmd_args_list.h
@@ -128,7 +128,12 @@ llvm::cl::opt<string> opt_report_dir(LLVM_ARGS_PREFIX "report-dir",
   llvm::cl::cat(alive_cmdargs));
 
 bool report_dir_created = false;
-string report_filename;
+fs::path report_filename;
+
+llvm::cl::opt<bool> opt_save_ir(LLVM_ARGS_PREFIX "save-ir",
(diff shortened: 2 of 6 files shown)


nunoplopes:
But why is it useful to store the original IR? That's easily available.
If you use `opt`, then that's the input file.
If you use `clang`, you can get the original IR by changing just the cmd line.

The hard IR to get is the one right before the buggy optimization. Even if we were to store the original IR, how would you get the that repro IR? That's the missing functionality I think.
We want some IR such that I can do `opt -S -some-optimation file.ll` and that exposes the bug. Then we can minimize that `file.ll`, for example.

Unless you have other plans in mind?

regehr:
the original IR is not convenient to grab when we're using alivecc to build some arbitrary 3rd party application, this requires someone to go look at the commands issued during the build, etc. we never want to interact with build systems other than at the most basic level such as `CC=alivecc`

what you seem to be asking for is "IR that is directly miscompiled by a pass" whereas what this patch saves is "IR that is miscompiled by a pass, but possibly requiring other passes to run first." but this should be more or less solved by my llvm-reduce pass that runs opt passes.

I'm just trying to keep things simple here, and for example I don't know how to get the name of the pass that causes the miscompile if we do this your way, the -O2 pipeline is long and how do you want to narrow down which one caused the problem?

nunoplopes:
> the -O2 pipeline is long and how do you want to narrow down which one caused the problem?

But we need to solve this problem. No one wants an IR file that is miscompiled by -O2. The best bug report is a reduced IR file that is miscompiled with a specific optimization.

I think we receive the pass names in the plugin as they are executed (from the new? pass manager). That's how we skip the IPO passes. It's a matter of saving the name somewhere, if it's not saved already.

My intuition is that duplicating the module should be peanuts compared to the time taken by Z3 🙂

regehr:
I'm running out of time for this but might give it another try. do you have a pointer to the saved pass names? the only thing I've seen is on opt-alive.sh but that's not useful without adding some more plumbing

PR #823, opened 2022-07-24

regehr:
want to look this over Nuno? it works for a few examples I tried (in the sense that it dumps a bitcode file that repros the original error using the pass that's printed) but I don't want to polish if it's not what you're looking for here

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -13,12 +13,14 @@
 #include "llvm/ADT/Any.h"
 #include "llvm/ADT/Triple.h"
 #include "llvm/Analysis/TargetLibraryInfo.h"
+#include "llvm/Bitcode/BitcodeWriter.h"
 #include "llvm/IR/LegacyPassManager.h"
 #include "llvm/IR/Module.h"
 #include "llvm/Pass.h"
 #include "llvm/Passes/PassBuilder.h"
 #include "llvm/Passes/PassPlugin.h"
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/Transforms/Utils/Cloning.h"
 #include <fstream>
 #include <iostream>
 #include <memory>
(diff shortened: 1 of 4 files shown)


tv/tv.cpp
```diff
@@ -616,32 +649,40 @@ llvmGetPassPluginInfo() {
...
+          bool do_finish = TVPass::batch_started && !do_skip(pass_name)
               && do_skip(P);
 
           if (do_start)
-            TVPass::batched_pass_begin_name = TVPass::pass_name;
+            TVPass::batched_pass_begin_name = pass_name;
           else if (is_first)
             TVPass::batched_pass_begin_name = "beginning";
 
+          if ((is_first || do_start) && opt_save_ir && !report_filename.empty())
```
nunoplopes, line 663: I think this bit is not needed.
For Alive2 verification we have to do this because the callback is after each pass, but since you are running a before callback, it shouldn't be needed. (running the before callback is the right choice for cloning the module)

nunoplopes:
It looks good to me, thank you!
Just left a comment of something that I think that can be removed. Please double check.

tv/tv.cpp
regehr: ok let's talk through this a bit...

by "this bit" do you mean the `(is_first || do_start)` check? if so, I think these are needed because we want to save the IR at the start of the entire batch of passes, instead of saving it before each pass in the batch. 

there's probably a bit more work to do here where we record the names of all passes in the batch so we can replay them in that order, to repro any bugs that get found, but I don't want to deal with that right now.

or perhaps we don't need to save IR at all for the batched case, since a bug found with batching will generally want to be reproduced with batching turned off?

nunoplopes: Oh, sorry, I didn't realize this was in the big if for batched ops.
Ignore my comment then.

Can I merge this PR?

regehr: yes please merge

PR #836, opened 2022-08-23
add an external cache

regehr:
this commit gives alive-tv and the opt pass the capacity to not
perform a refinement check that has been performed before, by asking
an external redis server if it has seen a checksum of the src/tgt
alive IR before


regehr:
there were a bunch of moving parts here and I forgot to put the CMake part of this behind BUILD_LLVM_UTILS, let me do that

regehr:
ok, the cache stuff is now only built along with the rest of the LLVM stuff

PR #845, opened 2022-09-20
forgot to close a file that I had opened, plus cosmetic changes

regehr:


regehr:
(closing the file actually matters since the jobserver is potentially long-lived)

PR #847, opened 2022-09-26
strip out flags that would cause clang to add debug info

regehr:


regehr:
I'm not 100% sure that stripping out every compiler flag starting with `-g` is the answer but it has worked well enough for a few things I tried it on, and also this looked ok based on scanning the clang `-help` output

PR #848, opened 2022-09-26
avoid using LLVM's CloneModule

regehr:
let's just avoid using CloneModule, not only is it slow and leaky, but also it contains a couple of correctness bugs

### DIFF
--- a/tv/tv.cpp
+++ b/tv/tv.cpp
@@ -21,7 +21,6 @@
 #include "llvm/Passes/PassBuilder.h"
 #include "llvm/Passes/PassPlugin.h"
 #include "llvm/Support/raw_ostream.h"
-#include "llvm/Transforms/Utils/Cloning.h"
 #include <fstream>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


tv/tv.cpp
```diff
@@ -326,8 +325,8 @@ struct TVLegacyPass final : public llvm::ModulePass {
         has_failure = true;
         *out << "\nPass: " << pass_name << '\n';
         emitCommandLine(out);
-        if (MClone)
-          writeBitcodeAtomically(report_filename);
+	if (SavedBitcode.size() > 0)
```
nunoplopes, line 328: This is writing the bitcode after the wrong transformation only. We need the bitcode before, which is pretty annoying. You would need to write it to disk all the times. And don't overwrite it later.

regehr: please look closer, the call at line 329 is writing to disk some bitcode that we saved into a memory buffer (a std::string) earlier

PR #869, opened 2022-12-25
add a few utility functions that we had to write to support the ARM lifter

regehr:


ir/function.cpp
```diff
@@ -223,6 +228,10 @@ void Function::addInput(unique_ptr<Value> &&i) {
   inputs.emplace_back(std::move(i));
 }
 
+void Function::addInputAt(std::unique_ptr<Value> &&c, unsigned idx) {
```
nunoplopes, line 231: can we rename this as replaceInput or something like that?  Because it cannot add a new input; it must replace an existing one.

PR #874, opened 2023-01-02
fix the build in the case where Redis is installed on the system

regehr:


regehr:
or let me know if you want me to fix this a different way, but the cache is mentioned in `cmd_args_def.h` so we either have to change how that works or else declare a cache in each tool that includes that file

nunoplopes:
oops. I need to check why the redis buildbots didn't fail 🤨

PR #902, opened 2023-04-30
Enable multiple 'src'/'tgt' pairs in a single source

goldsteinn:
Change:
Instead of only searching for 'src' and 'tgt' in single-file mode, search for an arbitrary number of pairs of 'src{+d}', 'tgt{+d}'.

I.e if have the following functions in a file:

```
define @src {}
define @tgt {}
define @src2 {}
define @tgt2 {}
define @src3 {}
define @tgt3 {}
define @src4 {}
define @tgt5 {}
```

Alive2 will have the following pairs: 'src'/'tgt', 'src2'/'tgt2', and 'src3'/'tgt3'. It will ignore 'src4' and 'tgt5'.

Motivation:
This just a QOL change. A lot of LLVM patches have assosiated alive2 links. In some cases there need to be many links (i.e https://reviews.llvm.org/D149521).

Being able to specify multiple 'src'/'tgt' pairs in a single file would enable verifying multiple cases with a single link making it easier.

### DIFF
--- a/tests/alive-tv/ssub-cmp-zero.v0.srctgt.ll
+++ b/tests/alive-tv/ssub-cmp-zero.v0.srctgt.ll
@@ -1,0 +1,44 @@
+declare i8 @llvm.ssub.sat.i8(i8, i8)
+define i1 @src(i8 %x, i8 %y) {
+  %m = call i8 @llvm.ssub.sat.i8(i8 %x, i8 %y)
+  %r = icmp eq i8 %m, 0
+  ret i1 %r
+}
+
+define i1 @tgt(i8 %x, i8 %y) {
+  %r = icmp eq i8 %x, %y
+  ret i1 %r
+}
+
+define i1 @src1(i8 %x, i8 %y) {
+  %m = call i8 @llvm.ssub.sat.i8(i8 %x, i8 %y)
+  %r = icmp ne i8 %m, 0
+  ret i1 %r
+}
+
+define i1 @tgt1(i8 %x, i8 %y) {
+  %r = icmp ne i8 %x, %y
+  ret i1 %r
+}
+
+define i1 @src2(i8 %x, i8 %y) {
+  %m = call i8 @llvm.ssub.sat.i8(i8 %x, i8 %y)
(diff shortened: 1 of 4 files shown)


regehr:
it's Nuno who makes decisions about this sort of thing, but I help out and would be happier if you added a unit test or two that checks that this functionality works as expected, thanks!

goldsteinn:
> it's Nuno who makes decisions about this sort of thing, but I help out and would be happier if you added a unit test or two that checks that this functionality works as expected, thanks!

Added a test for the new functionality.

goldsteinn:
> Looks good to me, thanks!
> 
Thought of this while writing the negative test. Would you consider instead of just `src{+d}` and `tgt{+d}` matching on `src_*` and `tgt_*`. The rationale is it make so that we could describe the proof with the matching postfix i.e

```
define i8 @src_incorrect(i8 %x) {
  %m = sub i8 %x, -1
  ret i8 %m
}

define i8 @tgt_incorrect(i8 %x) {
  %m = xor i8 %x, -1
  ret i8 %m
}
define i8 @src_verifies(i8 %x) {
  %m = sub i8 -1, %x
  ret i8 %m
}

define i8 @tgt_verifies(i8 %x) {
  %m = xor i8 %x, -1
  ret i8 %m
}

define i8 @src_error_mismatch_sig(i8 %x, i8 %y) {
  %m = add i8 %y, %x
  ret i8 %m
}

define i8 @tgt_error_mismatch_sig(i8 %x) {
  %m = xor i8 %x, -1
  ret i8 %m
}
```

instead of just arbitrary numbering.
What do you think?
> Could you please rebase because we had an issue with github's new images? Also, could you add a test with a buggy transformation? Thanks!

Done (ish). Not sure if I set it up right so you probably want to take a look.

regehr:
I like the string approach better!

goldsteinn:
> I like the string approach better!

Implemented + updated one of the tests.
(accidentally closed resetting my fork).

regehr:
this functionality should now be available in the CE instance, please feel free to start linking there (or let me know if anything's amiss)

Issue #748, opened 2021-09-20
Bug in unroller with nested loop


nunoplopes:
Reduced test case from @regehr:
```llvm
define void @test() {
entry:
  %i = alloca i32, align 4
  %j = alloca i32, align 4
  store i32 0, i32* %i, align 4
  br label %for.body

for.body:
  br label %for.body4

for.body4:
  br label %for.inc

for.inc:
  %n7 = load i32, i32* %j, align 4
  %inc = add i32 %n7, 1
  store i32 %inc, i32* %j, align 4
  %cmp2 = icmp ult i32 %inc, 2
  br i1 %cmp2, label %for.body4, label %for.inc5

for.inc5:
  %n9 = load i32, i32* %i, align 4
  %inc6 = add i32 %n9, 1
  store i32 %inc6, i32* %i, align 4
  %cmp = icmp ult i32 %inc6, 2
  br i1 %cmp, label %for.body, label %for.end7

for.end7:
  ret void
}
```

Equivalent C:
```C
void test() {
  for (unsigned i = 0; i < 2; i++) {
    for (unsigned j = 0; j < 2; j++) { }
  }
}
```

Repro:
```
$ opt -S -jump-threading bug.ll -o optimized.ll
$ alive-tv -src-unroll=2 -tgt-unroll=2 bug.ll optimized.ll

ERROR: The target program doesn't reach a return instruction.
```

regehr:
I have asked @zhengyang92 to look into this one

zhengyang92:
To eliminate the unnecessary jumps, two nested loops may share a common header. In this reduced example, the jump-threading pass turns the loops that have separate headers into loops sharing one common header. The Havlak's algorithm we're using recognizes such CFG as a single loop rather than two, and this causes the missed unroll of the inner loop. The LLVM loop analysis passes have similar problems. These passes first canonicalizes the loops back to loops using separated headers and then performs the analysis.

@nunoplopes  What should we do to handle such cases? I guess we need to implement a loop canonicalizer that runs before the unroller.

regehr:
@zhengyang92 I guess the canonicalization code you're talking about is in `Transforms/Utils/LoopSimplify.cpp`?
do you have a sense for how hard it will be to do this in alive?

zhengyang92:
[...]

PR #908, opened 2023-05-18
Take loop exit blocks into account in constructing edges for top sort when dealing with a loop header

jkubik101:
This patch fixes a [failing example](https://github.com/AliveToolkit/alive2/issues/748#issuecomment-1372248300) mentioned in issue [#748](https://github.com/AliveToolkit/alive2/issues/748). The original issue had likely gotten "fixed" by accident before this patch, so I only refer to that specific example, but it affects any code with nested loops (though sometimes the problem may not manifest itself).

Here is the reduced example:

```llvm
define i32 @partial_sub_loop_test_branch_loop(i32 %end) {
entry:
  br label %outer_header

outer_header:
  %j = phi i32 [0, %entry], [%j.inc, %outer_backedge]
  br label %preheader

preheader:
  br label %header

header:
  %i = phi i32 [0, %preheader], [%i.inc, %backedge]
  br label %backedge

backedge:
  %i.inc = add i32 %i, 1
  %cmp = icmp slt i32 %i.inc, %end
  br i1 %cmp, label %header, label %outer_backedge

outer_backedge:
  %j.inc = add i32 %j, 1
  %cmp.j = icmp slt i32 %j.inc, %end
  br i1 %cmp.j, label %outer_header, label %exit

exit:
  ret i32 %i.inc
}
```
And this is the CFG:

bug-cfg

Running `alive-tv bug.ll -src-unroll=2 -tgt-unroll=2 --passes=lcssa` currently produces a false positive (i.e. it says the transformation is incorrect even though it should be correct). This is fixed by this patch.

The problem is that when you have a nested loop and the outer loop is being unrolled and you need to topologically sort the body, the inner loops are compressed into a single node (i.e. they are represented only by the header), but the outgoing edges from the loop are not preserved, so there is a chance the basic blocks will not be in correct topological order (depending on tiebreaks). This then causes problems when values are replaced in cloning BBs, because a BB can come "sooner" and have the wrong predecessors (as it has a different `vmap`).
So in this example, when running top-sort on the body of the outer loop, which contains `preheader`, `header` (representing the inner loop), and `outer_backedge`, Alive2 doesn't know that there's a transitive dependency between `header` and `outer_backedge` (due to the loop being compressed), so it puts `outer_backedge` before `header`.

I fixed this by adding an `std::set` of exit basic blocks to the `BasicBlock` class. If a `BasicBlock` is a header of some loop, this property keeps track of the exit blocks of the loop. This of course only makes sense if the given basic block is a header of a loop, the set is empty otherwise. Thus, these exit blocks can be taken into account when creating edges for top sort.
I've tried to work within the constraints of the design and intervene in the existing code as little as possible, so I'm sure the implementation details are bound to change, but it is a working solution from what I've tested and hopefully gets the idea across.

### DIFF
--- a/ir/function.h
+++ b/ir/function.h
@@ -15,8 +15,8 @@
 #include <string_view>
 #include <tuple>
 #include <unordered_map>
-#include <vector>
 #include <unordered_set>
+#include <vector>
 
 namespace smt { class Model; }
 
--- a/ir/function.h
+++ b/ir/function.h
@@ -16,6 +16,7 @@
 #include <tuple>
 #include <unordered_map>
 #include <vector>
+#include <unordered_set>
 
 namespace smt { class Model; }
 
@@ -28,7 +29,7 @@ class BasicBlock final {
   std::vector<std::unique_ptr<Instr>> m_instrs;
 
   // If the basic block is a header, this holds all exit blocks of its loop
-  std::set<BasicBlock*> exit_blocks; 
+    std::unordered_set<BasicBlock*> exit_blocks; 
 
 public:
   BasicBlock(std::string_view name) : name(name) {}
@@ -45,7 +46,7 @@ public:
   void addExitBlock(BasicBlock* bb) {
     exit_blocks.emplace(bb);
   }
-  const std::set<BasicBlock*>& getExitBlocks() const {
+  const std::unordered_set<BasicBlock*>& getExitBlocks() const {
     return exit_blocks;
   }
 
--- a/ir/function.h
+++ b/ir/function.h
@@ -43,9 +43,7 @@ public:
   void addInstrAt(std::unique_ptr<Instr> &&i, const Instr *other, bool before);
   void delInstr(const Instr *i);
 
-  void addExitBlock(BasicBlock* bb) {
-    exit_blocks.emplace(bb);
-  }
+  void addExitBlock(BasicBlock* bb);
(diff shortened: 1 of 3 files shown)


ir/function.h
```diff
@@ -39,6 +42,13 @@ class BasicBlock final {
   void addInstrAt(std::unique_ptr<Instr> &&i, const Instr *other, bool before);
   void delInstr(const Instr *i);
 
+  void addExitBlock(BasicBlock* bb) {
+    exit_blocks.emplace(bb);
```
nunoplopes, line 46: please move this to the .cpp file speedup compilation

ir/function.h
```diff
@@ -27,6 +27,9 @@ class BasicBlock final {
   std::string name;
   std::vector<std::unique_ptr<Instr>> m_instrs;
 
+  // If the basic block is a header, this holds all exit blocks of its loop
+  std::set<BasicBlock*> exit_blocks;
```
nunoplopes, line 31: this can be an unordered set. you need need to include that header above.

ir/function.h
```diff
@@ -16,6 +16,7 @@
 #include <tuple>
 #include <unordered_map>
 #include <vector>
+#include <unordered_set>
```
nunoplopes, line 19: sort this pls :)

PR #978, opened 2023-11-27
Add support for disjoint

dtcxzyw:
This patch adds support for `disjoint` flags recently introduced by https://github.com/llvm/llvm-project/pull/72583.
See also https://llvm.org/docs/LangRef.html#or-instruction.


nunoplopes:
Thank you!

PR #991, opened 2023-12-19
unconditionally check for source functions that are UB for all executions, but don't error out

regehr:
I think this is a really important usability thing for Alive users


regehr:
```
Johns-MacBook-Pro:build regehr$ ./alive-tv test.ll 

define i16 @f(i16 %#0) {
#1:
  %x = sdiv i16 %#0, 0
  ret i16 %x
}
=>
define i16 @f(i16 %#0) nofree willreturn memory(none) {
#1:
  ret i16 poison
}

****************************************
WARNING: Source function is always UB.
It can be refined by any target function.
Please make sure this is what you wanted.
****************************************

Summary:
  1 correct transformations
  0 incorrect transformations
  0 failed-to-prove transformations
  0 Alive2 errors
Johns-MacBook-Pro:build regehr$ 
```

PR #1011, opened 2024-02-02
opt-alive.sh --verbose to stderr, hints on building with Xcode

FlashSheridan:
- Fix my 101be628 to redirect verbose `opt-alive.sh` output to stderr, since Lit can swallow it.
- Output report name to stderr in cmd_args_def.h (with a slightly different caption, and white space to make it prominent), since the output at tv.cpp:425 seems to have stopped working.  (Changing it to stderr didn’t work, and an Xcode breakpoint at that line was not hit when running under Lit, so it doesn’t seem to be an output redirection issue.)  It used to work at roughly bc51b72c.
- Add build hints for Xcode. 
  - Recommend disabling “Implicit conversion loses integer precision: 'uint64_t' (aka 'unsigned long long') to 'unsigned int'” warnings, since there are a dozen of them and they break the build with the default warnings as errors, which I’ll file [separately](https://github.com/AliveToolkit/alive2/issues/1012).
  - I didn’t think this belonged in the Alive2 ReadMe, but I’ll note here that for using Xcode breakpoints in `tv.dylib`, even when it’s called indirectly via Lit, the suggestions at https://stackoverflow.com/a/35641035/28312 may be helpful.

The output I’m adding is in executable code in a header file, so there may be uses I haven’t tested.

### DIFF
--- a/llvm_util/cmd_args_def.h
+++ b/llvm_util/cmd_args_def.h
@@ -50,7 +50,6 @@ if (!report_dir_created && !opt_report_dir.empty()) {
         newname = "in";
       newname += "_" + get_random_str(8) + ".txt";
       path.replace_filename(newname);
-      std::cerr << "\n\nWriting report to: " << newname <<  endl << endl;
     } while (fs::exists(path));
   }
 
--- a/llvm_util/cmd_args_def.h
+++ b/llvm_util/cmd_args_def.h
@@ -50,6 +50,7 @@ if (!report_dir_created && !opt_report_dir.empty()) {
         newname = "in";
       newname += "_" + get_random_str(8) + ".txt";
       path.replace_filename(newname);
+      std::cerr << "\n\nWriting report to: " << newname <<  endl << endl;
     } while (fs::exists(path));
   }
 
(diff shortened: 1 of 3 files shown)


llvm_util/cmd_args_def.h
```diff
@@ -50,6 +50,7 @@ if (!report_dir_created && !opt_report_dir.empty()) {
         newname = "in";
       newname += "_" + get_random_str(8) + ".txt";
       path.replace_filename(newname);
+      std::cerr << "\n\nWriting report to: " << newname <<  endl << endl;
```
nunoplopes, line 53: we can't do this. This will break the LLVM unit test suite. We can only add output when the test fails.

FlashSheridan: > This will break the LLVM unit test suite. We can only add output when the test fails.

Writing to stderr rather than stdout was meant to avoid that.  The Lit tests I tried before submitting the PR all passed, but the whole Lit suite overnight on Linux got a failure rate of 0.38% (193/50546).  Some of the failures passed in isolation (e.g. pr25342.ll); one which didn’t (logical-right-shift-until-zero.ll) also failed without my change.  
    I may no longer need the report name in the log for cases in which it’s not output (presumably with `has_failure` false in tv.cpp:424), so should I revert that part of the PR?
    Redirecting verbose `opt-alive.sh` output to stderr still seems worth fixing (I warned about it in the ReadMe in 3b631cbe, and this should reduce the problem), and the Xcode hints may be helpful at least for some of us who left Apple.

nunoplopes: Yes, this bit needs to be reverted.
I know for a fact that it will break the LLVM test suite. That's why we only print the report file name on failure. There are tests that grep for stuff on stderr.

Issue #1023, opened 2024-03-26
Add support for poison-generating flags `nuw`/`nsw` to `trunc`


antoniofrighetto:
As per https://github.com/llvm/llvm-project/pull/85592/commits/d1fb0cd61f5c4aa4373be8d1be3ab18d501e0466.

PR #1024, opened 2024-03-27
Add support for nowrap flags to `trunc`

antoniofrighetto:
Nowrap poison-generating flags to `trunc` have been recently introduced in LLVM.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -1507,6 +1507,21 @@ unique_ptr<Instr> TestOp::dup(Function &f, const string &suffix) const {
   return make_unique<TestOp>(getType(), getName() + suffix, *lhs, *rhs, op);
 }
 
+ConversionOp::ConversionOp(Type &type, std::string &&name, Value &val, Op op,
+                           unsigned flags)
+    : Instr(type, std::move(name)), val(&val), op(op), flags(flags) {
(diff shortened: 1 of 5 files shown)


ir/instr.cpp
```diff
@@ -1558,8 +1577,17 @@ StateValue ConversionOp::toSMT(State &s) const {
...
+      AndExpr non_poison;
+      unsigned orig_bits = val.bits();
+      unsigned trunc_bits = to_type.bits();
+      if (flags & NUW)
+        non_poison.add(val.trunc(trunc_bits).zext(orig_bits - trunc_bits) ==
+                       val);
+      if (flags & NSW)
+        non_poison.add(val.trunc(trunc_bits).sext(orig_bits - trunc_bits) ==
+                       val);
+      return {val.trunc(trunc_bits), non_poison()};
```
nunoplopes, line 1590: I suggest don't computing trunc() 3x times by saving this in a tmp var.
These functions do pre-processing and can be expensive.

The rest of the patch is great, thanks!

nunoplopes:
fixes #1023.

ir/instr.cpp
antoniofrighetto: Definitely, thanks for the quick review! :)

PR #1046, opened 2024-05-30
remove duplicate code on cmd_args_def.h

IamYJLee:
There is redundancy in the code that sets the value of the `config::src_unroll_cnt` variable.


IamYJLee:
@nunoplopes 
Looking at the whole source code, the `#ifdef ARGS_SRC_TGT` directive itself seems unnecessary. Could you please give your opinion on this?

nunoplopes:
It's a typo. it should be setting tgt unroll factor, not src.

PR #1047, opened 2024-05-30
Add support for getelementptr nusw and nuw flags (WIP)

nikic:
I'm struggling to make this actually work, test 4 and 5 currently fail. The problem is that for these flags we can't just require no overflow on the offset, we need to consider the address. However, the offset is one bit smaller than the offset, so we don't get the correct overflow properties. Some advice would be welcome.

### DIFF
--- a/ir/pointer.cpp
+++ b/ir/pointer.cpp
@@ -7,7 +7,6 @@
 #include "ir/globals.h"
 #include "ir/state.h"
 #include "util/compiler.h"
-#include <iostream>
 
 using namespace IR;
 using namespace smt;
--- a/ir/pointer.cpp
+++ b/ir/pointer.cpp
@@ -292,11 +292,15 @@ Pointer Pointer::maskOffset(const expr &mask) const {
            getAttrs() };
 }
 
-expr Pointer::addNoSOverflow(const expr &offset) const {
-  return getOffset().add_no_soverflow(offset);
+expr Pointer::addNoUSOverflow(const expr &offset, bool offset_only) const {
+  if (offset_only)
+    return getOffset().add_no_soverflow(offset);
+  return getAddress().zextOrTrunc(offset.bits()).add_no_usoverflow(offset);
 }
 
-expr Pointer::addNoUOverflow(const expr &offset) const {
+expr Pointer::addNoUOverflow(const expr &offset, bool offset_only) const {
+  if (offset_only)
+    return getOffset().add_no_uoverflow(offset);
   return getAddress().zextOrTrunc(offset.bits()).add_no_uoverflow(offset);
 }
 
--- a/ir/pointer.cpp
+++ b/ir/pointer.cpp
@@ -7,6 +7,7 @@
 #include "ir/globals.h"
 #include "ir/state.h"
 #include "util/compiler.h"
+#include <iostream>
 
 using namespace IR;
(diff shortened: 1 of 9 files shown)


ir/pointer.cpp
```diff
@@ -292,10 +292,14 @@ Pointer Pointer::maskOffset(const expr &mask) const {
            getAttrs() };
 }
 
-expr Pointer::addNoOverflow(const expr &offset) const {
+expr Pointer::addNoSOverflow(const expr &offset) const {
   return getOffset().add_no_soverflow(offset);
 }
 
+expr Pointer::addNoUOverflow(const expr &offset) const {
+  return getAddress().zextOrTrunc(offset.bits()).add_no_uoverflow(offset);
```
nunoplopes, line 300: I think you want something like:
return getAddress().add_no_uoverflow(offset.sextOrTrunc(bits_ptr_address));

An address is never smaller than offsets. We try to shrink the number of bits of offsets as an optimization, hence they are usually not 64 bit long.
(I don't have a fresh build of LLVM to try this out -- maybe in half an hour)

nunoplopes:
I think this test case is not correct:
```llvm
define ptr @src4(ptr %x, i8 %i, i8 %j) {
  %p = gep nuw ptr %x, 1 x i8 %i, 1 x i8 %j
  ret ptr %p
}
=>
define ptr @tgt4(ptr %x, i8 %i, i8 %j) {
  %i.j = add nuw i8 %i, %j
  %p = gep nuw ptr %x, 1 x i8 %i.j
  ret ptr %p
}
```

Say that %i nd %j are two large positive numbers (e.g. 0x7f).Adding then produces a negative number.
Sine gep indexes are sign extended, tgt will add a very large number, while src added two small numbers. nuw with gep is probably only useful when indexes have the same bitwidth as pointers (alive says the transformation is correct with i64 fwiw).

nikic:
@nunoplopes The test case uses a `p:8:8` data layout, so i8 is the full pointer index size. I agree it would not be correct if the type doesn't match the pointer index size.

nunoplopes:
> @nunoplopes The test case uses a `p:8:8` data layout, so i8 is the full pointer index size. I agree it would not be correct if the type doesn't match the pointer index size.

Ups, I missed that bit. I was forcing the pointer size.

nunoplopes:
There you go:
```diff
diff --git a/tools/transform.cpp b/tools/transform.cpp
index 59d58571..ace87307 100644
--- a/tools/transform.cpp
+++ b/tools/transform.cpp
@@ -1215,7 +1215,8 @@ static void calculateAndInitConstants(Transform &t) {
   // as an (unsound) optimization, we fix the first bit of the addr for
   // local/non-local if both exist (to reduce axiom fml size)
   bool has_local_bit = (num_locals_src || num_locals_tgt) && num_nonlocals;
-  bits_ptr_address = min(max(bits_size_t, bits_ptr_address) + has_local_bit,
+  bits_ptr_address = min(max(max(bits_for_offset, bits_size_t),
+                             bits_ptr_address) + has_local_bit,
                          bits_program_pointer);

   if (config::tgt_is_asm)
```

nikic:
Thanks! With that, I think I have a working implementation now.

nunoplopes:
thank you! 🙏

PR #1049, opened 2024-06-10
Add support for ucmp/scmp

dtcxzyw:
This patch adds support for llvm.ucmp/scmp intrinsics.
See https://llvm.org/docs/LangRef.html#llvm-scmp-intrinsic

PR #1050, opened 2024-06-10
Fix bitwidth for vector ucmp/scmp

dtcxzyw:
Fixes https://github.com/AliveToolkit/alive2/pull/1049/files#r1633629082.


nunoplopes:
thanks!

PR #1051, opened 2024-06-13
chore: fix some comments

fivecut:


nunoplopes:
The `tests/lit` is not our code and should not be changed.

Issue #916, opened 2023-06-23
Examples timing out


tanmaytirpankar:
Hi,

The following examples and some similar ones are timing out locally and on the online instance. Is this expected? Adding the --disable-undef-input flag did not help. Reducing the bit width helped. Are these examples too big for Alive2?

```
define float @src(float %0) {
  %2 = fmul float %0, 1.000000
  ret float %2
}
define float @tgt(float %0) {
  ret float %0
}
```

AND

```
define i32 @src(i32 %0, i32 %1) {
  %3 = bitcast i32 %0 to float
  %4 = fmul float %3, 1.000000
  %5 = bitcast float %4 to i32
  ret i32 %5
}
define i32 @tgt(i32 %0, i32 %1) {
  ret i32 %0
}
```

regehr:
yes, timeouts are expected for queries containing FP, the problem is Z3, not alive2. you should keep using the --disable-undef-input flag and then increase the timeout to something huge.

also, Zhengyang is working on support for optionally modeling floats using uninterpreted functions, but of course that sort of model won't work for queries like these. we're hoping it will help his superoptimizer make progress on optimizing FP queries that don't care about what the FP instructions are actually doing. 

you should also look at Juneyoung Lee's CAV paper about mlir-tv, where he played some interesting games with FP semantics that are more precise than a UF but less precise then the full circuit. I don't see why we couldn't use some of those tricks in alive2. Nuno may or may not be interested in merging this sort of stuff, but you and Zhengyang could collaborate on a fork of the repo.

nunoplopes:
Right, this requires fundamental work that we haven't done yet. It's a know issue.

PR #1057, opened 2024-06-18
Optionally encode floating point operations as uninterpreted functions

can-leh-emmtrix:
In our ongoing work with alive2, we often encountered cases where floating point operations remain largely unchanged while other parts of the function differ. Since, Z3's floating point theory is quite slow I am looking at using uninterpreted functions to (over) approximate these floating point operations.

This PR adds a new command line option `--uf-float` that encodes all floating point operations using uninterpreted functions. The encoding is a conservative approximation, meaning that if a counter-example for the approximation is found, it is not necessarily a valid counter example.

Commutativity for `FPBinOp` and `FCmp` is encoded using the encoding proposed by Seongwon Bang, Seunghyeon Nam, Inwhan Chun, Ho Young Jhoo, and Juneyoung Lee in "SMT-based Translation Validation for Machine Learning Compiler.". It does not use the other parts of the encoding. It might be interesting to encode properties other than commutativity as well though.

I decided to open this PR as a draft, to see if there is interest in upstreaming this feature and get some feedback on the direction.

Related: https://github.com/AliveToolkit/alive2/issues/916


regehr:
so my (now former) student Zhengyang has been using floats-as-UF in his Alive-based superoptimizer. @zhengyang92 do you have feedback on this?

can-leh-emmtrix:
I'd consider this ready for review now.

As the diff is quite large, I'd recommend looking at the first three commits individually and then looking at the diff between the third and last commit.

The main reason for the large diff is that implementing the `--uf-float` option required indenting the existing code:

```cpp
if (is_uf_float()) {
  <create UF>
} else {
  <existing code>
}
```

can-leh-emmtrix:
@nunoplopes Do you have any feedback on this?

nunoplopes:
> @nunoplopes Do you have any feedback on this?

Sorry for the delay. I still have a few bugs to fix on my queue. I'll get back to this afterwards.

llvm_util/cmd_args_def.h
```diff
@@ -29,6 +29,8 @@ if ((config::disallow_ub_exploitation = opt_disallow_ub_exploitation)) {
   config::disable_poison_input = true;
 }
 
+config::fp_encoding_mode = opt_uf_float ? config::FpEncodingMode::UninterpretedFunctions : config::FpEncodingMode::FloatingPoint;
```
nunoplopes, line 32: long line

ir/instr.cpp
```diff
@@ -629,21 +629,38 @@ void FpBinOp::rauw(const Value &what, Value &with) {
...
+  case FMaximum:
+    return true;
+  default:
+    return false;
+  }
+  UNREACHABLE();
+}
+
+void FpBinOp::print(ostream &os) const {
+  os << getName() << " = " << getOpName() << " " << fmath << *lhs << ", " << rhs->getName();
```
nunoplopes, line 663: long line

ir/instr.cpp
```diff
@@ -1103,75 +1173,88 @@ void FpUnaryOp::rauw(const Value &what, Value &with) {
...
-  auto scalar = [&](const StateValue &v, const Type &ty) {
-    return fm_poison(s, v.value, v.non_poison,
-                     [fn](auto &v, auto &rm) {return fn(v, rm);}, ty, fmath, rm,
-                     bitwise, false);
-  };
+      ostringstream name;
+      name << getOpName() << "." << ty;
+      return uf_float(std::move(name).str(), {v}, v.value, fmath, false);
+    };
+  } else {
```
nunoplopes, line 1214: could you please reduce the diff size by not doing these huge re-indentation. otherwise it's very hard to review what's going on.
You can have an in afterwards to override the function.

ir/instr.cpp
```diff
@@ -836,86 +853,139 @@ static StateValue fm_poison(State &s, expr a, const expr &ap,
...
+      auto poison_uf = expr::mkUF(np_name, arg_values, false);
+      if (is_commutative) {
+        assert(args.size() == 2);
+        poison_uf &= expr::mkUF(np_name, {arg_values[1], arg_values[0]}, false);
+      }
+      non_poison.add(poison_uf);
+    }
+  };
+  
+  fast_math_flag(FastMathFlags::NNaN, "nnan");
```
nunoplopes, line 891: I don't understand this part. Also, where are the other flags?

ir/instr.cpp
```diff
@@ -836,86 +853,139 @@ static StateValue fm_poison(State &s, expr a, const expr &ap,
...
+  vector<expr> arg_values;
+  arg_values.reserve(args.size());
+  for (auto &arg : args) {
+    arg_values.push_back(arg.value);
+  }
+
+  auto value = expr::mkUF(name, arg_values, res);
+  if (is_commutative) {
+    assert(args.size() == 2);
+    value = value & expr::mkUF(name, {arg_values[1], arg_values[0]}, res);
```
nunoplopes, line 871: I don't think this is a sound over-approximation. Do you need commutativity in your examples?
Would an axiom stating that `forall x, y . fadd(x, y) = fadd(y, x)` suffice?

ir/instr.cpp
can-leh-emmtrix: What would be your preferred solution for reviewing? Is just not having the existing code indented for now (and then using a second PR that indents the relevant parts) fine? i.e.

```cpp
if (is_uf_float()) {
  <create UF>
} else {
<existing code>
}
```

nunoplopes:
```
existing code
if (is_uf) {
  fn = xpto
}
```

ir/instr.cpp
can-leh-emmtrix: The encoding comes from "SMT-based Translation Validation for Machine Learning Compiler" by Seongwon Bang, Seunghyeon Nam, Inwhan Chun, Ho Young Jhoo, and Juneyoung Lee. The idea is to encode commutativity as

$\mathrm{op}(x, y) = \mathrm{op}'(x, y) \mspace{0.3em} \mathrm{and} \mspace{0.3em} \mathrm{op}'(y, x)$

where $\mathrm{and}$ is the bitwise and operator and $\mathrm{op}'$ is an uninterpreted function. There is a proof of correctness in the [supplementary material](https://zenodo.org/records/6615676) for the paper.

can-leh-emmtrix: I added a comment referencing the paper.

ir/instr.cpp
can-leh-emmtrix: I have moved the existing code, the diff is quite a bit more manageable now.

ir/instr.cpp
can-leh-emmtrix: It handles flags which may result in the operation returning poison. As far as I can tell the only flags which can result in poison are `nnan` and `ninf`. They are handled by creating additional uninterpreted functions `op.np_nnan(x, y): bool` and `op.np_ninf(x, y): bool` (in the binary case) whose results are added to the `non_poison` and expression.

The `nsz` flag is not relevant as in uf-float mode we do not really have a zero value.

I don't think I entirely understand how the remaining fast math flags are handled in the existing code, though it looks like the result of the operation is wrapped in a function call to another UF.

ir/instr.cpp
```diff
@@ -1443,13 +1537,24 @@ StateValue FpTernaryOp::toSMT(State &s) const {
     };
     break;
   }
-
-  auto scalar = [&](const StateValue &a, const StateValue &b,
-                    const StateValue &c, const Type &ty) {
+
```
nunoplopes, line 1540: extra space here

ir/instr.cpp
```diff
@@ -1443,13 +1537,24 @@ StateValue FpTernaryOp::toSMT(State &s) const {
     };
     break;
   }
-
-  auto scalar = [&](const StateValue &a, const StateValue &b,
-                    const StateValue &c, const Type &ty) {
+  
+  scalar = [&](const StateValue &a, const StateValue &b,
```
nunoplopes, line 1541: these can stay unchanged. no need to declare it upfront

ir/instr.cpp
```diff
@@ -1443,13 +1537,24 @@ StateValue FpTernaryOp::toSMT(State &s) const {
...
+               const StateValue &c, const Type &ty) {
     return fm_poison(s, a.value, a.non_poison, b.value, b.non_poison,
                      c.value, c.non_poison, fn, ty, fmath, rm, false);
   };
 
+  if (config::is_uf_float()) {
+    scalar = [&](const StateValue &a, const StateValue &b,
+                      const StateValue &c, const Type &ty) -> StateValue {
+      s.doesApproximation("uf_float", true);
```
nunoplopes, line 1550: these tags are not very helpful like this. The 2nd argument should be the created UF.

ir/instr.cpp
```diff
@@ -2769,27 +2911,30 @@ void FCmp::rauw(const Value &what, Value &with) {
...
-  case UGE:   condtxt = "uge "; break;
-  case ULT:   condtxt = "ult "; break;
-  case ULE:   condtxt = "ule "; break;
-  case UNE:   condtxt = "une "; break;
-  case UNO:   condtxt = "uno "; break;
-  case TRUE:  condtxt = "true "; break;
-  case FALSE: condtxt = "false "; break;
-  }
-  os << getName() << " = fcmp " << fmath << condtxt << *a << ", "
+  case OEQ:   return "oeq"; break;
```
nunoplopes, line 2916: this new function is not used below

ir/instr.cpp
```diff
@@ -1865,6 +1985,28 @@ StateValue FpConversionOp::toSMT(State &s) const {
...
+      s.doesApproximation("uf_float", true);
+
+      ostringstream os;
+      os << getOpName() << "." << from_type << ".to." << to_type;
+      expr range = to_type.getDummyValue(true).value;
+      auto value = expr::mkUF(os.str(), {sv.value}, range);
+
+      AndExpr non_poison;
+      non_poison.add(sv.non_poison);
+      if (op != SIntToFP && !(op == UIntToFP && !(flags & NNEG)) &&
```
nunoplopes, line 2000: what's this?

ir/instr.cpp
nunoplopes: The proof is very vague and doesn't show equisatisfiability clearly.
Why not use `x <= y ? f(x,y) : f(y, x)`?

ir/instr.cpp
can-leh-emmtrix: Some of the floating point conversion operators are only partially defined (i.e. they return poison for some inputs), while others can convert all values and never return poison. The idea was to only generate the UF for the poison case if the conversion is partially defined. I've refactored this to use the `uf_float` function as well now. I think it should be much clearer now.

ir/instr.cpp
can-leh-emmtrix: I would prefer to keep the current encoding, as I know that it works well on our test cases. While I could try other encodings, that might of course take some time. So here is my attempt at constructing a proof. It is of course inspired by the proof given in the supplementary material, but uses a simpler construction for $op'$ / $f'$.

Please correct me if I am wrong, but since we are already over approximating, we only need to prove that for all **commutative** functions $\mathrm{op}: \mathrm{BV}^n \times \mathrm{BV}^n \rightarrow \mathrm{BV}^n$ there is a function $\mathrm{op}': \mathrm{BV}^n \times \mathrm{BV}^n \rightarrow \mathrm{BV}^n$ such that

$$\mathrm{op}(x, y) = \mathrm{op}'(x, y) \mspace{0.3em} \mathrm{and} \mspace{0.3em} \mathrm{op}'(y, x)$$

(in the context of SMT solving this ensures that no model/counterexample can get lost.)

**Proof:** Let $\mathrm{BV}^n$ be the set of all bit vectors of size $n$ and let $\mathrm{op}: \mathrm{BV}^n \times \mathrm{BV}^n \rightarrow \mathrm{BV}^n$ be a commutative function.

Let $\mathrm{op}'(x, y) := \mathrm{op}(x, y)$. Then 

$$
\mathrm{op}'(x, y) \mspace{0.3em} \mathrm{and} \mspace{0.3em} \mathrm{op}'(y, x)
= \mathrm{op}(x, y) \mspace{0.3em} \mathrm{and} \mspace{0.3em} \mathrm{op}(y, x)
$$

By commutativity of $\mathrm{op}$

$$
= \mathrm{op}(x, y) \mspace{0.3em} \mathrm{and} \mspace{0.3em} \mathrm{op}(x, y)
$$

By $\forall x \in \mathrm{BV}^n: x \mspace{0.3em} \mathrm{and} \mspace{0.3em} x = x$

$$
= \mathrm{op}(x, y)
$$

Thus we have shown that $\mathrm{op}'$ fullfills the requirement.

ir/instr.cpp
can-leh-emmtrix: Leaving them as `auto` does not work as I later need to assign a different lambda to the `scalar` variable. Each lambda expression has a unique type. Assignment operators are actually explicitly deleted for lambdas with captures.

I've now moved the declarations to the same statement as the initial defintions.

ir/instr.cpp
```diff
@@ -1444,12 +1540,23 @@ StateValue FpTernaryOp::toSMT(State &s) const {
...
+        const StateValue &c, const Type &ty) {
     return fm_poison(s, a.value, a.non_poison, b.value, b.non_poison,
                      c.value, c.non_poison, fn, ty, fmath, rm, false);
   };
 
+  if (config::is_uf_float()) {
+    scalar = [&](const StateValue &a, const StateValue &b,
+                 const StateValue &c, const Type &ty) -> StateValue {
+      ostringstream name;
+      name << getOpName() << "." << ty;
```
nunoplopes, line 1555: '.'

ir/instr.cpp
```diff
@@ -1413,14 +1507,16 @@ void FpTernaryOp::rauw(const Value &what, Value &with) {
...
-  case MulAdd: str = "fmuladd "; break;
+  case FMA:    return "fma"; break;
+  case MulAdd: return "fmuladd"; break;
   }
+  UNREACHABLE();
+}
 
-  os << getName() << " = " << str << fmath << *a << ", " << *b << ", " << *c;
+void FpTernaryOp::print(ostream &os) const {
+  os << getName() << " = " << getOpName() << " " << fmath << *a << ", " << *b << ", " << *c;
```
nunoplopes, line 1519: line seems quite long

ir/instr.cpp
```diff
@@ -1171,12 +1257,20 @@ StateValue FpUnaryOp::toSMT(State &s) const {
     break;
   }
 
-  auto scalar = [&](const StateValue &v, const Type &ty) {
+  scalar = [&](const StateValue &v, const Type &ty) {
```
nunoplopes, line 1260: keep auto

ir/instr.cpp
```diff
@@ -1107,31 +1190,34 @@ void FpUnaryOp::rauw(const Value &what, Value &with) {
...
+  os << getName() << " = " << getOpName() << " " << fmath << *val;
   if (!rm.isDefault())
     os << ", rounding=" << rm;
   if (!ex.ignore())
     os << ", exceptions=" << ex;
 }
 
 StateValue FpUnaryOp::toSMT(State &s) const {
-  expr (*fn)(const expr&, const expr&) = nullptr;
+  function<expr(const expr&, const expr&)> fn = nullptr;
```
nunoplopes, line 1219: do you need to change the decl of fn?

ir/instr.cpp
```diff
@@ -840,6 +858,60 @@ static StateValue fm_poison(State &s, expr a, const expr &ap,
...
+    assert(args.size() == 2);
+    value = value & expr::mkUF(name, {arg_values[1], arg_values[0]}, res);
+  }
+
+  AndExpr non_poison;
+  for (auto &arg : args) {
+    non_poison.add(arg.non_poison);
+  }
+
+  auto poison_condition = [&](const char* suffix){
```
nunoplopes, line 892: missing space before {

nunoplopes:
I was just thinking about this again and I think there's a better approach, which is much less invasive.
We could implement this in smt/expr.cpp instead and have a global mode to switch to UF encoding. That way, instr.cpp wouldn't be changed, all fmath flags would work, and we have a single place to change things, instead of duplicating semantics in instr.cpp.

can-leh-emmtrix:
I played a bit with the idea, but I was unable to achieve the same performance as when implementing the feature in instr.cpp. I used the `uf-float/select` testcase as my benchmark.

I tried a few different things, but here are the most successful approaches I could come up with.

- `No Approximation` is just using Z3's floating point theory
- `expr.cpp / Floating Point` is implemented in `expr.cpp` and uses uninterpreted functions that take floating point values as arguments
- `expr.cpp / Bit Vector` is implemented in `expr.cpp` and uses uninterpreted functions that take floating point values as arguments
- `instr.cpp` is this PR

Here are the results. I measured the total runtime of `alive-tv` on `uf-float/select` over 8 runs per approach.

uf_float_impls_1

Here is a chart that shows only the approximations.

uf_float_impls_2

So at least on this benchmark the approach taken by this PR has a clear advantage over inserting the uninterpreted functions in `expr.cpp`. This also matches my prior observations in our internal examples.

I think that the reason for this difference are the other operations applied `fm_poison`. It uses `FloatType::fromFloat`, `any_fp_zero`, and `handle_subnormal`, all of which are not neccessary for the approach taken by this PR. Additionally, I conjecture that in examples using instructions which are implemented as a more complex series of Z3 operations (e.g. `FMin`), we will observe the same problem. This is something that can not be fixed in expr.cpp.

Regarding the remaining fast math flags: Using the same logic as in `fm_poison` to over approximate them should work for `uf_float` too, shouldn't it?

can-leh-emmtrix:
Here is the relevant part of the diff for `expr.cpp / Floating Point`. This is of course very much a prototype and only implements the operations I needed for benchmarking.

```diff
diff --git a/smt/expr.cpp b/smt/expr.cpp
index 70353e8a..0af924a2 100644
--- a/smt/expr.cpp
+++ b/smt/expr.cpp
@@ -1205,7 +1205,8 @@ expr expr::fadd(const expr &rhs, const expr &rm) const {
 
 expr expr::fsub(const expr &rhs, const expr &rm) const {
   C(rhs, rm);
-  return simplify_const(Z3_mk_fpa_sub(ctx(), rm(), ast(), rhs()), *this, rhs);
+  //return simplify_const(Z3_mk_fpa_sub(ctx(), rm(), ast(), rhs()), *this, rhs);
+  return expr::mkUF("fsub", {*this, rhs}, *this);
 }
 
 expr expr::fmul(const expr &rhs, const expr &rm) const {
@@ -1282,7 +1283,8 @@ expr expr::foge(const expr &rhs) const {
 }
 
 expr expr::folt(const expr &rhs) const {
-  return binop_fold(rhs, Z3_mk_fpa_lt);
+  //return binop_fold(rhs, Z3_mk_fpa_lt);
+  return expr::mkUF("flt", {*this, rhs}, true);
 }
 
 expr expr::fole(const expr &rhs) const {
@@ -1310,7 +1312,8 @@ expr expr::fuge(const expr &rhs) const {
 }
 
 expr expr::fult(const expr &rhs) const {
-  return funo(rhs) || binop_fold(rhs, Z3_mk_fpa_lt);
+  //return funo(rhs) || binop_fold(rhs, Z3_mk_fpa_lt);
+  return funo(rhs) || expr::mkUF("flt", {*this, rhs}, true);
 }
 
 expr expr::fule(const expr &rhs) const {
```

Here is the relevant part of the diff for `expr.cpp / Bit Vector`:

```diff
diff --git a/smt/expr.cpp b/smt/expr.cpp
index 70353e8a..729c8ac3 100644
--- a/smt/expr.cpp
+++ b/smt/expr.cpp
@@ -1205,7 +1205,8 @@ expr expr::fadd(const expr &rhs, const expr &rm) const {
 
 expr expr::fsub(const expr &rhs, const expr &rm) const {
   C(rhs, rm);
-  return simplify_const(Z3_mk_fpa_sub(ctx(), rm(), ast(), rhs()), *this, rhs);
+  //return simplify_const(Z3_mk_fpa_sub(ctx(), rm(), ast(), rhs()), *this, rhs);
+  return expr::mkUF("fsub", {float2BV(), rhs.float2BV()}, float2BV()).BV2float(rhs);
 }
 
 expr expr::fmul(const expr &rhs, const expr &rm) const {
@@ -1282,7 +1283,8 @@ expr expr::foge(const expr &rhs) const {
 }
 
 expr expr::folt(const expr &rhs) const {
-  return binop_fold(rhs, Z3_mk_fpa_lt);
+  //return binop_fold(rhs, Z3_mk_fpa_lt);
+  return expr::mkUF("flt", {float2BV(), rhs.float2BV()}, true);
 }
 
 expr expr::fole(const expr &rhs) const {
@@ -1310,7 +1312,8 @@ expr expr::fuge(const expr &rhs) const {
 }
 
 expr expr::fult(const expr &rhs) const {
-  return funo(rhs) || binop_fold(rhs, Z3_mk_fpa_lt);
+  //return funo(rhs) || binop_fold(rhs, Z3_mk_fpa_lt);
+  return funo(rhs) || expr::mkUF("flt", {float2BV(), rhs.float2BV()}, true);
 }
 
 expr expr::fule(const expr &rhs) const {
```

ir/instr.cpp
```diff
@@ -1522,10 +1631,19 @@ StateValue TestOp::toSMT(State &s) const {
     break;
   }
 
-  auto scalar = [&](const StateValue &v, const Type &ty) -> StateValue {
+  function<StateValue(const StateValue&, const Type&)> scalar =
```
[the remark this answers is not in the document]
can-leh-emmtrix: > Leaving them as `auto` does not work as I later need to assign a different lambda to the `scalar` variable. Each lambda expression has a unique type. Assignment operators are actually explicitly deleted for lambdas with captures.

ir/instr.cpp
```diff
@@ -1840,8 +1960,9 @@ StateValue FpConversionOp::toSMT(State &s) const {
     break;
   }
 
-  auto scalar = [&](const StateValue &sv, const Type &from_type,
-                    const Type &to_type) -> StateValue {
+  function<StateValue(const StateValue &, const Type &, const Type&)> scalar =
```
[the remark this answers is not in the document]
can-leh-emmtrix: This has the same problem as the other case:

> Leaving them as `auto` does not work as I later need to assign a different lambda to the `scalar` variable. Each lambda expression has a unique type. Assignment operators are actually explicitly deleted for lambdas with captures.

nunoplopes:
I think those results are very encouraging and really suggest we should go the expr.cpp way.
Note that I don't think we should be using float2BV() at all. Floats should not exist; only BVs. We can also have the bit-width be user-specified.

This solution is superior in terms of maintainability. And getting rid of floats altogether should close the performance gap.

can-leh-emmtrix:
It would still require some changes in `instr.cpp` or `type.cpp` to disable checking for NaN and inserting nondeteriministic values though. Do you have a solution for instructions implemented in terms of multipe floating point and non floating point Z3 operations? Ideally they would be reduced to a single UF as well.

nunoplopes:
> It would still require some changes in `instr.cpp` or `type.cpp` to disable checking for NaN and inserting nondeteriministic values though. Do you have a solution for instructions implemented in terms of multipe floating point and non floating point Z3 operations? Ideally they would be reduced to a single UF as well.

I don't remember of cases that are not 1-to-1 mapping with Z3. Do you have an example in mind?

(btw, I'm taking off for vacations tmr; I'm back in September)

can-leh-emmtrix:
> I don't remember of cases that are not 1-to-1 mapping with Z3. Do you have an example in mind?

Looking through the source code I found the following cases:

- `FMin`, `FMax`: https://github.com/AliveToolkit/alive2/blob/master/ir/instr.cpp#L884-L890
- `FMinimum`, `FMaximum`: https://github.com/AliveToolkit/alive2/blob/master/ir/instr.cpp#L897-L905
- `MulAdd`: https://github.com/AliveToolkit/alive2/blob/master/ir/instr.cpp#L1441-L1442
- `LRInt`, `LRound`: https://github.com/AliveToolkit/alive2/blob/master/ir/instr.cpp#L1788-L1823

For the `FpConversionOp` instructions `LRInt`, `LRound`, `FPToSInt`, `FPToUInt` the `non_poison` value is also computed using multiple floating point operations (even excluding the logic added by `fm_poison`): https://github.com/AliveToolkit/alive2/blob/master/ir/instr.cpp#L1788-L1832

> (btw, I'm taking off for vacations tmr; I'm back in September)

Have a nice vacation!

nunoplopes:
Closing this as we don't want to have to duplicate semantics. A solution in smt/expr.cpp would avoid that.
Plus it's a better solution longer term once we start supporting abstraction refinement.

PR #1064, opened 2024-06-29
callees metadata

Hels15:
Callees metadata support for alive2 IR. See here: https://llvm.org/docs/LangRef.html#callees-metadata

Sample code: tests/alive-tv/callees.srctgt.ll


Hels15:
for review: @nunoplopes

PR #1073, opened 2024-07-15
Add named queries

can-leh-emmtrix:
Adds prefixes to the output files of `--smt-bench` in order to make it easier to find the query you are looking for. Output files are named `<QUERY_NAME>_<RANDOM_STRING>.smt2`. The query names are also shown when using `--smt-verbose`.

### DIFF
--- a/smt/solver.cpp
+++ b/smt/solver.cpp
@@ -504,7 +504,7 @@ Result Solver::check(const char *query_name) const {
   if (print_queries) {
     dbg() << "\nSMT query";
     if (query_name != nullptr) {
-      dbg() << " (" << query_name << ")";
+      dbg() << " (" << query_name << ')';
     }
     dbg() << ":\n" << Z3_solver_to_string(ctx(), s) << endl;
   }
--- a/smt/solver.cpp
+++ b/smt/solver.cpp
@@ -467,7 +467,7 @@ expr Solver::assertions() const {
   return ret;
 }
 
-Result Solver::check() const {
+Result Solver::check(const char *query_name) const {
   if (!valid) {
     ++num_invalid;
     return Result::INVALID;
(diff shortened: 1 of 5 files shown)


util/file.h
```diff
@@ -23,6 +23,6 @@ class file_reader {
 
 struct FileIOException {};
 
-std::string get_random_filename(const std::string &dir, const char *extension);
+std::string get_random_filename(const std::string &dir, const char *extension, const char *prefix = nullptr);
```
nunoplopes, line 26: this way over the 80-char limit

util/file.cpp
```diff
@@ -26,9 +27,16 @@ file_reader::file_reader(const char *filename, unsigned padding) {
...
+string get_random_filename(const string &dir, const char *extension, const char *prefix) {
   // there's a low probability of race here
-  auto newname = [&]() { return get_random_str(12) + '.' + extension; };
+  auto newname = [&]() {
+    ostringstream name;
+    if (prefix) {
+      name << prefix << '_';
+    }
+    name << get_random_str(12) << '.' << extension;
+    return name.str();
```
nunoplopes, line 38: if you want to use a stream, do std::move(name).str()

smt/solver.cpp
```diff
@@ -501,8 +501,13 @@ Result Solver::check() const {
   }
 
   ++num_queries;
-  if (print_queries)
-    dbg() << "\nSMT query:\n" << Z3_solver_to_string(ctx(), s) << endl;
+  if (print_queries) {
+    dbg() << "\nSMT query";
+    if (query_name != nullptr) {
+      dbg() << " (" << query_name << ")";
```
nunoplopes, line 507: ')'

PR #1074, opened 2024-07-23
Refactor llvm_util

can-leh-emmtrix:
Reduces the amount of global state by moving `value_cache`, `value_names` and `value_id_counter` to the `llvm2alive_` visitor.

### DIFF
--- a/ir/type.cpp
+++ b/ir/type.cpp
@@ -24,6 +24,12 @@ namespace IR {
 
 VoidType Type::voidTy;
 
+FloatType Type::halfTy("half", FloatType::Half);
+FloatType Type::floatTy("float", FloatType::Float);
+FloatType Type::doubleTy("double", FloatType::Double);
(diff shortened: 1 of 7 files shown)


nunoplopes:
Is there any good reason for doing these massive changes?

can-leh-emmtrix:
We are interested in using alive2 as a library. This requires cleaning up the API, especially reducing the reliance on global variables. The main reason for the large diff is moving the functions previously defined in `utils.cpp` to the `llvm2alive_` visitor.

nunoplopes:
I understand, but you won't be able to use Alive2 in a multi-threaded way. Not without a majore refactoring of the SMT stuff.
So I don't see why reducing global variables helps in any way?

nunoplopes:
Yes, I would appreciate more background before wasting a lot of time reviewing massive changes, as well as increasing the size of files that are already huge.

PR #1082, opened 2024-08-03
initialize current_bb with entry bb

dtcxzyw:
Alive2: https://alive2.llvm.org/ce/z/LQSeyF
```
define half @src(i8 %num) {
init:
  %#range_%num = !range i8 %num, i8 0, i8 8
  br label %entry

entry:
  %conv1 = uitofp i8 %#range_%num to half
  %div = fdiv half %conv1, 0x4900
  ret half %div
}
=>
define half @tgt(i8 %num) {
entry:
  %conv1 = uitofp i8 %num to half
  %div = fdiv half 0x3c00, 0x4900
  %mul = fmul half %conv1, %div
  ret half %mul
}

ERROR: Value mismatch

Example:
i8 %num = #x03 (3)

Source:
i8 %#range_%num = #x03 (3)
UB triggered on br

Target:
half %conv1 = #x4200 (3)
half %div = #x2e66 (0.099975585937?)
half %mul = #x34cc (0.2998046875)
Source value: #x34cd (0.300048828125)
Target value: #x34cc (0.2998046875)
```

`State::current_bb` is initialized with nullptr. When handling the branch from `#init -> entry`, `State::addJump` will create an item in `predecessor_data[entry][nullptr]`. Therefore, `State::getJumpCond(#init, entry)` evaluates to false. It causes the interpreter to report `UB triggered on br`.

### DIFF
--- a/tests/alive-tv/fp/nondet-model.srctgt.ll
+++ b/tests/alive-tv/fp/nondet-model.srctgt.ll
@@ -1,3 +1,4 @@
+; TEST-ARGS: -disable-undef-input
 ; ERROR: Value mismatch
 ; CHECK: %m = #x7e00
 
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -252,10 +252,10 @@ State::VarArgsData::mkIf(const expr &cond, VarArgsData &&then,
 }
 
 State::State(const Function &f, bool source)
-    : f(f), source(source), current_bb(&f.getFirstBB()), memory(*this),
-      fp_rounding_mode(expr::mkVar("fp_rounding_mode", 3)),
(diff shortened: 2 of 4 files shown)


tests/alive-tv/fp/nondet-model.srctgt.ll
```diff
@@ -1,3 +1,4 @@
+; TEST-ARGS: -disable-undef-input
```
dtcxzyw, line 1: SMT timeout is reached on my machine.

nunoplopes:
Thanks. I've fixed the bug in a better way. This was a botched refactoring I did a while back.
I'm merging your test case anyway. Thank you!

PR #1089, opened 2024-09-19
Make memcpy/memset/memcmp inherit from `FnCall`

antoniofrighetto:
Let memcpy et alia, which are intrinsically call instructions, inherit from FnCall so as to extend FnCall attributes to their subclasses as well. This is a preparatory change for supporting tail calls.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2,6 +2,7 @@
 // Distributed under the MIT license that can be found in the LICENSE file.
 
 #include "ir/instr.h"
+#include "ir/attrs.h"
 #include "ir/function.h"
 #include "ir/globals.h"
 #include "ir/type.h"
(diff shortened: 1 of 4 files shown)


nunoplopes:
But it seems that attributes are not useful for anything here.
If it's just for tail calls, wouldn't a bool suffice?  Or are there other attributes that apply to  memcpy & friends?
Otherwise, this is just increasing memory consumption.

antoniofrighetto:
Might be wrong, but I believe technically this should be the correct thing to do, not just for the tail calls, but for whatever attribute we parse from LLVM for such calls, in addition to having the code more cleaned and having things correctly separated, instead of having a tail call bool for each fncall/memcpy/memset/etc (I think it would be quite strange to have it in the base MemInstr).

Also, if we inherit from FnCall, wouldn't the bulk of additional memory be the vector for the arguments (which should still be limited in size)? FnAttrs looks fairly constant in space. For the time spent on parsing attributes, we seem to already do this for other intrinsics (any default unknown intrinsic).

PR #1094, opened 2024-10-05
Refine tail call conditions for `tail` and `musttail` markers

antoniofrighetto:
musttail has stronger requirements than tail, so ensure this is taken into account.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2494,8 +2494,7 @@ StateValue FnCall::toSMT(State &s) const {
     s.addGuardableUB(expr(false));
 
   const auto &attrs = getAttributes();
-  if (tci.isTailCall())
-    tci.checkTailCall(*this, s);
+  tci.checkTailCall(*this, s);
 
   auto get_alloc_ptr = [&]() -> Value& {
     for (auto &[arg, flags] : args) {
@@ -4149,8 +4148,7 @@ StateValue Memset::toSMT(State &s) const {
     vptr = sv_ptr.value;
   }
   check_can_store(s, vptr);
-  if (tci.isTailCall())
-    tci.checkTailCall(*this, s);
+  tci.checkTailCall(*this, s);
 
   s.getMemory().memset(vptr, s[*val].zextOrTrunc(8), vbytes, align,
                        s.getUndefVars());
@@ -4212,8 +4210,7 @@ StateValue MemsetPattern::toSMT(State &s) const {
   auto &vbytes = s.getAndAddPoisonUB(*bytes, true).value;
   check_can_store(s, vptr);
   check_can_load(s, vpattern);
-  if (tci.isTailCall())
-    tci.checkTailCall(*this, s);
+  tci.checkTailCall(*this, s);
 
   s.getMemory().memset_pattern(vptr, vpattern, vbytes, pattern_length);
   return {};
@@ -4342,8 +4339,7 @@ StateValue Memcpy::toSMT(State &s) const {
 
   check_can_load(s, vsrc);
   check_can_store(s, vdst);
-  if (tci.isTailCall())
-    tci.checkTailCall(*this, s);
+  tci.checkTailCall(*this, s);
 
   s.getMemory().memcpy(vdst, vsrc, vbytes, align_dst, align_src, move);
   return {};
@@ -4402,8 +4398,7 @@ StateValue Memcmp::toSMT(State &s) const {
 
   check_can_load(s, vptr1);
   check_can_load(s, vptr2);
-  if (tci.isTailCall())
-    tci.checkTailCall(*this, s);
+  tci.checkTailCall(*this, s);
 
   Pointer p1(s.getMemory(), vptr1), p2(s.getMemory(), vptr2);
   // memcmp can be optimized to load & icmps, and it requires this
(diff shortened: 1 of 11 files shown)


tests/alive-tv/calls/tailcall-memcpy.srctgt.ll
nunoplopes, line 1: why delete this test?  it's correct

tests/alive-tv/calls/musttail-memcpy.srctgt.ll
```diff
@@ -0,0 +1,18 @@
+define tailcc i64 @src(i64 noundef %arg) {
+  %call2 = alloca i64
+  %call3 = alloca i64
+  store i64 %arg, ptr %call2, align 8
+  musttail call tailcc void @llvm.memcpy.p0.p0.i64(ptr byval(ptr) %call3, ptr byval(ptr) %call2, i64 8, i1 false)
```
nunoplopes, line 5: this not a good test because it's mixing several conditions at once

tests/alive-tv/calls/tailcall-memcpy.srctgt.ll
antoniofrighetto: Moved it to `musttail-memcpy.srctgt.ll` and replace tail w/ musttail (since with tail only marker it shouldn't be UB, we're verifying only isMustTailCall now).

tests/alive-tv/calls/musttail-memcpy.srctgt.ll
antoniofrighetto: Separated it in multiple functions.

antoniofrighetto:
I think we miss checking the following:
> In addition, if the calling convention is not `swifttailcc` or `tailcc`:
      - All ABI-impacting function attributes, such as sret, byval, inreg,
   returned, and inalloca, must match.
      - The caller and callee prototypes must match. Pointer types of parameters
   or return types may differ in pointee type, but not in address space.

(Not completely sure if this refers to musttail too though)

nunoplopes:
From a quick read, this doesn't seem to be what we want at all. We need both tail & musttail, as both have constraints.
I suggest create a new thing in ir/attr.h (tail call type or smth like that). It can then encode the conditions for each case.

antoniofrighetto:
Added a new type `TailCallInfo` to encode both tail call markers. Updated the preconditions such that they are verified for both markers: tail position only for musttail, whereas callee not accessing allocas, va_args, or byval arguments from the caller for both. Restored `tailcall-memcpy.srctgt.ll` test since it violates this last constraint.

ir/instr.cpp
```diff
@@ -2,6 +2,7 @@
 // Distributed under the MIT license that can be found in the LICENSE file.
 
 #include "ir/instr.h"
+#include "ir/attrs.h"
```
nunoplopes, line 5: not needed

llvm_util/utils.cpp
```diff
@@ -521,4 +525,67 @@ llvm::Function *findFunction(llvm::Module &M, const string &FName) {
   return F && !F->isDeclaration() ? F : nullptr;
 }
 
+TailCallInfo parse_tail_call_info(const llvm::CallInst &i) {
```
nunoplopes, line 528: we don't encode semantics like this. The semantics must be in the ir folder.
We support other IRs -> Alive2 IR, hence we do as little as possible in the translation.

llvm_util/known_fns.h
```diff
@@ -18,6 +18,7 @@ class FnAttrs;
 class Instr;
 class ParamAttrs;
 class Value;
+struct TailCallInfo;
```
nunoplopes, line 21: seeems unused

llvm_util/known_fns.cpp
```diff
@@ -520,27 +520,32 @@ known_call(llvm::CallInst &i, const llvm::TargetLibraryInfo &TLI,
...
     RETURN_EXACT();
 
-  bool is_tailcall = i.isTailCall();
+  auto tci = parse_tail_call_info(i);
+
   switch (libfn) {
   case llvm::LibFunc_memset: // void* memset(void *ptr, int val, size_t bytes)
-    BB.addInstr(make_unique<Memset>(*args[0], *args[1], *args[2], 1, is_tailcall));
+    BB.addInstr(make_unique<Memset>(*args[0], *args[1], *args[2], 1,
+                                    std::move(tci)));
```
nunoplopes, line 528: move is a bit overkill

ir/instr.cpp
```diff
@@ -2314,28 +2314,36 @@ void FnCall::print(ostream &os) const {
   os << ')' << attrs;
 }
 
-static void check_tailcall(const Instr &i, State &s) {
```
nunoplopes, line 2317: you can move this function into the tailcallinfo struct

tests/alive-tv/calls/musttail-call.srctgt.ll
```diff
@@ -0,0 +1,11 @@
+define tailcc i64 @src(i64 noundef %arg) {
```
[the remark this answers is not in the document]
antoniofrighetto: Dropped it.

llvm_util/utils.cpp
antoniofrighetto: I left `parse_tail_call_info` in utils.cpp so that it may be used from known_functions too (originally was in llvm2alive). I think we still need it there as we're parsing data from LLVM (like `parse_fn_attrs`). Do you mean that the semantics in `are_tailcall_preconditions_met` is to be moved in ir inside `check_tailcall`?

llvm_util/known_fns.cpp
antoniofrighetto: To be changed after the change above.

llvm_util/utils.cpp
antoniofrighetto: Also made it more coherent (`parse_tail_call_info` -> `parse_fn_tailcall`).

antoniofrighetto:
Updated, moved semantics in checkTailCall.

PR #1095, opened 2024-10-09
Encode floating point operations as uninterpreted functions

can-leh-emmtrix:
Approach based on modifying expr.cpp. Please let me know what you think.

Previous PR: https://github.com/AliveToolkit/alive2/pull/1057

### DIFF
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -255,7 +255,10 @@ State::State(const Function &f, bool source)
   : f(f), source(source), memory(*this),
     fp_rounding_mode(expr::mkVar("fp_rounding_mode", 3)),
     fp_denormal_mode(expr::mkVar("fp_denormal_mode", 2)),
-    return_val(DisjointExpr(f.getType().getDummyValue(false))) {}
+    return_val(DisjointExpr(f.getType().getDummyValue(false))) {
+
+  doesApproximation("uf-float");
+}
 
 void State::resetGlobals() {
   Memory::resetGlobals();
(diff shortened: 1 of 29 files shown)


nunoplopes:
(not forgotten, but review delayed for a few weeks, sorry. PLDI deadline approaching)

ir/state.cpp
```diff
@@ -255,7 +255,11 @@ State::State(const Function &f, bool source)
   : f(f), source(source), memory(*this),
     fp_rounding_mode(expr::mkVar("fp_rounding_mode", 3)),
     fp_denormal_mode(expr::mkVar("fp_denormal_mode", 2)),
-    return_val(DisjointExpr(f.getType().getDummyValue(false))) {}
+    return_val(DisjointExpr(f.getType().getDummyValue(false))) {
+
+  if (get_uf_float())
+    doesApproximation("uf-float");
```
can-leh-emmtrix, line 261: I am not sure what the best way to handle `doesApproximation` is with this approach. One option would be to call it in `fm_poison` in `instr.cpp`, however we don't have direct access to the UFs anymore. From what I can tell from the source code, the second argument to `doesApproximation` needs to be the APP node of the UF. However we do not have access to that anymore at this point. It may be possible to obtain the UFs by traversing the AST though. Another option might be to register a callback function with `expr.h` which is called for each created UF.

can-leh-emmtrix:
Thanks for letting me know and no worries about the delay.

tnuha:
@can-leh-emmtrix I have a local branch that catches these changes up to 28bb8c608b1e24822a93b7f9ae35ba462ef66dba, but wanted to run it by you before I push anything to your branch in case you have other work on your end that hasn't been broadcast yet.

Certainly https://github.com/AliveToolkit/alive2/pull/1095#discussion_r1823240876 still hasn't been resolved, but these changes overall seem pretty reasonable to me.

PR #1098, opened 2024-10-15
Add support for `samesign` flag

dtcxzyw:
RFC link: https://discourse.llvm.org/t/rfc-signedness-independent-icmps/81423

https://github.com/llvm/llvm-project/pull/111419 has been landed.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2635,8 +2635,11 @@ InlineAsm::InlineAsm(Type &type, string &&name, const string &asm_str,
            std::move(attrs)) {}
 
 
-ICmp::ICmp(Type &type, string &&name, Cond cond, Value &a, Value &b)
-  : Instr(type, std::move(name)), a(&a), b(&b), cond(cond), defined(cond != Any) {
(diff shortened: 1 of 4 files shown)


nunoplopes:
> What? @nunoplopes Can you check the CI configuration? I don't think this patch can be compiled as [llvm/llvm-project#111419](https://github.com/llvm/llvm-project/pull/111419) was just landed.

The CI doesn't link against LLVM. It builds just the 'alive' tool (because we need LLVM built with RTTI, and thus we would need to either build LLVM from scratch or cache an image somewhere)

ir/instr.cpp
```diff
@@ -2729,6 +2735,13 @@ StateValue ICmp::toSMT(State &s) const {
     UNREACHABLE();
   };
 
+  fn = [this, fn](const expr &av, const expr &bv, Cond cond) {
```
nunoplopes, line 2738: please change the function to return a StateValue instead of smt::expr. This BV concatenation is a bit of a hack.

dtcxzyw: Done.

PR #1111, opened 2024-11-12
Simplify applying bitwise operators on sign extended values

can-leh-emmtrix:
Since the result of comparison and logical operators in C is of type `int`, complex logical expressions may result in expression trees where sext, extract and logical operators are used together (especially when the programmer uses temporary variables of type `int` to store boolean values and instcombine has not been executed).

This PR introduces the following simplifications for bit vectors of the same width `a` and `b`:

```
sext(a) & sext(b) -> sext(a & b)
sext(a) | sext(b) -> sext(a | b)
sext(a) ^ sext(b) -> sext(a ^ b)
~sext(a) -> sext(~a)
```

The pattern `extract(sext(x), ...)` is already handled by the existing simplifications.

If this breaks other important cases, it might be possible to enforce that `a.bits() = b.bits() = 1`, although I have personally seen a major impact anywhere else yet.

### DIFF
--- a/smt/expr.cpp
+++ b/smt/expr.cpp
@@ -497,10 +497,6 @@ bool expr::isSignExt(expr &val) const {
   return isUnOp(val, Z3_OP_SIGN_EXT);
 }
 
-bool expr::isZeroExt(expr &val) const {
-  return isUnOp(val, Z3_OP_ZERO_EXT);
-}
-
 bool expr::isAShr(expr &a, expr &b) const {
   return isBinOp(a, b, Z3_OP_BASHR);
 }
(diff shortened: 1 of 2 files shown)


nunoplopes:
These rewrite may increase the size of the formula, by adding an additional sext, hence it's not always a win. In general we try to only have local & global optimal rewrites to avoid memory & perf impact.
Do you have any performance numbers?

nunoplopes:
I've run a bunch of benchmarks over the weekend and this patch doesn't help. Most benchmarks get a bit slower actually.

PR #1128, opened 2024-12-03
X86: add support for SSE2/AVX2 intrinsics

artagnon:
Add X86 vector SSE2/AVX2 intrinsics to the IR, complete with formal semantics for the verification flow.

The work was originally authored by Zhengyang Liu <liuiz@cs.utah.edu>, and was motivated by the development of the Minotaur project [https://arxiv.org/abs/2306.00229].

PR #1129, opened 2024-12-03
config: introduce use_exact_fp option

artagnon:
The configuration option is useful for returning the exact fp from FloatType::fromFloat, ignoring NaN, and is required by Minotaur [https://arxiv.org/abs/2306.00229] to synthesize correctly. use_exact_fp is turned off by default, and should pose no correctness issue when turned on.

### DIFF
--- a/ir/type.cpp
+++ b/ir/type.cpp
@@ -6,6 +6,7 @@
 #include "ir/state.h"
 #include "smt/solver.h"
 #include "util/compiler.h"
+#include "util/config.h"
 #include <array>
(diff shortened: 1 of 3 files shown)


artagnon:
I have checked that when this is applied along with #1128, all tests in Minotaur pass.

nunoplopes:
This is not correct. Verification done by disabling the whole NaN stuff cannot be relied upon.
Either Minotaur assumes that or it uses an over-approximating semantics, such as using UFs.

PR #1132, opened 2024-12-06
ir: thread parent BB and Function throughout

artagnon:
Alive IR suffers from the deficiency that the parent BasicBlock is not known in the Instr class, and the parent Function is not known in the BasicBlock class. The motivation for this patch is that, to support certain attributes like vscale, the dynamic vscale value (which will only be available on the function) needs to be known in the Instr class when checking programs with scalable vector types. The side-effect of this patch is that it cleans up the API of several dup() functions to be more intuitive.

### DIFF
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -101,11 +101,10 @@ void BasicBlock::replaceTargetWith(const BasicBlock *from,
   }
 }
 
-unique_ptr<BasicBlock>
-BasicBlock::dup(Function &f, const string &suffix) const {
-  auto newbb = make_unique<BasicBlock>(name + suffix);
+unique_ptr<BasicBlock> BasicBlock::dup(const string &suffix) const {
+  auto newbb = make_unique<BasicBlock>(p_function, name + suffix);
   for (auto &i : instrs()) {
-    newbb->addInstr(i.dup(f, suffix));
+    newbb->addInstr(i.dup(suffix));
   }
   return newbb;
 }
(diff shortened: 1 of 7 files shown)


nunoplopes:
This patch increases memory consumption and the number of heap allocations for no good reason.

artagnon:
> This patch increases memory consumption and the number of heap allocations for no good reason.

I tried my best to implement vscale_range without this patch, but it seems that the dynamic vscale value needs to be known, and it's only available on the Function: several functions in the VectorType require the dynamic vscale value to be passed for correctness. Kindly reconsider.

artagnon:
> > This patch increases memory consumption and the number of heap allocations for no good reason.
> 
> I tried my best to implement vscale_range without this patch, but it seems that the dynamic vscale value needs to be known, and it's only available on the Function: several functions in the VectorType require the dynamic vscale value to be passed for correctness. Kindly reconsider.

For example, VectorType::getNumElements() will change depending on the dynamic vscale value. It is impossible for a VectorType to know the dynamic vscale value, because there is no Function associated with it. The alternative to this patch is to get the vscale value from the Function associated with the State and pass it around everywhere, but I don't know if that's much better than this patch in terms of memory consumption.

nunoplopes:
Isn't the dynamic vscale an SMT variable? If so, it's parte of the state, not part of the IR.

artagnon:
> Isn't the dynamic vscale an SMT variable? If so, it's parte of the state, not part of the IR.

Thanks for the guidance; I will make it part of the State.

PR #1139, opened 2024-12-09
ir: thread vscale range from State to Type

artagnon:
Thread the vscale range information, which is an smt::expr in State to various functions in Type, cascading down to a final numElementsConst(), in preparation for the first step in supporting scalable vectors.

-- 8< --
Based on #1121.


nunoplopes:
not the right way of implementing this change in the type system inference system.

PR #1140, opened 2024-12-09
type: add preliminiary support for singular vscale

artagnon:
Change llvm_util to correctly fill aggregates and scalable vector splats, based on the vscale_range attribute on the function, and add a virtual numElementsConst() override, to operate on the maximum value in the vscale range. Not considering all possible vscale values is obviously a correctness issue, so ensure that we only pass the type-check when the vscale range has a singular value.

In practice, the entire codebase needs to be taught about scalable vectors, before all programs with a singular vscale value are supported. The rem-costvscale.srctgt.ll test fails to type-check, for instance, due to ShuffleVector using a non-scalable mask. These instances are left as a todo for follow-up patches.

-- 8< --
Based on #1139.


nunoplopes:
As discussed offline, this patch is not the right way of implementing this feature. The type system code needs to be more closely aligned to how inference is done for e.g. integer bitwidths. This would cut the size of the patch in half at least.

Issue #1037, opened 2024-05-15
False negative with the sign bit of nans


dtcxzyw:
Alive2: https://alive2.llvm.org/ce/z/g8marn
```llvm

----------------------------------------
define i1 @src1() {
#0:
  %x = sqrt float -2494298614702763169545800705376256.000000
  %y = bitcast float %x to i32
  %cmp = icmp slt i32 %y, 0
  ret i1 %cmp
}
=>
define i1 @tgt1() {
#0:
  ret i1 1
}
Transformation seems to be correct!


----------------------------------------
define i1 @src2() {
#0:
  %x = sqrt float -2494298614702763169545800705376256.000000
  %y = bitcast float %x to i32
  %cmp = icmp slt i32 %y, 0
  ret i1 %cmp
}
=>
define i1 @tgt2() {
#0:
  ret i1 0
}
Transformation seems to be correct!

Summary:
  2 correct transformations
  0 incorrect transformations
  0 failed-to-prove transformations
  0 Alive2 errors
```

sqrt with a negative value (except for -0) returns nan. But we cannot make assumptions about the sign bit of return values.

See also https://github.com/llvm/llvm-project/issues/92217

cc @regehr @zhengyang92

dtcxzyw:
Tests with non-constant input:
```
define i1 @src1(float nofpclass(zero nan pinf pnorm psub) %a) {
  %x = call float @llvm.sqrt.f32(float %a)
  %y = bitcast float %x to i32
  %cmp = icmp slt i32 %y, 0
  ret i1 %cmp
}

define i1 @tgt1(float %a) {
  ret i1 true
}

define i1 @src2(float nofpclass(zero nan pinf pnorm psub) %a) {
  %x = call float @llvm.sqrt.f32(float %a)
  %y = bitcast float %x to i32
  %cmp = icmp slt i32 %y, 0
  ret i1 %cmp
}

define i1 @tgt2(float %a) {
  ret i1 false
}
```

dtcxzyw:
cc @arsenm

RalfJung:
> But we cannot make assumptions about the sign bit of return values.

Indeed. Specifically, a programmer cannot make such assumptions. The compiler is allowed to make the sign go either way. And that's what alive shows.

This is one of these cases where it is crucial to think of compilation in terms of *refinement* (the final program has fewer behaviors than the original program), not equivalence.

dtcxzyw:
[...]

PR #1155, opened 2025-01-04
Ignore sign difference between two NaNs

dtcxzyw:
As we discussed in https://github.com/AliveToolkit/alive2/issues/1037, we don't care about the sign of NaN values. This patch only checks the NaN payload when both values are NaNs.

Here is a false positive case: https://alive2.llvm.org/ce/z/iSh4-Z
Related PR: https://github.com/llvm/llvm-project/pull/121580

### DIFF
--- a/ir/type.cpp
+++ b/ir/type.cpp
@@ -611,8 +611,13 @@ pair<expr, expr>
 FloatType::refines(State &src_s, State &tgt_s, const StateValue &src,
                    const StateValue &tgt) const {
   expr non_poison = src.non_poison && tgt.non_poison;
-  return { src.non_poison.implies(tgt.non_poison),
-           (src.non_poison && tgt.non_poison).implies(src.value == tgt.value) };
+  expr equal_payload =
+      src.value.trunc(fractionBits()) == tgt.value.trunc(fractionBits());
+  expr equal = src.value == tgt.value;
+  return {src.non_poison.implies(tgt.non_poison),
+          non_poison.implies(expr::mkIf(getFloat(src.value).isNaN() &&
+                                            getFloat(tgt.value).isNaN(),
+                                        equal_payload, equal))};
 }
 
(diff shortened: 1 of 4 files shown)


dtcxzyw:
I don't know how to fix the test failure in `tests/alive-tv/bugs/pr45152.srctgt.ll` :(

nunoplopes:
There was no discussion about this change of the semantics AFAICT. Could you please open a discuss first at discourse so we can reach consensus before changing Alive2?
Thanks!

dtcxzyw:
I decided to fix it in LLVM. This case is a little different from https://github.com/AliveToolkit/alive2/issues/1037 since there is no floating-point operation being performed. As per the IEEE 754-2008 standard, the sign bit of a NaN is only allowed to be refined in floating point operations except for copy, negate, abs, and copySign.

dtcxzyw:
See also https://github.com/llvm/llvm-project/issues/59279

regehr:
sorry to add to a closed issue, but I just want to add that using arm-tv I run across quite a lot of issues relating to LLVM and Alive disagreeing about whether NaN and +0/-1 transformations are valid. @dtcxzyw I would be super happy to feed you examples if you're interested!

dtcxzyw:
> sorry to add to a closed issue, but I just want to add that using arm-tv I run across quite a lot of issues relating to LLVM and Alive disagreeing about whether NaN and +0/-1 transformations are valid. @dtcxzyw I would be super happy to feed you examples if you're interested!

Feel free to create a post on Discourse, and then we can discuss these cases in public :)
I also found some disagreements between Alive2 and InstCombine with [mutation-based fuzzing](https://github.com/dtcxzyw/llvm-mutation-based-fuzz-service/blob/399843c8b90dec1bc80ebb6a79e8214c91fe1c52/fuzz_existing.py#L29-L61). I am drafting two RFCs to clarify the LangRef wording.

PR #1158, opened 2025-01-10
vector/test: pre-commit tests for vscale support

artagnon:
Pre-commit some tests with the vscale syntax, in preparation to add support for scalable vectors.


nunoplopes:
Please send this in together when support is added. No need to precommit stuff that doesn't work.

PR #1159, opened 2025-01-13
Remove length limit on error message output for unsupported instructions.

IamYJLee:
Previously, the error message for unsupported instructions was limited to 80 characters. This change removes the length limit, allowing the full error message to be displayed for better debugging and user feedback.

Before :
```
ERROR: Unsupported instruction:   %insert.reg2mem.0.insert.reg2mem.0.insert.reg2mem.0.insert.reg2mem.0.insert.re
ERROR: Could not translate 'f1' to Alive IR
```
After : 
```
ERROR: Unsupported instruction:   %insert.reg2mem.0.insert.reg2mem.0.insert.reg2mem.0.insert.reg2mem.0.insert.reload2 = load volatile <4 x i16>, ptr %insert.reg2mem, align 8
ERROR: Could not translate 'f1' to Alive IR
```

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -1331,7 +1331,7 @@ public:
     ss << i;
 
     *out << "ERROR: Unsupported instruction: "
-         << string_view(std::move(ss).str()).substr(0, 80) << '\n';
+         << std::move(ss).str() << '\n';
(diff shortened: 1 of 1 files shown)


nunoplopes:
We can't remove the limit. Some instructions are just too large to print.

PR #1165, opened 2025-01-30
support captures(none) instead of nocapture

regehr:
this gets Alive to build against current LLVM by mapping the new captures(none) to Alive's NoCapture.

this PR leaves undone two additional items
- upgrade our test cases correspondingly (right now this is taken care of by the parser, so no rush)
-  support captures() other than none -- this is more work

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -1570,8 +1570,11 @@ public:
         attrs.set(ParamAttrs::NonNull);
         break;
 
-      case llvm::Attribute::NoCapture:
-        attrs.set(ParamAttrs::NoCapture);
+      case llvm::Attribute::Captures:
+        if (capturesNothing(llvmattr.getCaptureInfo().getOtherComponents()))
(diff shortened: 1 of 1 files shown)


llvm_util/llvm2alive.cpp
```diff
@@ -1570,8 +1570,11 @@ class llvm2alive_ : public llvm::InstVisitor<llvm2alive_, unique_ptr<Instr>> {
         attrs.set(ParamAttrs::NonNull);
         break;
 
-      case llvm::Attribute::NoCapture:
-        attrs.set(ParamAttrs::NoCapture);
+      case llvm::Attribute::Captures:
+        if (capturesNothing(llvmattr.getCaptureInfo().getOtherComponents()))
```
nikic, line 1574: This should be just `capturesNothin(llvmattr.getCaptureInfo())`. Otherwise you're saying a return-only capture is nocapture as well.

nunoplopes: fixed, thank you!

PR #1176, opened 2025-02-26
docs: clarify LLVM_ENABLE_EH requirement in README.md

IamYJLee:
Error message : 
```
CMake Error at CMakeLists.txt:183 (message):
LLVM must be built with '-DLLVM_ENABLE_EH=ON'
```


nunoplopes:
Not anymore. You are using an old version of Alive2.

PR #1188, opened 2025-04-17
Transform : Add opt for src=tgt 

abhilash1910:
Adds in the optimization to rewrite "tgt /\ (src -> foo)" to "tgt /\ foo" if src=tgt .
Tested against  existing  and newer opt.ll cases.
@regehr @nunoplopes please help to review, thanks.

### DIFF
--- a/tools/transform.cpp
+++ b/tools/transform.cpp
@@ -571,10 +571,15 @@ check_refinement(Errors &errs, const Transform &t, State &src_state,
       auto pre_src_and = src_state.getPre();
       auto &pre_tgt_and = tgt_state.getPre();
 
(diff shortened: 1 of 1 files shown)


abhilash1910:
@nunoplopes , yes this aims to target the commented opt path where src = tgt. For src!=tgt the default opt path is taken (in else).   Initially I only did pre_tgt = pre_tgt_and() inside the if()  to make it tgt ^foo, but this did not look after pre_src, hence added it . Could you help to suggest on a better way to add it ?  Thanks.

nunoplopes:
Does this give any perf benefit?

abhilash1910:
I am checking on this for perf enhancements. Will report here , thanks

PR #1190, opened 2025-04-21
Test TV CI

nunoplopes:


nunoplopes:
LLVM's cmake files seem to be broken, so precompiled packages from APT don't seem to work well.

Issue #1204, opened 2025-05-17
Type check failure with `llvm.frexp.{f16/f64}.i32`


dtcxzyw:
Reproducer: https://godbolt.org/z/7ErK1oY6G
```
; ./alive-tv test.ll
define { half, i32 } @src(half %x) {
  %frexp0 = call { half, i32 } @llvm.frexp.f16.i32(half %x)
  %frexp0.0 = extractvalue { half, i32 } %frexp0, 0
  %frexp1 = call { half, i32 } @llvm.frexp.f16.i32(half %frexp0.0)
  ret { half, i32 } %frexp1
}

define { half, i32 } @tgt(half %x) {
  %frexp1 = call { half, i32 } @llvm.frexp.f16.i32(half %x)
  %insert = insertvalue { half, i32 } %frexp1, i32 0, 1
  ret { half, i32 } %insert
}
```
```
----------------------------------------
define {half, i16, i32} @src(half %x) {
#0:
  %frexp0 = frexp half %x
  %frexp0.0 = extractvalue {half, i16, i32} %frexp0, 0
  %frexp1 = frexp half %frexp0.0
  ret {half, i16, i32} %frexp1
}
=>
define {half, i16, i32} @tgt(half %x) {
#0:
  %frexp1 = frexp half %x
  %insert = insertvalue {half, i16, i32} %frexp1, i32 0, 2
  ret {half, i16, i32} %insert
}
Transformation doesn't verify!
ERROR: program doesn't type check!
```

PR #1217, opened 2025-06-24
Handle paddings in type check

dtcxzyw:
Closes https://github.com/AliveToolkit/alive2/issues/1204.


nunoplopes:
`Transforms/Scalarizer/frexp.ll` is now crashing in function `test_vector_half_frexp_half` because of this patch.

dtcxzyw:
I will take a look.

PR #1225, opened 2025-07-11
Fix build error on AppleClang due to lambda capture of structured bin…

aabhinavg1:
OS AppleClang does not support capturing structured bindings (e.g., `src_data`) by reference in lambdas, which leads to compilation errors.

This ``patch refactors`` the ``affected loop`` by introducing a named reference variable outside the lambda, and ``capturing that reference`` instead. This improves compatibility across multiple compilers,`` including AppleClang``, `GCC`, `Clang`, and `MSVC`, while preserving the original behaviour.

No functional change is intended his is a portability fix.

### DIFF
--- a/ir/state.cpp
+++ b/ir/state.cpp
@@ -753,9 +753,10 @@ void State::copyUBFromBB(
 
   for (auto *src_bb : I->second) {
     bool all_paths_ok = true;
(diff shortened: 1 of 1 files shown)


nunoplopes:
why can't you upgrade your XCode version instead??

Issue #1222, opened 2025-07-06
false alarm in asm mode


regehr:
I'm pretty sure this isn't one that I've reported before.

This (obviously correct) transformation validates using regular alive-tv:

```llvm
define void @f(ptr %0, ptr %1) {
  store i16 0, ptr %0, align 4
  store i16 0, ptr %1, align 8
  store i16 0, ptr %0, align 4
  ret void
}
```
```llvm
define void @f(ptr %0, ptr %1) {
arm_tv_entry:
  store i16 0, ptr %1, align 1
  store i16 0, ptr %0, align 1
  ret void
}
```
but in asm mode I get UB:
```
regehr@john-home:~/alive2-regehr/build$ ./alive-tv reduced.ll tgt.ll -tgt-is-asm

----------------------------------------
define void @f(ptr %#0, ptr %#1) {
#2:
  store i16 0, ptr %#0, align 4
  store i16 0, ptr %#1, align 8
  store i16 0, ptr %#0, align 4
  ret void
}
=>
define void @f(ptr %#0, ptr %#1) asm {
arm_tv_entry:
  store i16 0, ptr %#1, align 4
  store i16 0, ptr %#0, align 8
  ret void
}
Transformation doesn't verify!

ERROR: Source is more defined than target

Example:
ptr %#0 = pointer(non-local, block_id=1, offset=0) / Address=#x0000000000000004
ptr %#1 = pointer(non-local, block_id=2, offset=0) / Address=#x0000000000000008

Source:

SOURCE MEMORY STATE
===================
NON-LOCAL BLOCKS:
Block 0 >	size: 0	align: 2	alloc type: 0	alive: false	address: 0
Block 1 >	size: 4	align: 1	alloc type: 0	alive: true	address: 4
Block 2 >	size: 13	align: 1	alloc type: 0	alive: true	address: 8

Target:
void = UB triggered!


Summary:
  0 correct transformations
  1 incorrect transformations
  0 failed-to-prove transformations
  0 Alive2 errors
regehr@john-home:~/alive2-regehr/build$ 
```

regehr:
only Nuno can do that :)
but I'm focusing on other stuff and if you have time to take a crack at this, it would be awesome!

IamYJLee:
I'll dig into this and see if I can identify the problem.

nunoplopes:
Oh, this is a dup of #1214.

We don't want to remove alignment inference. The algorithm needs to be made dataflow sensitive.
[...]

PR #1228, opened 2025-08-08
[test-suite] Add XFAIL test case for issue #1222 (-tgt-is-asm alignme…

IamYJLee:
…nt inference)

This patch adds a regression test to reproduce the issue described in issue #1222. The test is currently marked as XFAIL since the failure is expected due to the known bug. Once the underlying problem is fixed, the XFAIL directive should be removed.


nunoplopes:
we don't usually work with xfail tests. we'll get it fixed someday.

PR #1240, opened 2025-09-02
Memset to poison `dead_on_return` ptr arguments

antoniofrighetto:
`dead_on_return` attribute was recently introduced in LLVM.


ir/memory.cpp
```diff
@@ -1295,7 +1295,8 @@ void Memory::store(const Pointer &ptr,
 void Memory::storeLambda(const Pointer &ptr, const expr &offset,
                          const expr &bytes,
                          const vector<pair<unsigned, expr>> &data,
-                         const set<expr> &undef, uint64_t align) {
+                         const set<expr> &undef, uint64_t align,
+                         bool full_write) {
```
antoniofrighetto, line 1299: Might be possible to avoid using this extra boolean, can consider this if correct direction.

PR #1241, opened 2025-09-03
[objectsize] Without allocsize, create a fresh variable for the allocation size. (#1238)

IamYJLee:
> It's already allocating a new block. It's just it's setting the size as null instead of creating a fresh var in src, and reusing the same var in tgt. It requires some changes in state.cpp & memory.cpp.

Since in this case execution never reaches `addFnCall`, it seems that the suggested modifications to `state.cpp` and `memory.cpp` are not necessary. Am I misunderstanding this? Also, would it be incorrect to introduce a fresh variable when the computed size is invalid?

https://github.com/AliveToolkit/alive2/blob/55cc4bb5c2aa2eb77bd8105685d0f07f75b40af1/ir/instr.cpp#L2758-L2760

Fixes https://github.com/AliveToolkit/alive2/issues/1238

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -2685,7 +2685,15 @@ StateValue FnCall::toSMT(State &s) const {
   if (attrs.has(AllocKind::Alloc) ||
       attrs.has(AllocKind::Realloc) ||
       attrs.has(FnAttrs::AllocSize)) {
-    auto [size, np_size] = attrs.computeAllocSize(s, args);
+    auto [comput_size, np_size] = attrs.computeAllocSize(s, args);
+    
+    expr size;
+    if(!comput_size.isValid()) {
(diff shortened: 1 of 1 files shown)


IamYJLee:
At the point shown here:
https://github.com/AliveToolkit/alive2/blob/e222e050e0edd906854deacfe26dd21d193e09df/ir/memory.cpp#L2170-L2174
we create a fresh variable for allocsize, and then here:
https://github.com/AliveToolkit/alive2/blob/e222e050e0edd906854deacfe26dd21d193e09df/ir/state.cpp#L1493-L1502
we use `syncSEdataWithSrc` to synchronize the source and target.
Is that correct?

nunoplopes:
No, that's not correct. The object size is a function of the arguments and the current memory. There are no shortcuts. The real fix involves state.cpp/h (Function output).

As I said, this is a lot more complicated than you trying to make it sound.

PR #1246, opened 2025-10-02
Allow ScalableVectorType in llvm2alive and replace llvm.vscale() with one

zhengyang92:
This PR tries to add support of scalable vectors to Alive2. Intuitively, vector operations that verifiy in vscale=1 should also verify for the other vscales. Thus, this patch assumes vscale=1 when running TV, and a constant one will be returned for llvm.vscales() calls.

I attached a few unit test cases.

I am still running TV on the test/Transforms folder. I'll paste diffs after the run finishes.

CC: @regehr

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -1179,6 +1179,14 @@ public:
       PARSE_BINOP();
       return make_unique<VaCopy>(*a, *b);
     }
+    case llvm::Intrinsic::vscale: {
+      auto ty = llvm_type2alive(i.getType());
+      if (!ty)
+        return error(i);
+      // llvm.vscale() always returns 1
+      auto val = make_intconst(1, ty->bits());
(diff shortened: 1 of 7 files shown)


regehr:
Zhengyang-- since your patch doesn't make Alive actually use vscale, are you sure Alive is doing the right thing when it allocates SMT terms for scalable vectors? can you include a negative test case to show that we can detect an incorrect transformation over scalable vectors?

nunoplopes:
Who is going to do the work of support scalable vectors properly?  This patch is just a hack, and doesn't move the implementation any closer to a proper implementation..

zhengyang92:
@nunoplopes @regehr with current semantics on scalable vectors, do we have to do allocate SMT terms for vscale>1? Does there exist transformations that pass when vscale=1 but fail when vscale>1?

zhengyang92:
After carefully reading the semantics of insertelement and extractelement, I now get the reason why we need to allocate full SMT terms. I'll get back with a new PR soon.

Issue #1201, opened 2025-05-01
bug in minnum and maxnum


regehr:
I think these functions are incorrect in one corner case.

take this IR:
```llvm
declare float @llvm.min.f32(float, float)
declare float @llvm.max.f32(float, float)

declare float @llvm.minnum.f32(float, float)
declare float @llvm.maxnum.f32(float, float)

declare float @llvm.minimum.f32(float, float)
declare float @llvm.maximum.f32(float, float)

declare float @llvm.minimumnum.f32(float, float)
declare float @llvm.maximumnum.f32(float, float)

define float @f() {
  %2 = call float @llvm.maxnum.f32(float 0x0000000000000000, float 0x8000000000000000)
  ret float %2
}
```

here's what alive-exec says:
```
~/svn/code/alive2/fp$ ~/alive2-regehr/build/alive-exec tmp.ll 

----------------------------------------
define float @f() {
#0:
  %#1 = fmax float 0.000000, -0.000000
  ret float %#1
}
Executing %#0
%#1 = #x80000000
Returned #x80000000 / true
~/svn/code/alive2/fp$ 
```

but LangRef says that "-0.0 < +0.0 for the purposes of this intrinsic."

the analogous case for `minnum` is also wrong:
```
~/svn/code/alive2/fp$ ~/alive2-regehr/build/alive-exec tmp.ll 

----------------------------------------
define float @f() {
#0:
  %#1 = fmin float -0.000000, 0.000000
  ret float %#1
}
Executing %#0
%#1 = #x00000000
Returned #x00000000 / true
```

regehr:
ah, thanks, I'll just put the text here

```
commit 363b05944f9212511ee6811d0eb1af841c177226
Author: YunQiang Su <yunqiang@isrc.iscas.ac.cn>
Date:   Thu Feb 27 11:22:30 2025 +0800

    LangRef: Clarify llvm.minnum and llvm.maxnum about sNaN and signed zero (#112852)

    The documents claims that it ignores sNaN, while in the current code it
    may be different.

    - as the finally callback, it use libc call fmin(3)/fmax(3). while C23
    clarifies that fmin(3)/fmax(3) should return NaN for sNaN vs NUM.
    - on some architectures, such as aarch64, it converts to `fmaxnm`, which
    returns qNaN for sNaN vs NUM.
    - on RISC-V (SPEC 2019+), it converts to `fmax`, which returns NUM for
    sNaN vs NUM.

[...]

PR #1268, opened 2025-12-28
Vector Reductions: add unary floating point aggregation

sweiglbosker:
Adds support for

- `llvm.vector.reduce.fmin` (minnum/IEEE-754 fp semantics)
- `llvm.vector.reduce.fmax` 
- `llvm.vector.reduce.fminimum`
- `llvm.vector.reduce.fmaximum`

I think this was discussed a bit https://github.com/AliveToolkit/alive2/pull/405. These are also no longer experimental, so i think now is an even better time to support them.

Also, i preserved the current fmin/fmax behavior of treating -0 = 0, which technically goes against the langref (see https://github.com/AliveToolkit/alive2/issues/1201). I haven't really followed the debate around this but it can probably be landed separately anyway if we want to mirror the langref here.

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -908,6 +908,22 @@ public:
       ret = make_unique<UnaryReductionOp>(*ty, value_name(i), *val, op);
       break;
     }
+    case llvm::Intrinsic::vector_reduce_fmin:
+    case llvm::Intrinsic::vector_reduce_fmax:
+    case llvm::Intrinsic::vector_reduce_fminimum:
+    case llvm::Instrinsic::vector_reduce_fmaximum: {
+    PARSE_UNOP();
+    UnaryReductionOp::Op op;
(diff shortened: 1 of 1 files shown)


nunoplopes:
This is missing the llvm2alive2.cpp file, at least.

nunoplopes:
It didn't compile (I fixed it..) and fails 4 tests:
```
FAIL: Alive2 :: alive-tv/fp/reduce_fmax.srctgt.ll (1067 of 1423)
FAIL: Alive2 :: alive-tv/fp/reduce_fminimum.srctgt.ll (1151 of 1423)
FAIL: Alive2 :: alive-tv/fp/reduce_fmaximum.srctgt.ll (1319 of 1423)
FAIL: Alive2 :: alive-tv/fp/reduce_fmin.srctgt.ll (1335 of 1423)
```

nunoplopes:
You didn't compile with TV mode on. See the huge number of unsupported tests.

PR #1271, opened 2026-01-09
Make `dead_on_return` optionally sized

antoniofrighetto:
LLVM `dead_on_return` attribute now supports an optional size.


antoniofrighetto:
(To undraft when [PR171712](https://github.com/llvm/llvm-project/pull/171712) lands in LLVM.)

PR #1300, opened 2026-04-13
Add support for individual loop unroll bounds

jelvani:
These changes allow setting individual loop unroll bounds from the command line. This is useful if we want to check a property for a single loop in a multi-loop function, where a single loop unroll bound may lead to code blow-up.

Usage: If `src` and `tgt` have two loops each then individual loop unroll bounds can be used as follows:`alive-tv test.ll --src-unroll=1,2 --tgt-unroll=1,2`
The unroll bounds are applied to loops in the order they appear in code. These changes preserve backwards compatability; if a single loop unroll bound is provided then it is applied to all loops in the function.
-Alborz

### DIFF
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -702,8 +702,7 @@ void Function::unroll(vector<unsigned> unroll_bounds) {
       }
     }
 
-    if (unroll_bounds.size()>1)
-    {
+    if (unroll_bounds.size() > 1) {
       unroll_bounds.erase(unroll_bounds.begin());
     }
 
--- a/ir/function.cpp
+++ b/ir/function.cpp
@@ -11,6 +11,7 @@
 #include <fstream>
 #include <set>
 #include <unordered_set>
+#include <iostream>
 
 using namespace smt;
 using namespace util;
@@ -575,8 +576,8 @@ static auto getPhiPredecessors(const Function &F) {
   return map;
 }
 
-void Function::unroll(unsigned k) {
-  if (k == 0)
+void Function::unroll(vector<unsigned> unroll_bounds) {
+  if (unroll_bounds.size() == 0)
     return;
 
   LoopAnalysis la(*this);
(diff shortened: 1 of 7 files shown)


IamYJLee:
I think it would be useful.
But if we have many functions, such as functions in transform unit tests, we would need too many coupled options.

llvm_util/cmd_args_def.h
```diff
@@ -2,11 +2,15 @@
 // Distributed under the MIT license that can be found in the LICENSE file.
 
 #ifdef ARGS_SRC_TGT
-config::src_unroll_cnt = opt_src_unrolling_factor;
-config::tgt_unroll_cnt = opt_tgt_unrolling_factor;
+std::vector<unsigned> unroll_src(opt_src_unrolling_factor.begin(), opt_src_unrolling_factor.end());
```
nunoplopes: remove these temporaries & copies

ir/function.cpp
```diff
@@ -667,6 +702,11 @@ void Function::unroll(unsigned k) {
       }
     }
 
+    if (unroll_bounds.size()>1)
```
nunoplopes: format..

ir/function.cpp
```diff
@@ -587,8 +588,8 @@ static auto getPhiPredecessors(const Function &F) {
   return map;
 }
 
-void Function::unroll(unsigned k) {
-  if (k == 0)
+void Function::unroll(vector<unsigned> unroll_bounds) {
```
regehr, line 591: pass by const reference

regehr (commented):
this seems fine but you need to add test cases demonstrating the in complex situations, the correct loop always gets unrolled

we also need some sort of documentations telling users how to perform this mapping. not sure where this goes, maybe @nunoplopes has a good idea 

we support many src/tgt pairs as src1/tgt1, src2/tgt2, etc. how does this interact with these new unroll bounds?

I think you've changed the behavior of alive2 for the `unroll=0` case. probably that needs to be fixed.

otherwise this LGTM and seems like a useful change!

llvm_util/cmd_args_list.h
```diff
@@ -19,13 +19,15 @@ llvm::cl::list<std::string> opt_funcs(LLVM_ARGS_PREFIX "func",
 set<string> func_names;
 
 #ifdef ARGS_SRC_TGT
-llvm::cl::opt<unsigned> opt_src_unrolling_factor(LLVM_ARGS_PREFIX "src-unroll",
-  llvm::cl::desc("Unrolling factor for src function (default=0)"),
-  llvm::cl::init(0), llvm::cl::cat(alive_cmdargs));
+llvm::cl::list<unsigned> opt_src_unrolling_factor(LLVM_ARGS_PREFIX "src-unroll",
+  llvm::cl::desc("Unrolling factors for src function (default=0)"),
```
regehr, line 23: don't add spurious whitespace at the end of a line

Issue #1308, opened 2026-04-22
dropping IEEE denormal attributes


regehr:
I believe that we're incorrectly judging this to be a miscompile:
```llvm
define float @test_float_fdiv_pzero_f32_out() #0 {
  %result = fdiv float 0x3810000000000000, -2.000000e+00
  ret float %result
}

attributes #0 = { denormal_fpenv(positivezero|ieee) }
```

You can see that the denormal attributes aren't even making it into the Alive IR. I saw some code for parsing those so I'm not sure what's going on. I can work on this if you like.
```
~/alive2-regehr$ ./build/alive-tv foo.ll 

----------------------------------------
define float @test_float_fdiv_pzero_f32_out() {
#0:
  %result = fdiv float 1.1754944e-38, -2e+00
  ret float %result
}
=>
define float @test_float_fdiv_pzero_f32_out() nofree noundef willreturn memory(none) {
#0:
  ret float 0e+00
}
Transformation doesn't verify!

ERROR: Value mismatch

NOTE: The counterexample is unique.

Example:

Source:
float %result = #x80400000 (-5.877472e-39)

Target:
Source value: #x80400000 (-5.877472e-39)
Target value: #x00000000 (+0.0)

Summary:
  0 correct transformations
  1 incorrect transformations
  0 failed-to-prove transformations
  0 Alive2 errors
~/alive2-regehr$ 
```

regehr:
ah, ok, this is what happened in Feb: https://github.com/llvm/llvm-project/commit/2502e3b7badc

PR #1313, opened 2026-04-24
support new first-class denormal attributes

regehr:
in Feb Matt redid FP denorm attributes so that they're real LLVM attrs instead of strings:
https://github.com/llvm/llvm-project/commit/2502e3b7badc

so alive2 now has some false alarms due to failing to parse the new-style attributes: #1308

we don't want to remove the old string-based code path / tests yet since it looks like the text IR parser does not run the IR auto-upgrader. we can remove this code path / tests once LLVM drops support for the old string attribute syntax.

if @arsenm happens to have time to take a quick look at the test cases attached to this PR, that would be helpful


arsenm:
> once LLVM drops support for the old string attribute syntax.

It was dropped, it's only handled for bitcode autoupgrade

PR #1315, opened 2026-05-01
normalize nary checks in `fm_poison`

tnuha:
This was the biggest difference I was referencing in https://github.com/AliveToolkit/alive2/pull/1095#issuecomment-4305819320. Ultimately I'd like #1095 to be a smooth merge if they beat me to it since the work is pretty solid; without this change, and assuming a simple merge, there's a de-referenced null pointer issue that comes about. I can track down some examples if desired.

### DIFF
--- a/ir/instr.cpp
+++ b/ir/instr.cpp
@@ -755,6 +755,8 @@ static StateValue fm_poison(State &s, expr a, const expr &ap, expr b,
                             bool flags_in_only = false,
                             const Type *to_ty = nullptr, int nary = 3) {
   AndExpr non_poison;
+  expr fp_a, fp_b, fp_c;
+
   non_poison.add(ap);
   if (nary >= 2)
     non_poison.add(bp);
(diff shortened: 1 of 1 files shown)


nunoplopes:
There's no dereference of a null pointer. A null expression just propagates. Can you find an example that exhibits a bug?

tnuha:
Yeah, sorry, I meant there was an assertion failure due to a null pointer propagation. This eventually crashes alive2 in certain cases like with a naive merge of master into #1095's current branch. e.g. `alive-tv --uf-float tests/alive-tv/uf-float/add-comm-double.srctgt.ll` has the attached report on [this commit](<https://github.com/tnuha/alive2/tree/cfbf1e3c63680c250cacb4673293dbf3fe47411a>) (an immediate merge of master into emmtrix's branch) but not [this one](<https://github.com/tnuha/alive2/tree/69f288e7629d2a52f18579b296a5dadc0cc8465f>).

[report.txt](https://github.com/user-attachments/files/27324613/report.txt)

To my knowledge this doesn't currently hit any bugs on master, it just seems like a necessary change as it's a current blocker for at least that branch, and it may inhibit some other work in the future. Totally understand any considerations that marks this as currently unnecessary, though.

PR #1316, opened 2026-05-01
ignore tbaa metadata when constructing ir

tommymcm:
`alive-tv` fails in `llvm2alive.cpp` when it sees TBAA metadata 
information.

Instead of requiring folks to strip this metadata themselves, we can just ignore it.
I've added it to the list of "irrelevant for correctness".

### DIFF
--- a/llvm_util/llvm2alive.cpp
+++ b/llvm_util/llvm2alive.cpp
@@ -1463,6 +1463,7 @@ public:
       case LLVMContext::MD_nosanitize:
       case LLVMContext::MD_prof:
       case LLVMContext::MD_unpredictable:
+      case LLVMContext::MD_tbaa:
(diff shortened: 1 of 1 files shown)


nunoplopes:
No, it's not irrelevant for correctness. This will cause false positives.

tommymcm:
Ope, my bad. I see that now, I had the wrong interpretation. Thanks for the quick response.

arsenm:
Can this be downgraded to a warning at least

tommymcm:
Alternatively, I was thinking of adding an option to say "I am okay with false positives caused by ignoring tbaa". 
Will defer to folks here on what the best solution is. Looking at the pr/issue history, this seems like a recurring problem for folks.

PR #1318, opened 2026-06-21
Add support for clmul

dtcxzyw:
See https://llvm.org/docs/LangRef.html#llvm-clmul-intrinsic

pext/pdep support will be added in subsequent patches.


regehr (approved):
this LGTM.

@dtcxzyw can you comment on what happens when Alive2+clmul is run over the LLVM test suite?

dtcxzyw:
> can you comment on what happens when Alive2+clmul is run over the LLVM test suite?

No new bugs.

PR #1320, opened 2026-06-22
Add support for pext/pdep

dtcxzyw:
Parallel Bits Extract: https://llvm.org/docs/LangRef.html#llvm-pext-intrinsic
Parallel Bits Deposit: https://llvm.org/docs/LangRef.html#llvm-pdep-intrinsic


regehr (approved):
this LGTM!

PR #1327, opened 2026-07-22
keep up with LLVM API changes

regehr:


llvm_util/llvm_optimizer.cpp
```diff
@@ -26,10 +26,10 @@ namespace llvm_util {
 string optimize_module(llvm::Module &M, string_view optArgs) {
   init_llvm_targets();
 
-  Triple ModuleTriple(M.getTargetTriple());
+  const Triple &ModuleTriple{M.getTargetTriple()};
   unique_ptr<llvm::TargetMachine> TM;
   if (ModuleTriple.getArch()) {
-    auto ETM = codegen::createTargetMachineForTriple(ModuleTriple);
+    auto ETM = codegen::createTargetMachineForTriple(ModuleTriple.str());
```
arsenm, line 32: This is going backwards? This should use the triple overlaid

regehr: ugh, yes, I pulled from my fork instead of upstream

